PR #9, opened 2017-11-09
Test on Node 9.

stephenplusplus:
RE: https://github.com/googleapis/nodejs-language/pull/24

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,8 +1,5 @@
 environment:
   matrix:
-  - nodejs_version: 4
-  - nodejs_version: 6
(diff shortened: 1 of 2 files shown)


callmehiphop:
Why are we still testing on Node@7 if 8 is in LTS mode?

stephenplusplus:
@lukesneeringer want to kill Node@7 tests?

stephenplusplus:
@callmehiphop I would say for now, let's keep rolling with this change. Other repos already have this structure in place, and we're just matching in the leftover repos. If we want to kill 7, we can start a new PR spree across all of the repos.

PR #11, opened 2017-11-14
Add missing error code to RowError

### LINKED ISSUE #5: Confused about row(key).exists()
_From @briangruber on October 27, 2017 16:4_

I'm confused about using the `exists()` method on a `row(key)`, hoping to get clarification. 
If I want to check if a row exists I thought I would do this

```
table.row(key).exists().then(result => {
  let exists = result[0]
}
```

If the row does exists then I do step into the `.then` and exists will be `true`. But if it doesn't exist I actually get an exception of `Unknown row: key`. So, then what is the point of `exists` if when it doesn't exist it throws an error? Is `result[0]` expected to always be `true` and if it doesn't exist I'm supposed to use the exception? Or am I missing something?

Thanks!

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2702_

kolodny:
Fixes #5

There was a missing err.code assignment that was missing in RowError, [Family](https://github.com/googleapis/nodejs-bigtable/blob/443ba9b2bf33de7de6cf49b3a1feb231eda88375/src/family.js#L27-L30) has a similar snippet.

Also added tests around the `row.exists()` method.

### DIFF
--- a/src/row.js
+++ b/src/row.js
@@ -34,6 +34,7 @@ var Mutation = require('./mutation.js');
  */
 var RowError = createErrorClass('RowError', function(row) {
   this.message = 'Unknown row: ' + row + '.';
+  this.code = 404;
 });
 
 /**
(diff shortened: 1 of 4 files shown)


callmehiphop:
@kolodny I went ahead and made a commit for unit tests, if it looks good to you I'll merge this asap.

kolodny:
Looks good to me!

PR #17, opened 2017-12-08
Add smart retries to table.mutate

kolodny:
- [x] Tests and linter pass

A couple of things to note about this PR:

- There's a bug where faketimers don't play nicely streams. I filed a bug on [process-nextick-args](https://github.com/calvinmetcalf/process-nextick-args/pull/10) which causes an [issue of readable-streams](https://github.com/nodejs/readable-stream/issues/316), so that's the deal with the patch files. Node `v4` has a nested node_modules structure while `>= v6` is flat.
- [`patch-package`](https://github.com/ds300/patch-package) doesn't handle possible missing packages so that's why the git commands are run manually


kolodny:
I can get coverage for unit tests back to 100% with the following patch, although I'm not sure it's the right thing to do

```diff
diff --git a/test/table.js b/test/table.js
index f348120..9173788 100644
--- a/test/table.js
+++ b/test/table.js
@@ -999,8 +999,10 @@ describe('Bigtable/Table', function() {
     describe('error', function() {
       describe('API errors', function() {
         var error = new Error('err');
+        var setTimeoutStub;

         beforeEach(function() {
+          setTimeoutStub = sinon.stub(global, 'setTimeout').callsFake(setImmediate);
           table.requestStream = function() {
             var stream = new Stream({
               objectMode: true,
@@ -1015,6 +1017,10 @@ describe('Bigtable/Table', function() {
           };
         });

+        afterEach(function() {
+          setTimeoutStub.restore();
+        });
+
         it('should return the error to the callback', function(done) {
           table.maxRetries = 0;
           table.mutate(entries, function(err) {
@@ -1022,6 +1028,17 @@ describe('Bigtable/Table', function() {
             done();
           });
         });
+
+        it('attempts to retry the mutation', function(done) {
+          table.maxRetries = 1;
+          table.mutate(entries, function(err) {
+            assert.strictEqual(err, error);
+            assert.strictEqual(setTimeoutStub.callCount, 1);
+            assert(setTimeoutStub.getCall(0).args[1] > 2000);
+            assert(setTimeoutStub.getCall(0).args[1] < 3000);
+            done();
+          });
+        });
       });

       describe('mutation errors', function() {
```

package.json
```diff
@@ -78,6 +81,7 @@
     "eslint-config-prettier": "^2.6.0",
     "eslint-plugin-node": "^5.2.0",
     "eslint-plugin-prettier": "^2.3.1",
+    "grpc": "^1.7.2",
```
stephenplusplus: You can actually get `grpc` through the common-grpc module:

```js
var grpc = require('@google-cloud/common-grpc').grpc;
```

stephenplusplus:
This PR introduces an "integration" test command, but we already have system tests that are supposed to be the same thing: "Does our code work when _integrated_ with the _system_?" (Using the name "system tests" was a decision that involved more languages than just us). I also see a reference to "acceptance" testing. I think we should keep everything under one roof, "system tests", since that's our pattern across all of the API libraries. It should only be one command for a contributor/maintainer to determine if their changes introduce any issues with the upstream API, and that's historically been "npm run system-test".

Regarding the patching, I would prefer to wait for https://github.com/calvinmetcalf/process-nextick-args/pull/10 to be merged and released.

src/table.js
```diff
@@ -927,45 +932,75 @@ Table.prototype.insert = function(entries, callback) {
...
 
-        var status = commonGrpc.Service.decorateStatus_(entry.status);
-        status.entry = entries[entry.index];
-
-        mutationErrors.push(status);
+  var onBatchResponse = error => {
+    if (pendingEntryIndices.size !== 0 && requestsMade <= maxRetries) {
+      setTimeout(
+        makeNextRequestBatch,
+        retryRequest.getNextRetryDelay(requestsMade)
```
stephenplusplus: @kolodny isn't this already done for us in retry-request? https://github.com/stephenplusplus/retry-request/blob/1b666a7d9b75599a99e84d78ee96fa4376e53c7e/index.js#L155

kolodny: This is for followup requests due to application errors (a mutation failed). With `retry-request` the request is only retried when the status code matches some value, in this case, the code is always 200 and the error happens way later in a future stream emit event. Retry request isn't the one retrying the request at all. The snippet linked is only happening in `retryAfterDelay` which we would never hit.

Technically the setTimeout here can live in retry-request and if the `currentRetryAttempt` is `> 0` do the initial request after whatever timeout is appropriate. However I feel it's more correct for the timeout to live here since the library can choose to make it a leading request (i.e. do the followup request without delay)

package.json
kolodny: Nice, thanks!

kolodny:
Regarding patching, after calvinmetcalf/process-nextick-args#10 is merged (it already was), nodejs/readable-stream#317 needs to be merged, publish a release (note this module is basically part of node core), and then wait for `through2` to publish a release with that updated readable-stream dependency. I figured this was a good enough workaround in the meantime

src/table.js
stephenplusplus: Sorry, I thought the initial holding logic was in retry-request. I mentioned that in that email thread we had, and thought it was implemented. Sorry for the confusion. I think that's the right way to go-- if retry-request is given `options.currentRetryAttempt`, it should hold until `getNextRetryDelay(options.currentRetryAttempt)`. This means the user-land application doesn't need to know about `getNextRetryDelay` (it is not intended to be a public method), and they can let retry-request worry about calculating those delays. If it ever creates problems in the future, we can always expose a setting in retry-request to disable the initial delay.

stephenplusplus:
I see, thanks for the rundown. Yeah, I suppose the patching will have to do for now.

src/table.js
kolodny: Sounds good. I sent stephenplusplus/retry-request#19 for review, when that gets published, common-grpc will need to push a release to use that updated module, then these two lines can change to `makeNextRequestBatch()`

kolodny: done

stephenplusplus:
Thanks for splitting that out. I believe there might be some work for the code coverage still.

kolodny:
I can add the following test to the unit tests to get it back to 100%. Is that something you'd be ok with?

```js
    describe('retries', function() {
      var fakeStatuses = [
        [
          {
            status: {
              code: 0,
            },
          },
          {
            status: {
              code: 4,
            },
          },
        ],
        [
          {
            status: {
              code: 0,
            },
          },

        ]
      ];

      beforeEach(function() {
        FakeGrpcService.decorateStatus_ = function(status) {
          return {};
        };
        table.requestStream = function() {
          var stream = new Stream({
            objectMode: true,
          });

          setImmediate(function() {
            stream.emit('request');
            stream.end({entries: fakeStatuses.shift()});
          });

          return stream;
        };
      });

      it('should succeed after a retry', function(done) {
        table.maxRetries = 1;
        table.mutate(entries, done);
      });
    });
  });
```

stephenplusplus:
That test looks good to me. I also think we should have another unit test to confirm only the correct subset of entries from the original request is sent when a retry is made, effectively testing the new `pendinEntryIndices`, `entryToIndex` holders.

kolodny:
Done, coverage appears to be back to 100%

stephenplusplus:
I made some small changes, just to conventionalize the code (stream error handling first, all caps file-wide const name, etc). Tests still pass, but please make sure these changes look harmless to you as well.

kolodny:
Look good to me!

PR #21, opened 2017-12-14
Fix#10 Implement read-rows-acceptance-test

### LINKED ISSUE #10: Implement read-rows-acceptance-test.
There is a standard set of tests driven by a json file.  Here's the [Python version](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/49e744c81c50db4484a00b4ab4ed1423fd21c5d1/bigtable/tests/unit/test_row_data.py#L435), the [GoLang version](https://github.com/GoogleCloudPlatform/google-cloud-go/blob/99eed3309ff7980d16154d2258600a5b38b129a9/bigtable/reader_test.go#L220) and the [Java version](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/scanner/ReadRowsAcceptanceTest.java).

It's a standard set of tests that show that a Cloud Bigtable client correctly parses `ReadRowResponse` `CellChunk`s.  This is a blocker to saying that this is an alpha level client.

ajaaym:
Fixes #10 Implement read-rows-acceptance-test

- [x] Tests and linter pass


ajaaym:
Closing this, will create new pull request with single commit to keep commit log clean and also stopping chunk processing loop for stream end.

PR #22, opened 2017-12-17
Fix #10 Implement read-rows-acceptance-test

### LINKED ISSUE #10: Implement read-rows-acceptance-test.

ajaaym:
Fixes #10 Implement read-rows-acceptance-test 
- [x] Tests and linter pass

### DIFF
--- a/src/chunkformatter.js
+++ b/src/chunkformatter.js
@@ -69,17 +69,21 @@ function ChunkFormatter() {
  */
 ChunkFormatter.prototype.formatChunks = function(chunks, options, callback) {
   options = options || {};
+  let shouldContinue;
   for (const chunk of chunks) {
     switch (this.state) {
       case RowStateEnum.NEW_ROW:
-        this.newRow(chunk, options, callback);
+        shouldContinue = this.newRow(chunk, options, callback);
         break;
       case RowStateEnum.ROW_IN_PROGRESS:
-        this.rowInProgress(chunk, options, callback);
+        shouldContinue = this.rowInProgress(chunk, options, callback);
         break;
       case RowStateEnum.CELL_IN_PROGRESS:
-        this.cellInProgress(chunk, options, callback);
+        shouldContinue = this.cellInProgress(chunk, options, callback);
         break;
+    }
+    if (shouldContinue === false) {
+      break;
     }
   }
 };
@@ -89,7 +93,7 @@ ChunkFormatter.prototype.formatChunks = function(chunks, options, callback) {
  * @throws Error if there is any uncommitted row.
  */
 ChunkFormatter.prototype.onStreamEnd = function() {
-  this.isError(
+  this.invariant(
     typeof this.row.key !== 'undefined',
     'Response ended with pending row without commit',
     this.row
@@ -123,7 +127,7 @@ ChunkFormatter.prototype.commit = function() {
  * @param {string} text Error text
  * @param {object} chunk chunk object to append to text
  */
-ChunkFormatter.prototype.isError = function(condition, text, chunk) {
+ChunkFormatter.prototype.invariant = function(condition, text, chunk) {
   if (condition) {
     throw new Error(`${text}: ${JSON.stringify(chunk)}`);
   }
@@ -134,7 +138,7 @@ ChunkFormatter.prototype.isError = function(condition, text, chunk) {
  * @param {chunk} chunk chunk to validate for valuesize and commitRow
  */
 ChunkFormatter.prototype.validateValueSizeAndCommitRow = function(chunk) {
-  this.isError(
+  this.invariant(
     chunk.valueSize > 0 && chunk.commitRow,
     'A row cannot be have a value size and be a commit row',
     chunk
@@ -146,7 +150,7 @@ ChunkFormatter.prototype.validateValueSizeAndCommitRow = function(chunk) {
  * @param {chunk} chunk chunk to validate for resetrow
  */
 ChunkFormatter.prototype.validateResetRow = function(chunk) {
-  this.isError(
+  this.invariant(
     chunk.resetRow &&
       (chunk.rowKey ||
         chunk.familyName ||
@@ -166,24 +170,24 @@ ChunkFormatter.prototype.validateNewRow = function(chunk) {
   const row = this.row;
   const prevRowKey = this.prevRowKey;
   const newRowKey = Mutation.convertFromBytes(chunk.rowKey);
-  this.isError(
+  this.invariant(
     typeof row.key !== 'undefined',
     'A new row cannot have existing state',
     chunk
   );
-  this.isError(
+  this.invariant(
     typeof chunk.rowKey === 'undefined' || chunk.rowKey === '',
     'A row key must be set',
     chunk
   );
-  this.isError(chunk.resetRow, 'A new row cannot be reset', chunk);
-  this.isError(
+  this.invariant(chunk.resetRow, 'A new row cannot be reset', chunk);
+  this.invariant(
     prevRowKey === newRowKey,
     'A commit happened but the same key followed',
     chunk
   );
-  this.isError(!chunk.familyName, 'A family must be set', chunk);
-  this.isError(!chunk.qualifier, 'A column qualifier must be set', chunk);
+  this.invariant(!chunk.familyName, 'A family must be set', chunk);
+  this.invariant(!chunk.qualifier, 'A column qualifier must be set', chunk);
   this.validateValueSizeAndCommitRow(chunk);
 };
 /**
@@ -195,13 +199,13 @@ ChunkFormatter.prototype.validateRowInProgress = function(chunk) {
   const row = this.row;
   this.validateResetRow(chunk);
   const newRowKey = Mutation.convertFromBytes(chunk.rowKey);
-  this.isError(
+  this.invariant(
     chunk.rowKey && newRowKey !== row.key,
     'A commit is required between row keys',
     chunk
   );
   this.validateValueSizeAndCommitRow(chunk);
-  this.isError(
+  this.invariant(
     chunk.familyName && !chunk.qualifier,
     'A qualifier must be specified',
     chunk
@@ -224,8 +228,9 @@ ChunkFormatter.prototype.validateCellInProgress = function(chunk) {
  */
 ChunkFormatter.prototype.moveToNextState = function(chunk, callback) {
   const row = this.row;
+  let shouldContinue = true;
   if (chunk.commitRow) {
-    callback(null, row);
+    shouldContinue = callback(null, row);
     this.commit();
   } else {
     if (chunk.valueSize > 0) {
@@ -234,6 +239,7 @@ ChunkFormatter.prototype.moveToNextState = function(chunk, callback) {
       this.state = RowStateEnum.ROW_IN_PROGRESS;
     }
   }
+  return shouldContinue;
 };
 /**
  * Process chunk when in NEW_ROW state.
@@ -241,6 +247,7 @@ ChunkFormatter.prototype.moveToNextState = function(chunk, callback) {
  * @param {chunks} chunk chunk to process
  * @param {options} options options
  * @param {callback} callback callback to call with row as and when generates
+ * @returns {boolean} return false to stop further processing.
  */
 ChunkFormatter.prototype.newRow = function(chunk, options, callback) {
   const newRowKey = Mutation.convertFromBytes(chunk.rowKey);
@@ -255,10 +262,9 @@ ChunkFormatter.prototype.newRow = function(chunk, options, callback) {
     value: Mutation.convertFromBytes(chunk.value, options),
     labels: chunk.labels,
     timestamp: chunk.timestampMicros,
-    size: chunk.valueSize,
   };
   this.qualifiers.push(this.qualifier);
-  this.moveToNextState(chunk, callback);
+  return this.moveToNextState(chunk, callback);
 };
 /**
  * Process chunk when in ROW_IN_PROGRESS state.
@@ -266,6 +272,7 @@ ChunkFormatter.prototype.newRow = function(chunk, options, callback) {
  * @param {chunk} chunk chunk to process
  * @param {*} options  option
  * @param {*} callback callback to call with row as and when generates
+ * @returns {boolean} return false to stop further processing.
  */
 ChunkFormatter.prototype.rowInProgress = function(chunk, options, callback) {
   this.validateRowInProgress(chunk);
@@ -286,10 +293,9 @@ ChunkFormatter.prototype.rowInProgress = function(chunk, options, callback) {
     value: Mutation.convertFromBytes(chunk.value, options),
     labels: chunk.labels,
     timestamp: chunk.timestampMicros,
-    size: chunk.valueSize,
   };
   this.qualifiers.push(this.qualifier);
-  this.moveToNextState(chunk, callback);
+  return this.moveToNextState(chunk, callback);
 };
 /**
  * Process chunk when in CELl_IN_PROGRESS state.
@@ -297,6 +303,7 @@ ChunkFormatter.prototype.rowInProgress = function(chunk, options, callback) {
  * @param {chunk} chunk chunk to process
  * @param {options} options options
  * @param {callback} callback callback to call with row as and when generates
+ * @returns {boolean} return false to stop further processing.
  */
 ChunkFormatter.prototype.cellInProgress = function(chunk, options, callback) {
   this.validateCellInProgress(chunk);
@@ -305,8 +312,7 @@ ChunkFormatter.prototype.cellInProgress = function(chunk, options, callback) {
   }
   this.qualifier.value =
     this.qualifier.value + Mutation.convertFromBytes(chunk.value, options);
-  this.qualifier.size = chunk.valueSize;
-  this.moveToNextState(chunk, callback);
+  return this.moveToNextState(chunk, callback);
 };
 
 module.exports = ChunkFormatter;
--- a/src/chunkformatter.js
+++ b/src/chunkformatter.js
@@ -301,7 +301,6 @@ ChunkFormatter.prototype.cellInProgress = function(chunk, options, callback) {
   if (chunk.resetRow) {
     return this.reset();
   }
-  const row = this.row;
   this.qualifier.value =
     this.qualifier.value + Mutation.convertFromBytes(chunk.value, options);
   this.qualifier.size = 0;
--- a/src/chunkformatter.js
+++ b/src/chunkformatter.js
@@ -215,6 +215,25 @@ ChunkFormatter.prototype.validateCellInProgress = function(chunk) {
   this.validateValueSizeAndCommitRow(chunk);
 };
 /**
+ * Moves to next state in processing.
+ * @private
+ * @param {chunk} chunk chunk in process
+ * @param {*} callback callback to call with row as and when generates
+ */
+ChunkFormatter.prototype.moveToNextState = function(chunk, callback) {
+  const row = this.row;
+  if (chunk.commitRow) {
+    callback(null, row);
+    this.commit();
+  } else {
+    if (chunk.valueSize > 0) {
+      this.state = RowStateEnum.CELL_IN_PROGRESS;
+    } else {
+      this.state = RowStateEnum.ROW_IN_PROGRESS;
+    }
+  }
+};
+/**
  * Process chunk when in NEW_ROW state.
  * @private
  * @param {chunks} chunk chunk to process
@@ -237,16 +256,7 @@ ChunkFormatter.prototype.newRow = function(chunk, options, callback) {
     size: chunk.valueSize,
   };
   this.qualifiers.push(this.qualifier);
-  if (chunk.commitRow) {
-    callback(null, row);
-    this.commit();
-  } else {
-    if (chunk.valueSize > 0) {
-      this.state = RowStateEnum.CELL_IN_PROGRESS;
-    } else {
-      this.state = RowStateEnum.ROW_IN_PROGRESS;
-    }
-  }
+  this.moveToNextState(chunk, callback);
 };
 /**
  * Process chunk when in ROW_IN_PROGRESS state.
@@ -277,14 +287,7 @@ ChunkFormatter.prototype.rowInProgress = function(chunk, options, callback) {
     size: chunk.valueSize,
   };
   this.qualifiers.push(this.qualifier);
-  if (chunk.commitRow) {
-    callback(null, row);
-    this.commit();
-  } else {
-    if (chunk.valueSize > 0) {
-      this.state = RowStateEnum.CELL_IN_PROGRESS;
-    }
-  }
+  this.moveToNextState(chunk, callback);
 };
 /**
  * Process chunk when in CELl_IN_PROGRESS state.
@@ -302,14 +305,7 @@ ChunkFormatter.prototype.cellInProgress = function(chunk, options, callback) {
   this.qualifier.value =
     this.qualifier.value + Mutation.convertFromBytes(chunk.value, options);
   this.qualifier.size = 0;
-  if (chunk.commitRow) {
-    callback(null, row);
-    this.commit();
-  } else {
-    if (chunk.valueSize === 0) {
-      this.state = RowStateEnum.ROW_IN_PROGRESS;
-    }
-  }
+  this.moveToNextState(chunk, callback);
 };
 
 module.exports = ChunkFormatter;
--- a/src/chunkformatter.js
+++ b/src/chunkformatter.js
@@ -89,9 +89,11 @@ ChunkFormatter.prototype.formatChunks = function(chunks, options, callback) {
  * @throws Error if there is any uncommitted row.
  */
 ChunkFormatter.prototype.onStreamEnd = function() {
-  if (typeof this.row.key !== 'undefined') {
-    throw new Error('Response ended with pending row without commit');
-  }
+  this.isError(
+    typeof this.row.key !== 'undefined',
+    'Response ended with pending row without commit',
+    this.row
+  );
 };
 /**
  * Resets state of formatter
@@ -123,7 +125,7 @@ ChunkFormatter.prototype.commit = function() {
  */
 ChunkFormatter.prototype.isError = function(condition, text, chunk) {
   if (condition) {
-    throw new Error(`${text}: ${chunk}`);
+    throw new Error(`${text}: ${JSON.stringify(chunk)}`);
   }
 };
 /**
--- a/src/chunkformatter.js
+++ b/src/chunkformatter.js
@@ -15,17 +15,27 @@
  */
 'use strict';
 const Mutation = require('./mutation.js');
+/**
+ * Enum for chunk formatter Row state.
+ * NEW_ROW: inital state or state after commitRow or resetRow
+ * ROW_IN_PROGRESS: state after first valid chunk without commitRow or resetRow
+ * CELL_IN_PROGRESS: state when valueSize > 0(partial cell)
+ */
+const RowStateEnum = Object.freeze({
+  NEW_ROW: 1,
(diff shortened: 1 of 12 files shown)


ajaaym:
I signed it!

src/chunkformatter.js
```diff
@@ -0,0 +1,217 @@
...
+ * `resetRow` This is a boolean telling us that all the previous chunks are to
+ *     be discarded.
+ *
+ * @public
+ *
+ * @param {chunk[]} chunks The list of chunks.
+ * @param {object} [options] Formatting options.
+ *
+ * @example
+ * Row.formatChunks_(chunks);
```
sduskis: Please change this to `ChunkFormatter. formatChunks_`

ajaaym:
@kolodny Thanks for feedback. I updated PR as per pull request except about callback. Had discussion about callbacks vs returning array of rows with @sduskis  earlier and decided to use callback so to keep less number of live reference and also it helps to end stream earlier without parsing all chunks in a message.

I will create new PR with single commit log once it is approved to keep commit log clean.

sduskis:
Please don't create a new PR.  When the review is complete, we will "squash and merge" the PR, which will turn all commits into a single commit.

ajaaym:
Ok.

src/chunkformatter.js
```diff
@@ -0,0 +1,311 @@
...
+ChunkFormatter.prototype.newRow = function(chunk, options, callback) {
+  const newRowKey = Mutation.convertFromBytes(chunk.rowKey);
+  this.validateNewRow(chunk);
+  const row = this.row;
+  row.key = newRowKey;
+  row.data = {};
+  this.family = row.data[chunk.familyName.value] = {};
+  const qualifierName = Mutation.convertFromBytes(chunk.qualifier.value);
+  this.qualifiers = this.family[qualifierName] = [];
+  this.qualifier = {
```
sduskis: `this.qualifier` should probably be named `this.cellInProgress`.

src/chunkformatter.js
```diff
@@ -0,0 +1,311 @@
...
+ * @param {callback} callback callback to call with row as and when generates
+ */
+ChunkFormatter.prototype.cellInProgress = function(chunk, options, callback) {
+  this.validateCellInProgress(chunk);
+  if (chunk.resetRow) {
+    return this.reset();
+  }
+  this.qualifier.value =
+    this.qualifier.value + Mutation.convertFromBytes(chunk.value, options);
+  this.qualifier.size = 0;
```
sduskis: Shouldn't `this.qualifier.size = chunk.size`?

src/chunkformatter.js
ajaaym: this.row, this.family and this.qualifier in the ChunkFormatter instance is in progress row, family & qualifier. Also there I named a method cellInProgress to process CELL_IN_PROGRESS state, It will be confusing with that name. If we want to make it explicit, I can rename state processing method to processNewRow, processRowInProgress, processCellInProgress & rename state holder to rowInProgress, familyInProgress & cellInProgress. Let me know.

src/chunkformatter.js
ajaaym: yes will update PR with this fix.

src/chunkformatter.js
```diff
@@ -0,0 +1,311 @@
...
+  CELL_IN_PROGRESS: 3,
+});
+/**
+ * ChunkFormatter formats all incoming chunks in to row
+ * keeps all intermediate state until end of stream.
+ * Should use new instance for each request.
+ */
+function ChunkFormatter() {
+  if (!(this instanceof ChunkFormatter)) {
+    return new ChunkFormatter();
```
chauhraj: Why is this block required?

src/chunkformatter.js
```diff
@@ -0,0 +1,311 @@
...
+    }
+  }
+};
+/**
+ * should be called at end of the stream to check if there is any pending row.
+ * @public
+ * @throws Error if there is any uncommitted row.
+ */
+ChunkFormatter.prototype.onStreamEnd = function() {
+  if (typeof this.row.key !== 'undefined') {
```
chauhraj: Use `isError`.

src/chunkformatter.js
```diff
@@ -0,0 +1,311 @@
...
+  this.prevRowKey = row.key;
+};
+/**
+ * Error checker if true throws error
+ * @private
+ * @param {boolean} condition condition to evaluate
+ * @param {string} text Error text
+ * @param {object} chunk chunk object to append to text
+ */
+ChunkFormatter.prototype.isError = function(condition, text, chunk) {
```
chauhraj: Can we reuse 'assert' which can be renamed to 'require' or use as is.

src/chunkformatter.js
```diff
@@ -0,0 +1,311 @@
...
+  if (typeof this.row.key !== 'undefined') {
+    throw new Error('Response ended with pending row without commit');
+  }
+};
+/**
+ * Resets state of formatter
+ * @private
+ */
+ChunkFormatter.prototype.reset = function() {
+  this.prevRowKey = '';
```
chauhraj: Shouldn't this be either undefined or null. It can be string or raw bytes depending on `decode` value.

src/chunkformatter.js
```diff
@@ -0,0 +1,311 @@
...
+/**
+ * Error checker if true throws error
+ * @private
+ * @param {boolean} condition condition to evaluate
+ * @param {string} text Error text
+ * @param {object} chunk chunk object to append to text
+ */
+ChunkFormatter.prototype.isError = function(condition, text, chunk) {
+  if (condition) {
+    throw new Error(`${text}: ${chunk}`);
```
chauhraj: `${chunk}` will just convert to `[object Object]` To have a meaningful information about chunk, maybe use `${JSON.stringify(chunk)}` or something else.

src/chunkformatter.js
```diff
@@ -0,0 +1,311 @@
...
+  row.key = newRowKey;
+  row.data = {};
+  this.family = row.data[chunk.familyName.value] = {};
+  const qualifierName = Mutation.convertFromBytes(chunk.qualifier.value);
+  this.qualifiers = this.family[qualifierName] = [];
+  this.qualifier = {
+    value: Mutation.convertFromBytes(chunk.value, options),
+    labels: chunk.labels,
+    timestamp: chunk.timestampMicros,
+    size: chunk.valueSize,
```
chauhraj: Is there any value in having size as part of the data? this.qualifier is the actual data in which user is interested in. chunk.valueSize is meant for the API to indicate when the Cell is complete. It helps API to pre-allocate a buffer to store the information and usually in bytes. Besides, everything is converted from bytes so it looses its relevance.

src/chunkformatter.js
```diff
@@ -0,0 +1,311 @@
...
+    return this.reset();
+  }
+  this.qualifier.value =
+    this.qualifier.value + Mutation.convertFromBytes(chunk.value, options);
+  this.qualifier.size = chunk.valueSize;
+  this.moveToNextState(chunk, callback);
+};
+
+module.exports = ChunkFormatter;
+module.exports.RowStateEnum = RowStateEnum;
```
chauhraj: Is there a reason to export RowStateEnum? Shouldn't this be private?

src/chunkformatter.js
ajaaym: updated

src/chunkformatter.js
ajaaym: updated

src/chunkformatter.js
ajaaym: previous row key is assigned the row key of committed row which is always converted from bytes. Row key should never be blank even.

src/chunkformatter.js
ajaaym: I agree we need better name for this. I dont think we should use assert for this as assert is mainly for debugging/test case purpose.

src/chunkformatter.js
ajaaym: I wanted to use it in test case, since its just based on number, I could have created new fake for enum in test case. I think its ok to export this, as this is immutable. But I will wait for more feedback if we shouldnt export this.

sduskis:
@ajaaym, what's the status on this?  Is this PR ready for review from @stephenplusplus yet?

ajaaym:
There are three open que where I need more feedback. 

1. Method name isError if anybody has suggestion for better name or anybody has preference to use assert instead?

2. Any issue exporting RowStateEnum. 

3. Size in output row. Previous code was sending out size in output. I agree with Rajiv that we dont need to output size. So should we keep it or remove it.

sduskis:
1. Maybe `checkCondition` or `validate` instead of `isError`?  There's a `[Preconditions](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/base/Preconditions.html)` class in guava with this type of functionality

2. I don't see anything wrong.

3. The size of the output is necessary, and in the default cause (0) it's downright confusing.  I think it should be removed.

src/chunkformatter.js
chauhraj: This is not necessarily true. Java/C++ both have the notion of asserts and they can exist very much in the code. Essentially, it is like preconditions. However, given that one can rename assert and call it either 'require', or 'precondition'.

src/chunkformatter.js
chauhraj: Only reason, I was saying that if one look at [CellChunk::row-key] (https://github.com/googleapis/nodejs-bigtable/blob/master/protos/google/bigtable/v2/bigtable.proto#L105) datatype is bytes.

src/table.js
```diff
@@ -488,30 +489,45 @@ Table.prototype.createReadStream = function(options) {
...
+        var throughStream = this;
+        try {
+          chunkFormatter.formatChunks(
+            data.chunks,
+            {
+              decode: options.decode,
+            },
+            (err, rowData) => {
+              if (stream._ended) {
+                throw new Error('stream ended');
```
kolodny: It appears that you throw this error to "break" out of the loop in formatChunks, and catch the error on line 515. This doesn't feel like idiomatic javascript, can you instead return false in this loop and that will signal to `formatChunks` that you want to terminate early? Lodash follows a similar pattern https://lodash.com/docs/4.17.4#forEach

src/chunkformatter.js
kolodny: I believe `invariant` would be the proper term https://github.com/facebook/react/blob/ea9714807b0d694e6f67d52ccf7aac93439b16f7/packages/react-native-renderer/src/ReactNativeComponentTree.js#L28

ajaaym:
@sduskis @kolodny @chauhraj Updated with following.
1. removed Size from output row.
2. Changed to stop processing of loop based on callback returns value.
3. Changed name isError to invariant

sduskis:
@stephenplusplus, we're ready for your review.

src/table.js
```diff
@@ -488,30 +489,44 @@ Table.prototype.createReadStream = function(options) {
...
+              decode: options.decode,
+            },
+            (err, rowData) => {
+              if (stream._ended) {
+                return false;
+              } else {
+                var row = self.row(rowData.key);
+                row.data = rowData.data;
+                throughStream.push(row);
+                return true;
```
stephenplusplus: No return necessary. This whole block could look like:

```js
if (stream._ended) {
  return;
}

var row = self.row(rowData.key);
row.data = rowData.data;
throughStream.push(row);
```

src/table.js
```diff
@@ -488,30 +489,44 @@ Table.prototype.createReadStream = function(options) {
...
-        row.data = rowData.data;
-        throughStream.push(row);
-      });
-
-      next();
-    }),
+        next();
+      },
+      function(callback) {
+        try {
```
stephenplusplus: It would be nice to remove the try/catch here as well, something more like:

```js
through.obj(
  function(data, enc, next) {
    // ...
    next();
  },
  chunkFormatter.onStreamEnd
);
```

`onStreamEnd` should accept a callback, and return an error if it ended prematurely.

src/chunkformatter.js
```diff
@@ -0,0 +1,319 @@
...
+ * //   }
+ * // }
+ */
+ChunkFormatter.prototype.formatChunks = function(chunks, options, callback) {
+  options = options || {};
+  let shouldContinue;
+  for (const chunk of chunks) {
+    switch (this.state) {
+      case RowStateEnum.NEW_ROW:
+        shouldContinue = this.newRow(chunk, options, callback);
```
stephenplusplus: Is it possible to use an instance variable to track if it should continue? This is not an obvious pattern-- passing a callback and receiving a value that doesn't match the purpose of the method.

src/table.js
stephenplusplus: After further digging, I see how the `return` value is used as the meaning of `shouldContinue` in ChunkFormatter. Can we be more direct about that, and just have a method like `chunkFormatter.exit()` (or `stopProcessing()`, etc). It's not clear that this function has any duties that work backwards into ChunkFormatter, and it would be easy to imagine a contributor or maintainer in the future getting rid of the `true`/`false` aspect in a refactor.

src/chunkformatter.js
```diff
@@ -0,0 +1,319 @@
...
+        break;
+      case RowStateEnum.ROW_IN_PROGRESS:
+        shouldContinue = this.rowInProgress(chunk, options, callback);
+        break;
+      case RowStateEnum.CELL_IN_PROGRESS:
+        shouldContinue = this.cellInProgress(chunk, options, callback);
+        break;
+    }
+    if (shouldContinue === false) {
```
stephenplusplus: Following my previous comment, this would be something like:

```js
if (this.wasProcessingExited) {
  break;
}
```

src/chunkformatter.js
stephenplusplus: Pushing the `try/catch` here seems fine, like:

```js
for (const chunk of chunks) {
  switch (this.state) {
    case RowStateEnum.NEW_ROW:
      try {
        this.processNewRow(chunk, options);
      } catch (e) {
        callback(e);
      }
  }
}
```

ajaaym:
After giving more thoughts to @stephenplusplus feedback, I think best way to implement this is stateful transform. So I renamed ChunkFormatter to ChunkTransformer and extended it with Streams Transform interface. Also added destroy method to make it compatible with pump & pumpify.

src/chunktransformer.js
```diff
@@ -0,0 +1,337 @@
...
+ *     be discarded.
+ *
+ * @public
+ *
+ * @param {object} data readrows response containing array of chunks.
+ * @param {object} [enc] encoding options.
+ * @param {callback} next callback will be called once data is processed, with error if any error in processing
+ */
+ChunkTransformer.prototype._transform = function(data, enc, next) {
+  try {
```
stephenplusplus: Doing the `try/catch` inside each data event seems heavy, and like perf may be negatively affected. Is it possible that if the data is invalid, the called method, i.e. `processNewRow()`, could just destroy the stream right there with the error?

src/chunktransformer.js
```diff
@@ -0,0 +1,337 @@
...
+  this.prevRowKey = row.key;
+};
+/**
+ * Error checker if true throws error
+ * @private
+ * @param {boolean} condition condition to evaluate
+ * @param {string} text Error text
+ * @param {object} chunk chunk object to append to text
+ */
+ChunkTransformer.prototype.invariant = function(condition, text, chunk) {
```
stephenplusplus: When scanning the code, it's not clear that any call to `this.invariant()` could result in a thrown error. I would rather see something like this, so we aren't try/catching anywhere:

```diff
- this.invariant(condition, 'Error message.', value);
+ if (condition) {
+   this.destroy(this.generateError('Error message.', value));
+ }
```

You could use [create-error-class](https://github.com/floatdrop/create-error-class):

```js
const createErrorClass = require('create-error-class');

const TransformError = createErrorClass('TransformError', function (props) {
  this.message = `${props.message}: ${JSON.stringify(props.chunk)}`;
});

// ...

ChunkTransformer.prototype.validationMethod = function(chunk) {
  if (condition) {
    this.destroy(new TransformError({
      message: 'Error message.',
      chunk: chunk
    });
  }
};
```

src/chunktransformer.js
ajaaym: @stephenplusplus even after calling destroy right there, current execution path from point of error to for loop needs to be short circuit. I see below option to achieve this. Let me know if there is any other way.


 Use _error instance object and it will be initialized in constructor as below:

```javascript
  this._error = {
    hasError: false,
    errorObj: null,
  };
```
Invariant will be 
```javascript
  if (condition) {
    this._error.hasError = true;
    this._error.errorObj = new TransformError({
      message: text,
      chunk: chunk,
    });
  }
```
processNewRow will be
```javascript
ChunkTransformer.prototype.processNewRow = function(chunk) {
  const newRowKey = Mutation.convertFromBytes(chunk.rowKey);
  this.validateNewRow(chunk);
  if (this._error.hasError) {
    return;
  }
```
And transform will come out of for loop by checking on hasError
```javascript
ChunkTransformer.prototype._transform = function(data, enc, next) {
  for (const chunk of data.chunks) {
    if (this._error.hasError) {
      return next(this._error.errorObj);
    }
    switch (this.state) {
      case RowStateEnum.NEW_ROW:
        this.processNewRow(chunk, this.options);
        break;
      case RowStateEnum.ROW_IN_PROGRESS:
        this.processRowInProgress(chunk, this.options);
        break;
      case RowStateEnum.CELL_IN_PROGRESS:
        this.processCellInProgress(chunk, this.options);
        break;
    }
  }
  next();
};
```

In this case all conditions in validation method would still be checked only process* and transform/flush will be short circuited. If we make invariant return the condition we can use that to short circuit validation method as well like below.
```javascript
ChunkTransformer.prototype.invariant = function(condition, text, chunk) {
  if (condition) {
    this._error.hasError = true;
    this._error.errorObj = new TransformError({
      message: text,
      chunk: chunk,
    });
  }
  return condition;
};
```
validations method would be like this:
```javascript
ChunkTransformer.prototype.validateNewRow = function(chunk) {
  const row = this.row;
  const prevRowKey = this.prevRowKey;
  const newRowKey = Mutation.convertFromBytes(chunk.rowKey);
  return (
    this.invariant(
      typeof row.key !== 'undefined',
      'A new row cannot have existing state',
      chunk
    ) ||
    this.invariant(
      typeof chunk.rowKey === 'undefined' || chunk.rowKey === '',
      'A row key must be set',
      chunk
    ) ||
    this.invariant(chunk.resetRow, 'A new row cannot be reset', chunk) ||
    this.invariant(
      prevRowKey === newRowKey,
      'A commit happened but the same key followed',
      chunk
    ) ||
    this.invariant(!chunk.familyName, 'A family must be set', chunk) ||
    this.invariant(!chunk.qualifier, 'A column qualifier must be set', chunk) ||
    this.validateValueSizeAndCommitRow(chunk)
  );
};
```

if we make use of return value then we can use this return value all the way up to transform function to break for loop. 

My preference would be to use return value in case of error all the way up to transform and instance variable for error object. Let me know if you have any suggestion or better way of handling this.

stephenplusplus: You should be able to re-use `this._destroyed` (set by `.destroy()`) to break out of the loop:

```js
ChunkTransformer.prototype._transform = function(data, enc, next) {
  for (const chunk of data.chunks) {
    if (this._destroyed) {
      return;
    }

    switch (this.state) {
      case RowStateEnum.NEW_ROW:
        this.processNewRow(chunk, this.options);
        break;
      case RowStateEnum.ROW_IN_PROGRESS:
        this.processRowInProgress(chunk, this.options);
        break;
      case RowStateEnum.CELL_IN_PROGRESS:
        this.processCellInProgress(chunk, this.options);
        break;
    }
  }

  next();
}
```

ajaaym: Totally missed that. Thanks alot for pointing that out. I will update PR.

ajaaym:
@stephenplusplus updated as per your feedback. Please review.

src/table.js
```diff
@@ -488,30 +489,21 @@ Table.prototype.createReadStream = function(options) {
...
-
+    chunkTransformer,
+    through.obj(function(rowData, enc, next) {
+      if (stream._ended) {
+        return next();
+      }
+      const throughStream = this;
+      const row = self.row(rowData.key);
+      row.data = rowData.data;
+      throughStream.push(row);
```
stephenplusplus: You can simplify this into:

```js
const row = self.row(rowData.key);
row.data = rowData.data;
next(null, row);
```

src/chunktransformer.js
```diff
@@ -0,0 +1,390 @@
...
+ * @private
+ * @param {chunks} chunk chunk to process
+ * @param {options} options options
+ * @param {callback} callback callback to call with row as and when generates
+ * @returns {boolean} return false to stop further processing.
+ */
+ChunkTransformer.prototype.processNewRow = function(chunk) {
+  const newRowKey = Mutation.convertFromBytes(chunk.rowKey);
+  this.validateNewRow(chunk);
+  if (this._destroyed) {
```
stephenplusplus: I think we can get rid of this. We do the same check in the transform function.

src/chunktransformer.js
```diff
@@ -0,0 +1,390 @@
...
+ * Process chunk when in ROW_IN_PROGRESS state.
+ * @private
+ * @param {chunk} chunk chunk to process
+ * @param {*} options  option
+ * @param {*} callback callback to call with row as and when generates
+ * @returns {boolean} return false to stop further processing.
+ */
+ChunkTransformer.prototype.processRowInProgress = function(chunk) {
+  this.validateRowInProgress(chunk);
+  if (this._destroyed) {
```
stephenplusplus: Same as last comment; I believe we can get rid of this.

src/chunktransformer.js
```diff
@@ -0,0 +1,390 @@
...
+ * Process chunk when in CELl_IN_PROGRESS state.
+ * @private
+ * @param {chunk} chunk chunk to process
+ * @param {options} options options
+ * @param {callback} callback callback to call with row as and when generates
+ * @returns {boolean} return false to stop further processing.
+ */
+ChunkTransformer.prototype.processCellInProgress = function(chunk) {
+  this.validateCellInProgress(chunk);
+  if (this._destroyed) {
```
stephenplusplus: Same as last comment; I believe we can get rid of this.

src/chunktransformer.js
```diff
@@ -0,0 +1,390 @@
...
+/**
+ * Validates state for new row.
+ * @private
+ * @param {chunk} chunk chunk to validate
+ */
+ChunkTransformer.prototype.validateNewRow = function(chunk) {
+  const row = this.row;
+  const prevRowKey = this.prevRowKey;
+  const newRowKey = Mutation.convertFromBytes(chunk.rowKey);
+  if (typeof row.key !== 'undefined') {
```
stephenplusplus: We could condense these a little bit:

```js
ChunkTransformer.prototype.validateNewRow = function(chunk) {
  const row = this.row;
  const prevRowKey = this.prevRowKey;
  const newRowKey = Mutation.convertFromBytes(chunk.rowKey);

  var errorMessage;

  if (typeof row.key !== 'undefined') {
    errorMessage = 'A new row cannot have existing state';
  } else if (typeof chunk.rowKey === 'undefined' || chunk.rowKey === '') {
    errorMessage = 'A row key must be set';
  } else if (chunk.resetRow) {
    errorMessage = 'A new row cannot be reset';
  } else if (prevRowKey === newRowKey) {
    errorMessage = 'A commit happened but the same key followed';
  } else if (!chunk.familyName) {
    errorMessage = 'A family must be set';
  } else if (!chunk.qualifier) {
    errorMessage = 'A column qualifier must be set';
  }

  if (errorMessage) {
    this.destroy(new TransformError({message: errorMessage, chunk}));
    return;
  }

  this.validateValueSizeAndCommitRow(chunk);
};
```

src/chunktransformer.js
```diff
@@ -0,0 +1,390 @@
...
+    );
+  }
+};
+/**
+ * Validates resetRow condition for chunk
+ * @private
+ * @param {chunk} chunk chunk to validate for resetrow
+ */
+ChunkTransformer.prototype.validateResetRow = function(chunk) {
+  if (
```
stephenplusplus: Could we refactor this to something more like:

```js
ChunkTransformer.prototype.validateResetRow = function(chunk) {
  const containsData =
    chunk.rowKey ||
    chunk.familyName ||
    chunk.qualifier ||
    chunk.value ||
    chunk.timestampMicros > 0;

  if (chunk.resetRow && containsData) {
    this.destroy(
      new TransformError({
        message: 'A reset should have no data',
        chunk: chunk,
      })
    );
  }
};
```

src/chunktransformer.js
```diff
@@ -0,0 +1,390 @@
...
+ */
+ChunkTransformer.prototype.processCellInProgress = function(chunk) {
+  this.validateCellInProgress(chunk);
+  if (this._destroyed) {
+    return;
+  }
+  if (chunk.resetRow) {
+    return this.reset();
+  }
+  this.qualifier.value =
```
stephenplusplus: We can save some space:

```js
this.qualifier.value += Mutation.convertFromBytes(chunk.value, this.options);
```

src/chunktransformer.js
ajaaym: validateNewRow has a check for existence of familyName and chunk.qualifier in chunk. which I later use it without checking.  

without check
```javascript
this.family = row.data[chunk.familyName.value] = {};
  const qualifierName = Mutation.convertFromBytes(chunk.qualifier.value);
```

Do you suggest to check for existence of that instead?
```javascript
  if (chunk.familyName && chunk.qualifier) {
    this.family = row.data[chunk.familyName.value] = {};
    const qualifierName = Mutation.convertFromBytes(chunk.qualifier.value);
    this.qualifiers = this.family[qualifierName] = [];
    this.qualifier = {
      value: Mutation.convertFromBytes(chunk.value, this.options),
      labels: chunk.labels,
      timestamp: chunk.timestampMicros,
    };
    this.qualifiers.push(this.qualifier);
    this.moveToNextState(chunk);
  }
```

stephenplusplus: I just meant we should be able to get rid of the check if the stream was destroyed, as we do that same check from the transform function.

ajaaym: I updated PR as per your latest feedback except this one to get rid of checks completely. I replaced the check this._destroyed with chunk.familyName && chunk.qualifier otherwise its throwing error bcz of this statement
```javascript
this.family = row.data[chunk.familyName.value] = {};
    const qualifierName = Mutation.convertFromBytes(chunk.qualifier.value);
```

stephenplusplus:
@ajaaym @sduskis This LGTM. (Thanks for putting up with my comments). Is this ready to go from your end?

sduskis:
+1 from me.

sduskis:
@stephenplusplus, what does @ajaaym need to do to fix the CLA warning?

ajaaym:
Looks good to me

stephenplusplus:
No work needed for the CLA bot. It's a known limitation when two members collaborate, it can't validate both authors. But since we know we're both signed, we can ignore the warning.

Ping me whenever you'd like a new release 👍

PR #27, opened 2018-01-07
Smart read retries

### LINKED ISSUE #7: bigtable: retry partially failed reads and writes
_From @arbesfeld on September 14, 2016 20:58_

My calls to bigtable `getRows()` fail intermittently, so I have had to wrap all of these methods in retry blocks. I was wondering:
1. Is this flakiness expected?
2. Have you considered adding a retry mechanism to these RPCs?

Thanks!

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#1595_

kolodny:
Fixes #7 

- [x] Tests and linter pass

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -482,7 +482,7 @@ Table.prototype.createReadStream = function(options) {
     rowsLimit = options.limit;
   }
 
-  const retryStream = through.obj();
+  const userStream = through.obj();
   let chunkTransformer;
 
   const makeNewRequest = () => {
@@ -502,17 +502,12 @@ Table.prototype.createReadStream = function(options) {
     };
     if (lastRowKey) {
       const lessThan = (lhs, rhs) => {
-        return (
-          Mutation.convertToBytes(lhs).compare(Mutation.convertToBytes(rhs)) ===
-          -1
-        );
+        const lhsBytes = Mutation.convertToBytes(lhs);
+        const rhsBytes = Mutation.convertToBytes(rhs);
+        return lhsBytes.compare(rhsBytes) === -1;
       };
-      const greaterThan = (lhs, rhs) => {
-        return lessThan(rhs, lhs);
-      };
-      const greaterThanOrEqualTo = (lhs, rhs) => {
-        return !lessThan(rhs, lhs);
-      };
+      const greaterThan = (lhs, rhs) => lessThan(rhs, lhs);
+      const greaterThanOrEqualTo = (lhs, rhs) => !lessThan(rhs, lhs);
 
       if (ranges.length === 0) {
         ranges.push({
@@ -580,11 +575,11 @@ Table.prototype.createReadStream = function(options) {
     const requestStream = this.requestStream(grpcOpts, reqOpts);
     requestStream.on('request', () => numRequestsMade++);
 
-    const stream = pumpify.obj([
+    const rowStream = pumpify.obj([
       requestStream,
       chunkTransformer,
       through.obj(function(rowData, enc, next) {
-        if (chunkTransformer._destroyed || retryStream._writableState.ended) {
+        if (chunkTransformer._destroyed || userStream._writableState.ended) {
           return next();
         }
         numRequestsMade = 0;
@@ -595,23 +590,22 @@ Table.prototype.createReadStream = function(options) {
       }),
     ]);
 
-    stream.on('error', error => {
-      // TODO - DO NOT MERGE - @stephenplusplus is this unpipe needed?
-      stream.unpipe(retryStream);
+    rowStream.on('error', error => {
+      rowStream.unpipe(userStream);
       if (
         numRequestsMade <= maxRetries &&
         READ_ROWS_RETRYABLE_STATUS_CODES.has(error.code)
       ) {
         makeNewRequest();
       } else {
-        retryStream.emit('error', error);
+        userStream.emit('error', error);
       }
     });
-    stream.pipe(retryStream);
+    rowStream.pipe(userStream);
   };
 
   makeNewRequest();
-  return retryStream;
+  return userStream;
 };
 
 /**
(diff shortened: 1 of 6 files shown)


src/table.js
```diff
@@ -445,64 +450,168 @@ Table.prototype.createReadStream = function(options) {
...
+        numRequestsMade = 0;
+        rowsRead++;
+        const row = self.row(rowData.key);
+        row.data = rowData.data;
+        next(null, row);
+      }),
+    ]);
+
+    stream.on('error', error => {
+      // TODO - DO NOT MERGE - @stephenplusplus is this unpipe needed?
```
stephenplusplus: Can you elaborate on the situation that would cause this, and what needs to happen?

kolodny: I'm not sure if an error on the `steam` will unpipe it from `retryStream`, considering that this stream could live for hours and multiple retried streams are piping to it, does this leak memory or add any significant overhead?

src/table.js
```diff
@@ -445,64 +450,168 @@ Table.prototype.createReadStream = function(options) {
...
   }
-  const chunkTransformer = new ChunkTransformer({decode: options.decode});
-  const stream = pumpify.obj([
-    this.requestStream(grpcOpts, reqOpts),
-    chunkTransformer,
-    through.obj(function(rowData, enc, next) {
-      if (stream._ended) {
-        return next();
+
+  const retryStream = through.obj();
```
stephenplusplus: Let's call the stream the user is holding either just `stream` or `userStream`.

src/table.js
```diff
@@ -445,64 +450,168 @@ Table.prototype.createReadStream = function(options) {
...
+    }
+
+    if (rowsLimit) {
+      reqOpts.rowsLimit = rowsLimit - rowsRead;
+    }
+
+    const requestStream = this.requestStream(grpcOpts, reqOpts);
+    requestStream.on('request', () => numRequestsMade++);
+
+    const stream = pumpify.obj([
```
stephenplusplus: Let's call this `requestStream` or `rowStream`. Ideally "requestStream" is piped to "userStream".

src/table.js
stephenplusplus: I'm not sure what will happen internally. Calling unpipe should be safe to do.

src/table.js
```diff
@@ -445,64 +450,168 @@ Table.prototype.createReadStream = function(options) {
...
+      objectMode: true,
+    };
+    if (lastRowKey) {
+      const lessThan = (lhs, rhs) => {
+        return (
+          Mutation.convertToBytes(lhs).compare(Mutation.convertToBytes(rhs)) ===
+          -1
+        );
+      };
+      const greaterThan = (lhs, rhs) => {
```
stephenplusplus: This should work:

```js
const greaterThan = (lhs, rhs) => lessThan(rhs, lhs);
```

src/table.js
```diff
@@ -445,64 +450,168 @@ Table.prototype.createReadStream = function(options) {
...
+        return lessThan(rhs, lhs);
+      };
+      const greaterThanOrEqualTo = (lhs, rhs) => {
```
stephenplusplus: This should work:

```js
const greaterThanOrEqualTo = (lhs, rhs) => !lessThan(rhs, lhs);
```

src/table.js
```diff
@@ -445,64 +450,168 @@ Table.prototype.createReadStream = function(options) {
...
+      } else {
+        // Readjust and/or remove ranges based on previous valid row reads.
+
+        // Iterate backward since items may need to be removed.
+        for (let index = ranges.length - 1; index >= 0; index--) {
+          const range = ranges[index];
+          const normalizeRangeValue = rangeSide => {
+            return is.object(rangeSide) ? rangeSide.value : rangeSide;
+          };
+          const startValue = normalizeRangeValue(range.start);
```
stephenplusplus: I think we can avoid the inline function `normalizeRangeValue` if we just:

```js
const startValue = range.start.value || range.start;
const endValue = range.end.value || range.end;
```

kolodny: It's possible for `range.start.value` to be an empty string which would cause the value to be an object. Probably not the intended behavior. Is something like this better than the current version?

```js
const startValue = is.object(range.start) ? range.start.value : range.start;
const endValue = is.object(range.end) ? range.end.value : range.end;
```

src/table.js
kolodny: Cool, calling `.unpipe()` just to be sure.

src/table.js
```diff
@@ -35,7 +35,12 @@ const ChunkTransformer = require('./chunktransformer.js');
 
 // See protos/google/rpc/code.proto
 // (4=DEADLINE_EXCEEDED, 10=ABORTED, 14=UNAVAILABLE)
-const RETRY_STATUS_CODES = new Set([4, 10, 14]);
+const MUTATE_ROWS_RETRYABLE_STATUS_CODES = new Set([4, 10, 14]);
+
+// TODO - DO NOT MERGE - Figure out exactly what codes need to be here. Some combination of
```
kolodny: @sduskis Any input for this?

src/table.js
stephenplusplus: Sure, that LGTM.

src/table.js
sduskis: The Java client uses the same statuses for MutateRows and ReadRows.  I'm having a bit of trouble figuring out how the HTTP codes translate... Why can't we just use gRPC status codes?

kolodny: [Because the commonGRPC library "helpfully" converts the grpc code back to it's HTTP equivalent](https://github.com/googleapis/nodejs-common-grpc/blob/master/src/service.js#L570)

kolodny:
@stephenplusplus Were there any other concerns or comments you had? If not I think we're good to merge

PR #29, opened 2018-01-09
Fix #28 table.row().get() dont accept filter in options

### LINKED ISSUE #28: table.row().get()  dont accept filter in options
table.row().get() overwrites filter created out of first argument ignoring filter provided as options. This limits the filter options can be passed to get row request.

ajaaym:
Fixes #28 

- [x] Tests and linter pass

### DIFF
--- a/test/row.js
+++ b/test/row.js
@@ -900,6 +900,22 @@ describe('Bigtable/Row', function() {
 
       row.get(keys, options, assert.ifError);
     });
+    it('should respect filter in options object', function(done) {
+      var keys = [];
+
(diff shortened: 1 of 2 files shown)


test/row.js
```diff
@@ -906,6 +906,22 @@ describe('Bigtable/Row', function() {
 
       row.get(keys, options, assert.ifError);
     });
+    it('should respect filter in options object', function(done) {
```
stephenplusplus: Please add a newline to separate these test blocks consistently.

PR #35, opened 2018-01-31
Convert to GAPIC

### LINKED ISSUE #23: Convert API to use GAPIC
Continuing from Bigtable [Convert grpc APIs to use GAPIC](https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1859), Bigtable is the only API left to complete these steps:

- [ ] Bigtable
  - [ ] Generated files are in repo
  - [ ] Handwritten API is using generated API
  - [ ] Passing through `gaxOptions` is available on all requests
  - [x] Automatic projectId insertion works

stephenplusplus:
Fixes #23 

## Blockers

- https://github.com/googleapis/nodejs-bigtable/pull/35/files#r170264277

## Breaking Changes

```diff
- table.createFamily(name, rule)
+ table.createFamily(name, {
+   rule: rule,
+   gaxOptions: gaxOptions
+ })

- family.create(rule)
+ family.create({
+   rule: rule,
+   gaxOptions: gaxOptions
+ })

- row.create({ property: true })
+ row.create({
+   entry: {
+     property: true
+   },
+   gaxOptions: gaxOptions
+ })

- row.filter(filter, [onMatchEntries], [onNoMatchEntries])
+ row.filter(filter, {
+   onMatch: [onMatchEntries],
+   onNoMatch: [onNoMatchEntries],
+   gaxOptions: gaxOptions
+ })

- row.save('follows:jadams', 1)
+ row.save({
+   follows: {
+     jadams: 1
+   }
+ }, gaxOptions)

- bigtable.getInstances([ListInstancesRequest], callback)
+ bigtable.getInstances([gaxOptions], callback)

- bigtable.getInstancesStream()

- instance.getClusters([ListClustersRequest], callback)
+ instance.getClusters([gaxOptions], callback)

- instance.getClustersStream()
```

## Todos

- [x] Use GAPIC autopagination for all list methods (https://github.com/googleapis/nodejs-bigtable/pull/35/files#r170264277)
- [x] system tests
- index.js
  - [x] Make requests through GAPIC
  - [x] Expose `gaxOptions`
  - [x] Document `gaxOptions`
  - [x] Unit test
  - [x] Support emulator and API endpoint overrides
- cluster.js
  - [x] Make requests through GAPIC
  - [x] Expose `gaxOptions`
  - [x] Document `gaxOptions`
  - [x] Unit test
- family.js
  - [x] Make requests through GAPIC
  - [x] Expose `gaxOptions`
  - [x] Document `gaxOptions`
  - [x] Unit test
- instance.js
  - [x] Make requests through GAPIC
  - [x] Expose `gaxOptions`
  - [x] Document `gaxOptions`
  - [x] Unit test
- row.js
  - [x] Make requests through GAPIC
  - [x] Expose `gaxOptions`
  - [x] Document `gaxOptions`
  - [x] Unit test
- table.js
  - [x] Make requests through GAPIC
  - [x] Expose `gaxOptions`
  - [x] Document `gaxOptions`
  - [x] Unit test
  - [x] Figure out retry logic switch from retry-request to GAPIC


stephenplusplus:
@kolodny this is still a work in progress, but would you mind trying to run the "rows" system tests in `system-test/bigtable.js`? I currently have them skipped, but you can replace `describe.skip` with `describe.only` to isolate them.

The problem seems to be that where we received a string version of a Buffer before, we now receive a Buffer directly from GAPIC. For example, this is a row that is handled by ChunkTransformer, before and after:

```js
// Before
{ chunks:
   [ { row_status: null,
       rowKey: 'YWxpbmNvbG4=',
       familyName: [Object],
       qualifier: [Object],
       timestampMicros: '1518711386138000',
       labels: [],
       value: 'AAAAAAAAAAE=',
       valueSize: 0,
       resetRow: false,
       commitRow: false },
     { row_status: null,
       rowKey: '',
       familyName: null,
       qualifier: [Object],
       timestampMicros: '1518711386138000',
       labels: [],
       value: 'AAAAAAAAAAE=',
       valueSize: 0,
       resetRow: false,
       commitRow: false },
     { row_status: 'commitRow',
       rowKey: '',
       familyName: null,
       qualifier: [Object],
       timestampMicros: '1518711386138000',
       labels: [],
       value: 'AAAAAAAAAAE=',
       valueSize: 0,
       resetRow: false,
       commitRow: true } ],
  lastScannedRowKey: '' }

// After
{ chunks:
   [ { labels: [],
       rowKey: <Buffer 61 6c 69 6e 63 6f 6c 6e>,
       familyName: [Object],
       qualifier: [Object],
       timestampMicros: '1518711494344000',
       value: <Buffer 00 00 00 00 00 00 00 01>,
       valueSize: 0 },
     { labels: [],
       rowKey: [],
       familyName: null,
       qualifier: [Object],
       timestampMicros: '1518711494344000',
       value: <Buffer 00 00 00 00 00 00 00 01>,
       valueSize: 0 },
     { labels: [],
       rowKey: [],
       familyName: null,
       qualifier: [Object],
       timestampMicros: '1518711494344000',
       value: <Buffer 00 00 00 00 00 00 00 01>,
       valueSize: 0,
       commitRow: true,
       rowStatus: 'commitRow' } ],
  lastScannedRowKey: [] }
```

There are lots of methods that seem to handle converting between strings and Buffers, and I'm wondering if these are still necessary, or if this new response type is problematic for other reasons.

Please take a look and feel free to make any corrections necessary. Your help will surely be quicker and more accurate. Let me know if you have any questions about the new GAPIC layer or run into other issues.

sduskis:
@ajaaym, can you also please take a look?

src/table.js
```diff
@@ -1053,13 +1141,12 @@ Table.prototype.mutate = function(entries, callback) {
...
 
   function makeNextBatchRequest() {
-    var grpcOpts = {
-      service: 'Bigtable',
-      method: 'mutateRows',
-      retryOpts: {
-        currentRetryAttempt: numRequestsMade,
-      },
-    };
+    // @todo Figure out how to do this in gapic.
```
stephenplusplus: @kolodny Just flagging this, as I'm not currently sure how to handle this. GAPIC has its own retry logic built in: https://github.com/googleapis/gax-nodejs/blob/ec0c8b0805c31d8a91ea69cb19fe50f42a38bf87/lib/gax.js#L279. Are these options enough to configure the desired behavior?

src/v2/bigtable_instance_admin_client.js
```diff
@@ -127,6 +127,11 @@ class BigtableInstanceAdminClient {
         'nextPageToken',
         'appProfiles'
       ),
+      listClusters: new gax.PageDescriptor(
```
stephenplusplus: @alexander-fenster  I had to manually add this to enable the auto-pagination behavior that other `list` endpoints have.

src/table.js
kolodny: This looks tricky to work with as is. There's always the options of not using retries from gax and just wrapping the API in our own but that should be avoided at all costs.

My main concern is what happens when gax thinks the call was successful yet there was an error with a couple mutations out of an entire batch. Consider this (parens are timeline for step):

1. Start mutation call (0ms)
1. Can't connect to service (0ms to 100ms)
1. Wait 500ms (100ms to 600ms)
1. Some mutations failed (600ms to 650ms)
1. Wait 2500ms (650ms to 3150ms)
1. Can't connect to service (3150ms to 3250ms)
1. Wait 4500ms (3250ms to 7750ms)

The tricky bit will be the wait times of step 7 after step 5. We can have an initialRetryDelayMillis but then the multiplier will be out of sync with how the retries should behave for the remaining calls. We can either go through the pass current retry attempt dance that we did for common-grpc/retry-request or just roll our own.

@stephenplusplus @sduskis thoughts?

src/v2/bigtable_instance_admin_client.js
alexander-fenster: this is the second time we edit something in the generated code (the first case was in pubsub here https://github.com/googleapis/nodejs-pubsub/blob/d0dc20ceddf6a897d5143d5342ea1d084d0da57d/src/v1/subscriber_client.js#L225) and we might face problems if we continue doing it. At least we need to make sure that unit tests will fail if this is not defined here (in case if someone regenerates and overwrites the code).

stephenplusplus:
I definitely don’t want to leave this here. Can we instead fix the
generator?

src/v2/bigtable_instance_admin_client.js
stephenplusplus: I definitely didn't want to leave this here. Can we instead fix the generator?

src/table.js
stephenplusplus: > There's always the options of not using retries from gax and just wrapping the API in our own but that should be avoided at all costs.

It might not actually be that bad from a performance overhead perspective. From a "consolidating dependencies" perspective, yeah, this would deduct a few points.

Also, a new discovery: https://github.com/googleapis/gax-nodejs/blob/ec0c8b0805c31d8a91ea69cb19fe50f42a38bf87/lib/streaming.js#L230

> // Right now retrying does not work with gRPC-streaming, because retryable
> // assumes an API call returns an event emitter while gRPC-streaming methods
> // return Stream.
> // TODO: support retrying

Considering this, I've updated this PR to use retry-request.

stephenplusplus:
I've updated the PR as far as I think I can go without some help:

- @kolodny Please verify retry logic is working. The system-test/mutate-rows.js file is hard to work with, since it is faking timers. Could someone please take a look? I've updated it to override the new request methods, but it's still having trouble.

- @alexander-fenster The generated file had to be changed. Can we fix this in the generator? (https://github.com/googleapis/nodejs-bigtable/pull/35#discussion_r170264277)

kolodny:
@stephenplusplus I believe you needed to fake out slightly more than just setTimeout. To be honest, I'm not sure why the read-rows work. I have a proof of concept at https://github.com/kolodny/nodejs-bigtable/commit/eaff5c3432948da19fc7e3bef81f7515f4574495

I made the patching a little more aggressive but I'm not sure that's the best approach either. We can always completely remove the fake timers but we'd need to set the mocha timeout to at least a minute to take into account the exponential backoffs.

Let me know if that unblocks the system tests.

kolodny:
Sure https://github.com/googleapis/nodejs-bigtable/compare/f6d6dea...kolodny:pr-35 has a cross platform patching mechanism. It just replaces the file (wherever it finds it) with the patched version

kolodny:
Patching is probably the wrong term for it now since we're overwriting the file completely, so feel free to rename it to "fixed" or something similar.

stephenplusplus:
Thanks for doing that. I've tried applying the changes from that commit, but end up with 102 errors from the system test run that look like:

```
  102) Bigtable/Table
       createReadStream
         does the previous 5 things in one giant test case:
     TypeError: processNextTick is not a function
      at resume (node_modules\readable-stream\lib\_stream_readable.js:783:5)
      at DestroyableTransform.Readable.resume (node_modules\readable-stream\lib\_stream_readable.js:775:5)
      at DestroyableTransform.Readable.on (node_modules\readable-stream\lib\_stream_readable.js:745:53)
      at DestroyableTransform.Readable.pipe (node_modules\readable-stream\lib\_stream_readable.js:615:7)
      at pipe (node_modules\pump\index.js:56:15)
      at Array.reduce (<anonymous>)
      at pump (node_modules\pump\index.js:79:11)
      at Pumpify.setPipeline (node_modules\pumpify\index.js:39:5)
      at new Pumpify (node_modules\pumpify\index.js:15:30)
      at Function.Pumpify [as obj] (node_modules\pumpify\index.js:13:44)
      at makeNewRequest (src\table.js:511:31)
      at Table.createReadStream (src\table.js:540:3)
      at Context.it (system-test\read-rows.js:128:15)
```

I've confirmed that the patched file does overwrite the originals inside the node_modules directory.

kolodny:
Strange it works for me. What is `node_modules\readable-stream\lib\_stream_readable.js` line 783 ? Is it. Is is `processNextTick(resume_, stream, state);` if so does this line exist in that file too:

```
var processNextTick = require('process-nextick-args');
```

And if that is also found, can you try just doing something simple with `process-nextick-args` like running a script as follows:

```js
var processNextTick = require('process-nextick-args');
processNextTick(console.log, 'test');
```

stephenplusplus:
>  Is is processNextTick(resume_, stream, state);

Yep.

> if so does this line exist in that file too:
> var processNextTick = require('process-nextick-args');

Yep.

> can you try just doing something simple with process-nextick-args like running a script

```js
> require('process-nextick-args')(console.log, 'test')
undefined
> test
```

I noticed that the top of _stream_readable.js requires:

```js
var processNextTick = require('process-nextick-args').nextTick;
```

But our patch doesn't export `nextTick`. My patched file for `process-nextick-args/index.js` is:

```js
module.exports = function() {
  return process.nextTick.apply(this, arguments);
};
```

stephenplusplus:
Oh sorry, maybe I answered question 2 incorrectly. It does require the module, but it looks for the `nextTick` property.

kolodny:
Interesting, it looks like the `process-nextick-args` fix may have finally been published. Can you try deleting your node_modules folder and try again once again. If that fails, try deleting the node_modules and patch folders, remove the `presystem-test` line in package.json and try again

stephenplusplus:
I deleted the patching pieces, although I can't be sure that was a good move. The system tests began to pass again, up until `system-test/mutate-rows.js`. They simply end abruptly without any error output. What happens locally when you `rm -rf node_modules`, npm install, and `npm run system-test`?

stephenplusplus:
I had to re-work the script a bit to work, but the same early exiting without any output happens for me.

revised script:
```js
const fs = require('fs');
const glob = require('glob');
const path = require('path');

const patched = `${__dirname}/patched-process-nextick-args.js`;
glob.sync('./**/process-nextick-args').forEach(pnaDirectoryPath => {
  const packageJson = require(path.join('../', pnaDirectoryPath, 'package.json'));
  const majorVersion = parseInt(packageJson.version[0], 10);

  if (majorVersion <= 1) {
    fs.copyFileSync(patched, path.join(pnaDirectoryPath, 'index.js'));
  }
});
```

kolodny:
There's probably some error happening in the assert which is in async mode for mocha, try using a pattern like this:

```js
// OLDWAY
      it(test.name, done => {
        responses = test.responses;
        TABLE.maxRetries = test.max_retries;
        TABLE.mutate(test.mutations_request, error => {
          assert.deepEqual(
            mutationBatchesInvoked,
            test.mutation_batches_invoked
          );
          getDeltas(mutationCallTimes).forEach((delta, index) => {
            if (index === 0) {
              const message = 'First request should happen Immediately';
              assert.strictEqual(index, 0, message);
              return;
            }
            const minBackoff = 1000 * Math.pow(2, index);

            // Adjust for some flakiness with the fake timers.
            const maxBackoff = minBackoff + 1010;
            const message =
              `Backoff for retry #${index} should be between ` +
              `${minBackoff} and ${maxBackoff}, was ${delta}`;
            assert(delta > minBackoff, message);
            assert(delta < maxBackoff, message);
          });
          if (test.errors) {
            const expectedIndices = test.errors.map(error => {
              return error.index_in_mutations_request;
            });
            const actualIndices = error.errors.map(error => {
              return test.mutations_request.indexOf(error.entry);
            });
            assert.deepEqual(expectedIndices, actualIndices);
          } else {
            assert.ifError(error);
          }
          done();
        });
        clock.runAll();
      });
    });
```

to this:

```js
      it(test.name, done => {
        responses = test.responses;
        TABLE.maxRetries = test.max_retries;
        TABLE.mutate(test.mutations_request, error => {
          try {
            assert.deepEqual(
              mutationBatchesInvoked,
              test.mutation_batches_invoked
            );
            getDeltas(mutationCallTimes).forEach((delta, index) => {
              if (index === 0) {
                const message = 'First request should happen Immediately';
                assert.strictEqual(index, 0, message);
                return;
              }
              const minBackoff = 1000 * Math.pow(2, index);

              // Adjust for some flakiness with the fake timers.
              const maxBackoff = minBackoff + 1010;
              const message =
                `Backoff for retry #${index} should be between ` +
                `${minBackoff} and ${maxBackoff}, was ${delta}`;
              assert(delta > minBackoff, message);
              assert(delta < maxBackoff, message);
            });
            if (test.errors) {
              const expectedIndices = test.errors.map(error => {
                return error.index_in_mutations_request;
              });
              const actualIndices = error.errors.map(error => {
                return test.mutations_request.indexOf(error.entry);
              });
              assert.deepEqual(expectedIndices, actualIndices);
            } else {
              assert.ifError(error);
            }
            done();
          } catch (error) {
            done(error);
          }
        });
        clock.runAll();
      });
    });
```

Note the try catch block in the `TABLE.mutate` callback. You should still get a failing test but now you can see what the failing test is

stephenplusplus:
Tried that, but still no output. Take a look at the console.logs I added to the test here: https://github.com/googleapis/nodejs-bigtable/pull/35/commits/7b8580066e53d9f40e91a72ea8bc53394c23b98c

kolodny:
hmm, this is what I get:

```
Bigtable/Table
    mutate()
Executes
Never executes
Never executes
Never executes
      ✓ valid mutation
Never executes
Executes
Never executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
      ✓ retries the failed mutations
Never executes
Executes
Never executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
      ✓ has a `PartialFailureError` error when an entry fails after the retries
Never executes
Executes
Never executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
      ✓ does not retry unretryable mutations
Never executes
Executes
Never executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
Never executes
Executes
Never executes
Never executes
Never executes
      ✓ considers network errors towards the retry count
Never executes


  5 passing (48ms)

run: Success! Test finished in 1.03s
```

This may be related to something weird happening on Windows.

stephenplusplus:
I would believe that. I've pushed a commit that will hopefully get Appveyor to run the test.

kolodny:
This is really bizarre, it still works fine on my macbookpro. Try changing the fake timer code to mock out every possible async function:

```js
      // OLD
      clock = sinon.useFakeTimers({
        toFake: [
          'setTimeout',
          'setImmediate',
          'Date',
          'nextTick',
        ],
      });
```

To this

```js
      // NEW
      clock = sinon.useFakeTimers({
        toFake: [
          'setTimeout',
          'clearTimeout',
          'setImmediate',
          'clearImmediate',
          'setInterval',
          'clearInterval',
          'Date',
          'nextTick',
        ],
      });
```

stephenplusplus:
Same problem, sadly: 

```
  Bigtable/Table
    mutate()
Executes
run: Success! Test finished in 1.58s
```

One thing to just make sure you're on the latest dependencies, try a hard delete of node_modules, and before npm installing again, you can clear the npm cache: `npm cache clean -f`

kolodny:
Just tried that and it still works for me. I guess we can always just nuke the fake timers code for now and see if that helps. Just remove all references to the `clock` variable.

kolodny:
Wow, OK I found what's causing the weirdness on my system. I had an old packae-lock.json file. When I removed it, I can replicate the behavior. Let me also see what I can do to fix this bug

kolodny:
😠 https://github.com/nodejs/readable-stream/blob/f94381cc3f0f8e91672c7695e0a63a1ee05c8e92/lib/_stream_readable.js#L26

```js
var processNextTick = require('process-nextick-args').nextTick;
```

The whole point of that change was to make nextTick a "live" property. This is a bug in readable-stream, I'll open a issue in that repo to fix this. In the meantime I'll update the patch code to work with v1 or v2 of process-arg-nextick

kolodny:
OK. I opened https://github.com/nodejs/readable-stream/issues/323 to track the readable-stream issue. I have a quick and dirty fix at https://github.com/kolodny/nodejs-bigtable/commit/a6d0d19e655cda6b26bd4668cc1ccddcaf595184

Let me know if that works.

stephenplusplus:
It worked! I've pushed a commit: https://github.com/googleapis/nodejs-bigtable/pull/35/commits/46ce90cc91d60d73333572c260c25f5e4844e52e

sduskis:
Do we have any more blockers at this point?

stephenplusplus:
Just the generated file having a mistake in it. (@alexander-fenster)

src/v2/bigtable_instance_admin_client.js
alexander-fenster: Took a closer look - it must be fixed in [here](https://github.com/googleapis/googleapis/blob/d90956b8adf6c820969457d848db833328076879/google/bigtable/admin/v2/bigtableadmin_gapic.yaml#L200), not in the generator. Working on it.

alexander-fenster: @stephenplusplus, I just got an explanation from bigtable owners (@igorbernstein2). Neither listClusters nor listInstances should be paginaged, and the page tokens for these methods will be removed in v3. The reasoning is: (a) pagination of clusters is token is complicated on the server side, and (b) these list will never grow too large. So the pagination code is not generated here for a reason.

stephenplusplus: Alright, fixing now.

stephenplusplus:
Updated to remove the pagination behavior from bigtable.getInstances() and instance.getClusters(). See the additions to the diff section from my first post.

This should be ready to go. ✈️

src/index.js
```diff
@@ -337,58 +339,88 @@ function Bigtable(options) {
...
-      Bigtable: {
-        baseUrl: baseUrl,
-        path: 'google/bigtable/v2/bigtable.proto',
-        service: 'bigtable.v2',
-      },
-      BigtableTableAdmin: {
-        baseUrl: adminBaseUrl,
-        path: 'google/bigtable/admin/v2/bigtable_table_admin.proto',
-        service: 'bigtable.admin.v2',
+  var options_ = extend(
```
kolodny: This should probably have better naming, for example userOptions, defaultOptions, and options

src/instance.js
```diff
@@ -553,110 +518,113 @@ Instance.prototype.cluster = function(name) {
...
+  this.bigtable.request(
+    {
+      client: 'BigtableInstanceAdminClient',
+      method: 'listClusters',
+      reqOpts: reqOpts,
+      gaxOpts: gaxOptions,
+    },
+    function(err, resp) {
+      if (err) {
+        callback(err);
```
kolodny: What determines when we pass additional data with the error?

src/row.js
```diff
@@ -553,14 +451,173 @@ Row.prototype.delete = function(callback) {
...
+ */
+Row.prototype.exists = function(gaxOptions, callback) {
+  if (is.fn(gaxOptions)) {
+    callback = gaxOptions;
+    gaxOptions = {};
+  }
+
+  this.getMetadata(gaxOptions, function(err) {
+    if (err) {
+      if (err instanceof RowError) {
```
kolodny: I'm curious as to why this is sometimes `error.name === 'ThingError'` sometimes `err.code === 5`, and sometimes `err instanceof RowError`.

src/table.js
```diff
@@ -1036,8 +1136,8 @@ Table.prototype.mutate = function(entries, callback) {
   var entryToIndex = new Map(entries.map((entry, index) => [entry, index]));
   var mutationErrorsByEntryIndex = new Map();
 
-  function onBatchResponse(previousNumRequestsMade, err) {
-    if (previousNumRequestsMade === numRequestsMade && err) {
```
kolodny: Was this intentionally  reverted?

src/row.js
```diff
@@ -275,21 +249,21 @@ Row.formatFamilies_ = function(families, options) {
  *   var apiResponse = data[0];
  * });
  */
-Row.prototype.create = function(entry, callback) {
+Row.prototype.create = function(options, callback) {
```
kolodny: Should we allow users to use the old `row.create(rowData)` syntax?

test/filter.js
```diff
@@ -42,8 +42,7 @@ describe('Bigtable/Filter', function() {
   });
 
   afterEach(function() {
-    sinon.restore();
```
kolodny: We should probably start using http://sinonjs.org/releases/v4.4.4/sandbox/ on the afterEach since the world won't be in a broken state if an assert fails

src/cluster.js
```diff
@@ -249,15 +98,214 @@ Cluster.getStorageType_ = function(type) {
...
+ *     .on('error', console.error)
+ *     .on('complete', function() {
+ *       // The cluster was created successfully.
+ *     });
+ * });
+ *
+ * //-
+ * // If the callback is omitted, we'll return a Promise.
+ * //-
+ * cluster.create().then(function(data) {
```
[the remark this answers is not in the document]
stephenplusplus: This has a long history :) https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2556

src/instance.js
stephenplusplus: The GAPIC layer only returns `err`, so we only give it back as such. If there are places that say they return more, it's probably a mistake.

sduskis:
LGTM as well.  It looks like `table.createFamily()`'s parameters changed a bit, and there might be some other changes lurking.  I'm ok with the changes being committed, as long as we document them in some sort of release notes.

src/row.js
stephenplusplus: I found one `err.name === 'FamilyError'` check, and I changed that to `err instanceof FamilyError`. Were there others you found?

For this line we're commenting on (`err instanceof RowError`), when a row doesn't exist, that's the error our API returns from `row.get()`, which is called by `this.getMetadata()`.

For the other places, the code 5 is what GAPIC returns, a raw GRPC error object.

kolodny: I haven't found any others. Makes sense

src/row.js
stephenplusplus: This would prevent the `gaxOptions` argument, because we can't have two consecutive, optional object arguments. I'm fine with not allowing `gaxOptions` here, although I believe that would be our first deviation, and could come back to bite us if users find they needed that type of control.

stephenplusplus:
@sduskis all of the breaking changes are noted in the opening post (https://github.com/googleapis/nodejs-bigtable/pull/35#issue-166394887) -- let me know if any of those look like a problem.

src/row.js
kolodny: Oh I meant doing some testing to see if options is a rowData object then treating it like this:

```js
if (isRowData(options) {
  options = {entry: options}
}
```

I'm not sure how to implement `isRowData` though. Something similar can also be done for other methods

src/cluster.js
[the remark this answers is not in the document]
kolodny: Thanks for that context, I left a comment on a possible solution https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2556#issuecomment-372682191

That should probably be addressed since (IMO) it would drastically improve the developer experience

src/table.js
stephenplusplus: Yes, although feel free to correct me on this. There was a failing test, and I couldn't understand exactly what was happening (both why we needed the code that I removed & why the test was failing). I attributed it to how we were mocking the endpoints in the test, and now that we switched to GAPIC, the extra code wasn't needed anymore.

test/filter.js
stephenplusplus: Hmm, I'm not sure why I removed this line, since it looks like it's still valid. I'll put it back -- does that fix what you were calling out, or is there more to do?

src/table.js
kolodny: The issue at hand was that if an error happened without a `request` ever being emitted there would be an infinite (albeit async) infinite loop. This check was a guard against it. The async nature of this method is hiding the fact that it's a glorified `while (true)` loop

test/filter.js
kolodny: I think it's still worth using sandboxes since you need to remember to call restore for each spy. It doesn't need to happen right now, but if a test fails in the future it can cascade to every other test after it

src/table.js
stephenplusplus: I see now, thanks. Putting it back.

src/row.js
stephenplusplus: > I'm not sure how to implement isRowData though.

I think that's the problem I meant to illustrate as well, since both options and entries would be optional POJOs. If we special-case `gaxOptions` and `entry`, then we would be blocking row data where they meant to use those property names as part of their entry:

```js
row.create({ name: '1b', entry: true })
// isRowData() returns false
```

test/filter.js
stephenplusplus: Earlier in the file, `sinon` is assigned as:

```js
var sinon = require('sinon').sandbox.create();
```

Maybe I don't need the `FakeMutation.convertToBytes.resetHistory();` in the afterEach now?

src/row.js
kolodny: Got it, makes sense

test/filter.js
kolodny: Whoops missed that bit, reverting this line should be all that's needed then

stephenplusplus: Great, here's the new afterEach:

```js
afterEach(function() {
  sinon.restore();
});
```

system-test/read-rows.js
```diff
@@ -131,20 +130,23 @@ describe('Bigtable/Table', () => {
...
-        }
-        assert.deepStrictEqual(rowKeysRead, test.row_keys_read);
-        assert.strictEqual(
-          responses.length,
-          0,
-          'not all the responses were used'
-        );
-        assert.deepStrictEqual(requestedOptions, test.request_options);
+
+        setImmediate(function() {
```
kolodny: I'm missed this in the initial review, but putting this in a setImmediate after the call to `clock.runAll()` doesn't actually run this block of code in the test. It looks like read row retries are not working as expected since we're now getting grpc error codes and not http error code (since we're not wrapped by a GRPC service layer). I'll enable this block of code and send a proper PR to fix reties soon

PR #64, opened 2018-03-16
Fix read retires.

kolodny:
Fix broken read row retries

- [x] Tests and linter pass

I used the following to generate some heavy rows:

```js
const bigtable = require('./');

const TABLE_NAME = 'Hello-Bigtable-my-perf';
const COLUMN_FAMILY_NAME = 'cf1';
const COLUMN_NAME = 'greeting';
const INSTANCE_ID = 'helloworld';

const measure = title => {
  console.log(`measuring task: ${title}`);
  console.time(title);
  return () => console.timeEnd(title);
}

let done;
(async () => {
  try {
    const bigtableClient = bigtable();
    const instance = bigtableClient.instance(INSTANCE_ID);

    const table = instance.table(TABLE_NAME);
    const [tableExists] = await table.exists();
    if (tableExists) {
      done = measure('Table already exists, delete the table');
      await table.delete();
      done();
    }
    const options = {
      families: [COLUMN_FAMILY_NAME],
    };
    done = measure('Creating table');
    await table.create(options);
    done()

    console.log('begin inserting rows');

    const bytes =
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' +
      '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef'


    let rowsToInsert = [];
    for (let index = 0; index < 1000000; index++) {
      if (index != 0 && index % 100000 === 0) {
        done = measure('inserting 100000 rows');
        await table.insert(rowsToInsert);
        done()
        rowsToInsert = [];
      }
      rowsToInsert.push({
        key: `row${index}`,
        data: {
          [COLUMN_FAMILY_NAME]: {
            [COLUMN_NAME]: bytes,
          },
        },
      });
    }
    done = measure('inserting 100000 rows');
    await table.insert(rowsToInsert);
    done();

  } catch (error) {
    console.error('Something went wrong:', error);
  }
})();
```

And the following to read it:

```js
const bigtable = require('./');
const throttle = require('lodash').throttle;

const TABLE_NAME = 'Hello-Bigtable-my-perf';
const COLUMN_FAMILY_NAME = 'cf1';
const COLUMN_NAME = 'greeting';
const INSTANCE_ID = 'helloworld';

const log = throttle(msg => console.log(msg), 1000, { leading: true });

(async () => {
  try {
    const bigtableClient = bigtable();
    const instance = bigtableClient.instance(INSTANCE_ID);

    const table = instance.table(TABLE_NAME);
    let index = 0;
    table.createReadStream().on('data', row => {
      index++
      if (index % 10000 === 0) console.log('got row ' + row.id)
      // log('Got row ' + row.id);
    })

  } catch (error) {
    console.error('Something went wrong:', error);
  }
})();

```

Before this change an error would not be retried, after this change it retries from the row it was up to

### DIFF
--- a/src/chunktransformer.js
+++ b/src/chunktransformer.js
@@ -87,8 +87,8 @@ ChunkTransformer.prototype._transform = function(data, enc, next) {
       return;
     }
   }
-  if (data.lastScannedRowKey) {
-    this.lastRowKey = data.lastScannedRowKey;
+  if (data.lastScannedRowKey && data.lastScannedRowKey.length > 0) {
+    this.lastRowKey = Mutation.convertFromBytes(data.lastScannedRowKey);
   }
   next();
 };
--- a/src/table.js
+++ b/src/table.js
@@ -34,10 +34,7 @@ const ChunkTransformer = require('./chunktransformer.js');
 
 // See protos/google/rpc/code.proto
 // (4=DEADLINE_EXCEEDED, 10=ABORTED, 14=UNAVAILABLE)
-const GRPC_RETRYABLE_STATUS_CODES = new Set([4, 10, 14]);
-
-// (409=ABORTED, 503=UNAVAILABLE, 14=DEADLINE_EXCEEDED)
-const HTTP_RETRYABLE_STATUS_CODES = new Set([409, 503, 504]);
+const RETRYABLE_STATUS_CODES = new Set([4, 10, 14]);
 
 /**
  * Create a Table object to interact with a Cloud Bigtable table.
@@ -527,7 +524,7 @@ Table.prototype.createReadStream = function(options) {
       rowStream.unpipe(userStream);
       if (
         numRequestsMade <= maxRetries &&
-        HTTP_RETRYABLE_STATUS_CODES.has(error.code)
+        RETRYABLE_STATUS_CODES.has(error.code)
       ) {
         makeNewRequest();
       } else {
@@ -1200,7 +1197,7 @@ Table.prototype.mutate = function(entries, gaxOptions, callback) {
             return;
           }
 
-          if (!GRPC_RETRYABLE_STATUS_CODES.has(entry.status.code)) {
+          if (!RETRYABLE_STATUS_CODES.has(entry.status.code)) {
             pendingEntryIndices.delete(originalEntriesIndex);
           }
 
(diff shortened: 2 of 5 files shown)


src/chunktransformer.js
```diff
@@ -87,8 +87,8 @@ ChunkTransformer.prototype._transform = function(data, enc, next) {
       return;
     }
   }
-  if (data.lastScannedRowKey) {
-    this.lastRowKey = data.lastScannedRowKey;
+  if (data.lastScannedRowKey && data.lastScannedRowKey.length > 0) {
```
kolodny: [lastScannedRowKey is defined as bytes](https://github.com/googleapis/nodejs-bigtable/blob/8db7bf35f79de791dd5adbad55625048c79694ad/protos/google/bigtable/v2/bigtable.proto#L181) so it will always be set to at least an empty array even when the service didn't actually set it

src/table.js
```diff
@@ -527,7 +524,7 @@ Table.prototype.createReadStream = function(options) {
       rowStream.unpipe(userStream);
       if (
         numRequestsMade <= maxRetries &&
-        HTTP_RETRYABLE_STATUS_CODES.has(error.code)
+        RETRYABLE_STATUS_CODES.has(error.code)
```
kolodny: We're no longer using `grpc-common` so GRPC error codes are no longer converted to HTTP error codes

PR #74, opened 2018-03-29
Implemented appProfileId support to bigtable APIs

vijay-qlogic:
## Blocked by public feature release

Implement appProfileId support for APIs
- [ ] Tests and linter pass

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -421,7 +421,7 @@ function Bigtable(options) {
   this.api = {};
   this.auth = googleAuth(options);
   this.projectId = options.projectId || '{{projectId}}';
-  this.appProfileId = options.appProfileId || undefined;
+  this.appProfileId = options.appProfileId;
   this.projectName = 'projects/' + this.projectId;
 }
 
--- a/src/index.js
+++ b/src/index.js
@@ -421,7 +421,7 @@ function Bigtable(options) {
   this.api = {};
   this.auth = googleAuth(options);
   this.projectId = options.projectId || '{{projectId}}';
-  this.appProfileId = options.appProfileId;
+  this.appProfileId = options.appProfileId || undefined;
   this.projectName = 'projects/' + this.projectId;
 }
 
--- a/src/index.js
+++ b/src/index.js
@@ -47,6 +47,9 @@ const v2 = require('./v2');
  *     downloaded from the Google Developers Console. If you provide a path to a
  *     JSON file, the `projectId` option above is not necessary. NOTE: .pem and
  *     .p12 require you to specify the `email` option as well.
+ * @property {string} [appProfileId] An application profile ID, a configuration
+ *     string value describing how Cloud Bigtable should treat traffic from a
+ *     particular end user application.
  * @property {string} [email] Account email address. Required when using a .pem
  *     or .p12 keyFilename.
  * @property {object} [credentials] Credentials object.
@@ -418,6 +421,7 @@ function Bigtable(options) {
   this.api = {};
   this.auth = googleAuth(options);
   this.projectId = options.projectId || '{{projectId}}';
+  this.appProfileId = options.appProfileId || null;
   this.projectName = 'projects/' + this.projectId;
 }
 
--- a/src/index.js
+++ b/src/index.js
@@ -421,7 +421,7 @@ function Bigtable(options) {
   this.api = {};
   this.auth = googleAuth(options);
   this.projectId = options.projectId || '{{projectId}}';
-  this.appProfileId = options.appProfileId || null;
+  this.appProfileId = options.appProfileId;
   this.projectName = 'projects/' + this.projectId;
 }
 
--- a/src/row.js
+++ b/src/row.js
@@ -361,7 +361,7 @@ Row.prototype.createRules = function(rules, gaxOptions, callback) {
 
   var reqOpts = {
     tableName: this.table.id,
-    appProfileId: this.appProfileId,
+    appProfileId: this.bigtable.appProfileId,
     rowKey: Mutation.convertToBytes(this.id),
     rules: rules,
   };
@@ -591,7 +591,7 @@ Row.prototype.exists = function(gaxOptions, callback) {
 Row.prototype.filter = function(filter, config, callback) {
   var reqOpts = {
     tableName: this.table.id,
-    appProfileId: this.appProfileId,
+    appProfileId: this.bigtable.appProfileId,
     rowKey: Mutation.convertToBytes(this.id),
     predicateFilter: Filter.parse(filter),
     trueMutations: createFlatMutationsList(config.onMatch),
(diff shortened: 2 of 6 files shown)


src/row.js
```diff
@@ -285,6 +285,7 @@ Row.prototype.create = function(options, callback) {
  * @param {object|object[]} rules The rules to apply to this row.
  * @param {object} [gaxOptions] Request configuration options, outlined here:
```
stephenplusplus: We should convert this to an `options` object:

- [options]
- [options.gaxOptions]
- [options.appProfileId]

stephenplusplus:
Are there any docs about what an app profile ID is?

stephenplusplus:
@alexander-fenster it looks like Waffle is getting sloppy. I'm not sure what happened at first:

"@wafflebot wafflebot bot added the cla: yes  label 2 hours ago"

But when I tried to move the card for this PR from "Backlog" to "PRs" (https://waffle.io/GoogleCloudPlatform/google-cloud-node), it tried adding the `cla: yes` label. Why would it even try to do that? Did something change in the googlebot code?

sduskis:
@stephenplusplus: [This is the proto docs for appProfile](https://github.com/googleapis/googleapis/blob/master/google/bigtable/admin/v2/instance.proto#L156).

stephenplusplus:
I believe @lukesneeringer and @alexander-fenster wanted to try to keep whitelist features away from public releases-- did I get that wrong, guys?

sduskis:
@stephenplusplus, I'd like to get these features out in all languages before the launch.  The features will be available in the service in the very near future.

stephenplusplus:
Some general comments that apply throughout all of the changes:

- `gaxOptions` should not be merged with other properties. Where this is currently `gaxOptions`, replace with an `options` object with properties: `options.gaxOptions` and `options.appProfileId`.
- Check for grammar-- "Application" typos and if we can, please use sentences to describe the properties, such as "{string} ... An application profile ID, used to: ____"

And a question-- can the user provide the ID from the Bigtable constructor? Then, it would be applied to all API requests, e.g.:

```js
const bigtable = new Bigtable({ appProfileId: '...' })
```

sduskis:
I'm ok with a constructor parameter instead of a per-method option for app-profile-id.

vijay-qlogic:
@stephenplusplus @sduskis 
Alright, if we can accept appProfileId as constructor parameter, 'll do below updates to the PR
- update comments for Bigtable class, including appProfileId property
- remove appProfileId from method arguments, current changes
- update reqOpts for methods, using appProfileId from object property(this.appProfileId)
- update test cases accordingly

vijay-qlogic:
@stephenplusplus @sduskis 
Updated PR to accept `appProfileId` in Bigtable constructor

src/index.js
```diff
@@ -418,6 +421,7 @@ function Bigtable(options) {
   this.api = {};
   this.auth = googleAuth(options);
   this.projectId = options.projectId || '{{projectId}}';
+  this.appProfileId = options.appProfileId || null;
```
stephenplusplus: Just `this.appProfileId = options.appProfileId` is fine. If the user didn't provide it, it will resolve it to `undefined`, which will be handled the same as `null`.

vijay-qlogic: ok, updating that

stephenplusplus:
Thanks @vijay-qlogic! Is `appProfileId` a valid property for all requests? If so, we can just assign it in the "request" function, which catches all outgoing requests: https://github.com/vijay-qlogic/nodejs-bigtable/blob/8b8e9176472ec20960eef957059fa9721758a0aa/src/index.js#L661

src/row.js
```diff
@@ -361,6 +361,7 @@ Row.prototype.createRules = function(rules, gaxOptions, callback) {
 
   var reqOpts = {
     tableName: this.table.id,
+    appProfileId: this.appProfileId,
```
stephenplusplus: I think we will need to locate `appProfileId` from `this.bigtable.appProfileId`.

src/row.js
```diff
@@ -590,6 +591,7 @@ Row.prototype.exists = function(gaxOptions, callback) {
 Row.prototype.filter = function(filter, config, callback) {
   var reqOpts = {
     tableName: this.table.id,
+    appProfileId: this.appProfileId,
```
stephenplusplus: I think we will need to locate `appProfileId` from `this.bigtable.appProfileId`.

vijay-qlogic:
Corrected `this.appProfileId` to` this.bigtable.appProfileId `where ever used.

Not sure on  "_Is `appProfileId` a valid property for all requests_", so as to assign the `appProfileId` from "Bigtable.prototype.request"

I have used `appProfileId` only for methods whose proto accepts `appProfileId` for example [here](https://github.com/googleapis/nodejs-bigtable/blob/master/protos/google/bigtable/v2/bigtable.proto#L80)

test/table.js
```diff
@@ -1438,6 +1438,19 @@ describe('Bigtable/Table', function() {
           done();
         });
       });
+
+      it('should make the correct request', function(done) {
```
stephenplusplus: `getRows` doesn't actually make this request, so this test is misplaced. We should only have tests under the `describe` block that matches the method's name.

test/table.js
```diff
@@ -1759,6 +1772,18 @@ describe('Bigtable/Table', function() {
         });
       });
     });
+
+    it('should use an appProfileId', function(done) {
```
stephenplusplus: This test looks perfect 👍 Could you just bump it up a bit, so that it is directly after the test called `'should provide the proper request options'`?

test/table.js
```diff
@@ -1854,6 +1879,18 @@ describe('Bigtable/Table', function() {
...
```
stephenplusplus: This test is misplaced as well. Instead, it should live under `sampleRowKeysStream`, as that is the function that was modified. The new test can go right after:

```js
  describe('sampleRowKeysStream', function() {
    it('should provide the proper request options', function(done) {
      // ...
    });

    // here
```

vijay-qlogic:
@stephenplusplus 
Thanks! for reviewing the code changes, it helped me to get a good understanding of BigTable Client code.

alexander-fenster:
I confirm that we have CLA on file. Will try to change labels back.

stephenplusplus:
Thanks for all of the work on the PR, it looks perfect 👍 

Per @jonparrott, we'd like to wait for this feature to be publicly available before merging and releasing. For now, testers can install this branch directly with npm:

```sh
$ npm install --save vijay-qlogic/nodejs-bigtable#app_profile_id
```

sduskis:
We'd like to launch the server-side feature soon.  We really would like this to be a feature of a publicly released client ASAP.

vijay-qlogic:
@stephenplusplus I see "cla: no" and "do not merge" status on this PR
Kindly let me know, anything pending here?

stephenplusplus:
The `cla: no` is some type of bug between the CLA Bot and Waffle.io-- nothing to worry about in this case. I put `do not merge` per https://github.com/googleapis/nodejs-bigtable/pull/74#issuecomment-379321204.

@sduskis please reach out to @jonparrott about the release timing.

PR #78, opened 2018-04-05
fix: do not require projectId in custom envs

### LINKED ISSUE #75: Default credentials error while using with a Bigtable emulator
I try to use this library with a Bigtable emulator. Everything I try to do ends with an error:
`Error: Unexpected error while acquiring application default credentials: Could not load the default credentials. Browse to https://developers.google.com/accounts/docs/application-default-credentials for more information.
`

#### Environment details

  - OS: Ubuntu 16.04
  - Node.js version: v6.14.1
  - npm version: 3.10.10
  - `@google-cloud/bigtable` version: 0.13.0

#### Steps to reproduce

  1. Download install and run Bigtable emulator.
  2. Export env variable using: ` $(gcloud beta emulators bigtable env-init)`
  3. Run an example: `node example.js`

```javascript
const Bigtable = require('@google-cloud/bigtable');

var client = new Bigtable.v2.BigtableTableAdminClient();

client.createTable("test").then().catch(function (err) {
    console.log(err)
})
```

stephenplusplus:
Fixes #75 

This will bypass the `authClient.getProjectId()` call:

- When we already have one from the user (we don't want to override theirs)
- When the user provides `apiEndpoint`
- When the user is using the emulator

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -370,8 +370,8 @@ function Bigtable(options) {
   var defaultBaseUrl = 'bigtable.googleapis.com';
   var defaultAdminBaseUrl = 'bigtableadmin.googleapis.com';
 
-  var customEndpoint =
-    options.apiEndpoint || process.env.BIGTABLE_EMULATOR_HOST;
+  var customEndpoint = (this.customEndpoint =
+    options.apiEndpoint || process.env.BIGTABLE_EMULATOR_HOST);
   var customEndpointBaseUrl;
   var customEndpointPort;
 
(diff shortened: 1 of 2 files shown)


sduskis:
In general, this looks right to me.  We could have custom endpoints for our dev/staging environments, but it looks like we can explicitly set a projectId in those cases.

stephenplusplus:
Thanks for reviewing! That's correct, if you provide a `projectId`, it will be used.

sduskis:
@kolodny, can you please review this code and provide an LGTM if things look good?  Once you do that, either feel free to approve the PR, or I will.

src/index.js
```diff
@@ -367,8 +367,8 @@ function Bigtable(options) {
   var defaultBaseUrl = 'bigtable.googleapis.com';
   var defaultAdminBaseUrl = 'bigtableadmin.googleapis.com';
 
-  var customEndpoint =
-    options.apiEndpoint || process.env.BIGTABLE_EMULATOR_HOST;
+  var customEndpoint = (this.customEndpoint =
```
kolodny: Can you assign `this.customEndpoint` on the line above?

```js
this.customEndpoint = options.apiEndpoint || process.env.BIGTABLE_EMULATOR_HOST
var customEndpoint = this.customEndpoint
```

PR #82, opened 2018-04-11
added waitForReplication method to implement consistency-token check

### LINKED ISSUE #117: Support labels on instances, and use PartialUpdateInstance
Instances support a notion of "labels" (key value pairs meaningful to users, like "project" = "myProject" or "env" = "staging").  Create and update instance should have a labels options which is basically a map of {string, string}.

Also, instance.setMetadata should use "PartialUpdateInstance" instead of "UpdateInstance"

vijay-qlogic:
Fixes #117 (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass
I'll update test-cases shortly


sduskis:
These features are not released yet. I'll update the labels once this feature is released.

sduskis:
This new method is intended to polls every 5 seconds until 10 minutes expire or checkConsistency returns `true`.  I'm not completely comfortable with the syntax, but in general, this looks right to me.  @kolodny, if the syntax is right, please LGTM this.

src/table.js
```diff
@@ -1322,6 +1322,72 @@ Table.prototype.sampleRowKeysStream = function(gaxOptions) {
...
+    reqOpts.consistencyToken = consistencyToken;
+
+    this.bigtable.request(
+      {
+        client: 'BigtableTableAdminClient',
+        method: 'checkConsistency',
+        reqOpts: reqOpts,
+      },
+      (err, resp) => {
+        return resp ? resp.consistent === 1 : false;
```
stephenplusplus: This is an asynchronous function, so returning here will not work as expected. You will have to provide a callback to the method `checkConsistency`, and then execute the callback with this value, i.e.:

```js
function checkConsistency(consistencyToken, callback) {
  reqOpts.consistencyToken = consistencyToken;

  this.bigtable.request(
    {
      client: 'BigtableTableAdminClient',
      method: 'checkConsistency',
      reqOpts: reqOpts,
    },
    (err, resp) => {
      // Added error handling:
      if (err) {
        callback(err);
        return;
      }
      callback(null, resp.consistent === 1);
    }
  );
}
```

src/table.js
```diff
@@ -1322,6 +1322,72 @@ Table.prototype.sampleRowKeysStream = function(gaxOptions) {
   ]);
 };
 
+/**
+ * Returns true in callback if consitency check is true for
+ * generated consistency token.
+ * In-case consistency check returns false, retrial is done in interval
+ * of 5 seconds till 10 minutes, after that it returns false.
+ *
+ * @param {function(?Error, ?boolean)} [callback]
```
stephenplusplus: The other methods break down the components of the response, e.g. from `sampleRowKeys`:

```
 * @param {function} [callback] The callback function.
 * @param {?error} callback.err An error returned while making this request.
 * @param {object[]} callback.keys The list of keys.
```

src/table.js
```diff
@@ -1322,6 +1322,72 @@ Table.prototype.sampleRowKeysStream = function(gaxOptions) {
...
+ * The function which will be called with the result of the API call.
+ */
+Table.prototype.waitForReplication = function(callback) {
```
stephenplusplus: We should expose the `gaxOpts` argument to this method, to be consistent with the other methods.

src/table.js
```diff
@@ -1322,6 +1322,72 @@ Table.prototype.sampleRowKeysStream = function(gaxOptions) {
   ]);
 };
 
+/**
+ * Returns true in callback if consitency check is true for
+ * generated consistency token.
+ * In-case consistency check returns false, retrial is done in interval
+ * of 5 seconds till 10 minutes, after that it returns false.
```
stephenplusplus: Should this be user-configurable?

sduskis: That's a good idea.  We did not allow it in Java since we did exponential backoff type settings, but it seems like a useful feature with this implementation.

src/table.js
vijay-qlogic: Done

src/table.js
vijay-qlogic: Could not understand, 
If we are not using gaxOptions in this method, the why we should add this as  an argument to this method?

src/table.js
```diff
@@ -1322,6 +1322,84 @@ Table.prototype.sampleRowKeysStream = function(gaxOptions) {
...
+      reqOpts: reqOpts,
+    })
+    .then(resp => {
+      const cToken = resp[0].consistencyToken;
+
+      checkConsistency(cToken, (err, res) => {
+        if (!err && res === true) {
+          return callback(null, true);
+        } else {
+          let intervalCall = setInterval(
```
stephenplusplus: I rewrote this a bit to switch from the interval to timeouts. The code below also incorporates some other suggestions, which I left comments for (please remove):

```js
Table.prototype.waitForReplication = function(callback) {
  const self = this;

  const reqOpts = {
    name: this.id,
  };

  this.bigtable
    .request({
      client: 'BigtableTableAdminClient',
      method: 'generateConsistencyToken',
      reqOpts: reqOpts,
    },
    // Switched to using a callback here, for consistency with our other methods
    (err, resp) => {
      if (err) {
        callback(err);
        return;
      }

      const timeoutAfterTenMinutes = setTimeout(() => {
        callback(null, false);
      }, 10 * 60 * 1000);

      function retryIfNecessary(err, res) {
        // Should we retry on specific errors, as opposed to all?
        // For example, if there is a network error or auth fails
        // for some reason, maybe the user needs to know?
        if (!err && res === true) {
          clearTimeout(timeoutAfterTenMinutes);
          callback(null, true);
          return;
        }

        // Using a `setTimeout` here instead of the interval. This
        // will guarantee there isn't a race condition between two
        // calls, i.e. if we send two requests, 5 seconds apart, but
        // for some reason, the first request responds after the 
        // second one.
        setTimeout(checkConsistency, 5000);
      }

      function checkConsistency() {
        const token = resp[0].consistencyToken;
        self.checkConsistency(token, retryIfNecessary);
      }
    }
  );
};

// I moved this to a new method, since it seems useful on its own.
// This will also make testing a bit easier.
Table.prototype.checkConsistency = function(token, callback) {
  const reqOpts = {
    name: this.id,
    consistencyToken: token,
  };

  this.bigtable.request(
    {
      client: 'BigtableTableAdminClient',
      method: 'checkConsistency',
      reqOpts: reqOpts,
    },
    (err, resp) => {
      if (err) {
        callback(err);
        return;
      }

      callback(null, resp.consistent === 1);
    }
  );
}

```

test/table.js
```diff
@@ -1312,6 +1312,19 @@ describe('Bigtable/Table', function() {
     });
   });
 
+  describe('waitForReplication', function() {
```
stephenplusplus: Here are some tests we should have for this PR:

- it should generate a token (check `client`, `method`, and `reqOpts` args for accuracy)
- it should return error if the token could not be generated (line 1379)

I recommend moving `checkConsistency` to its own method, i.e. `Table.prototype.checkConsistency`. After that, you can test:

- it should check consistency (check `client`, `method`, and `reqOpts` args for accuracy)
- it should return error if consistency check failed
- it should return true if consistent
- it should return false if not consistent

Back to the `waitForReplication` tests:

- it should set a timeout to return false after 10 minutes
- it should check if consistent (did it call the new method, `checkConsistency` correctly)
- it should return true if consistent
- it should clear timeout if consistent
- it should try again after 5 seconds if not consistent

I believe that covers it all!

src/table.js
vijay-qlogic: perfect, 'll update this by today

vijay-qlogic:
@stephenplusplus 'll be adding more test cases shortly

src/table.js
```diff
@@ -1322,6 +1322,87 @@ Table.prototype.sampleRowKeysStream = function(gaxOptions) {
...
+      function retryIfNecessary(err, res) {
+        if (!err && res === true) {
+          clearTimeout(timeoutAfterTenMinutes);
+          callback(null, true);
+          return;
+        }
+        setTimeout(checkConsistency, 5000);
+      }
+
+      function checkConsistency() {
```
kolodny: using checkConsistency for this inner function as well as the table method is confusing, rename this to something like recheck or poll

src/table.js
```diff
@@ -1322,6 +1322,87 @@ Table.prototype.sampleRowKeysStream = function(gaxOptions) {
...
+      if (err) {
+        callback(err);
+        return;
+      }
+
+      const timeoutAfterTenMinutes = setTimeout(() => {
+        callback(null, false);
+      }, 10 * 60 * 1000);
+
+      function retryIfNecessary(err, res) {
```
kolodny: move this function to after the poller function

src/table.js
```diff
@@ -1322,6 +1322,87 @@ Table.prototype.sampleRowKeysStream = function(gaxOptions) {
...
+        if (!err && res === true) {
```
kolodny: @sduskis if this call errors should we continue to poll

src/table.js
```diff
@@ -1322,6 +1322,87 @@ Table.prototype.sampleRowKeysStream = function(gaxOptions) {
...
+      method: 'checkConsistency',
+      reqOpts: reqOpts,
+    },
+    (err, resp) => {
+      if (err) {
+        callback(err);
+        return;
+      }
+
+      callback(null, resp.consistent === 1);
```
kolodny: Please verify that [rpc bools](https://github.com/googleapis/nodejs-bigtable/blob/29c21589efe3137813fda19969897f3cab6d7f56/protos/google/bigtable/admin/v2/bigtable_table_admin.proto#L371) come back as `1` or `true`

kolodny:
I'll take a look at the tests and see what input I can give to unblock you

vijay-qlogic:
Done rebase with master and conflicts resolved for src/table.js

src/table.js
vijay-qlogic: 'll verify this

src/table.js
vijay-qlogic: error handling updated https://github.com/googleapis/nodejs-bigtable/pull/82/commits/c65d7e05eef0b81b14cbd0e45e8cacf011bafc8d

src/table.js
vijay-qlogic: renamed function to launchCheck

kolodny:
I threw together a rough outline for async tests. Take a look at the sinon docs, since it probably makes sense to call `clock.next()` or `clock.tick(someFutureTime)` instead of `runAll`. Also what I have should be split more finely per `it()`.  You should also be testing the fake time (eg `Date.now() === 5000`) in the checkConsistancy callback


```js
describe.only('waitForReplication', function() {
    var clock;
    var setTimeoutSpy;
    var clearTimeoutSpy;

    beforeEach(() => {
      clock = sinon.useFakeTimers({
        toFake: [
          'setTimeout',
          'clearTimeout',
          'Date',
        ],
      });
      setTimeoutSpy = sinon.spy(global, 'setTimeout');
      clearTimeoutSpy = sinon.spy(global, 'clearTimeout');
    });

    afterEach(() => {
      clock.uninstall();
    });

   
    it('checks stuff', function(done) {
      var firstCall = true;
      table.bigtable.request = function(config, callback) {
        if (firstCall) { // probably better to just check config.method
          firstCall = false;
          return callback(null, [{consistencyToken: 'foobar'}]);
        } else {
          return callback(null, {consistent:1});
        }
      };
      table.waitForReplication(function(err, isConsistent) {
        // Checks that a 10 minute timer was set.
        setTimeoutSpy.calledWith(sinon.match.func, 10 * 60 * 1000);

        // Checks that clearInterval was called.
        setTimeoutSpy.called // true;
        assert.ifError(err);
        assert.strictEqual(isConsistent, true);
        done()
      });
      clock.runAll();
    });
```

sduskis:
Once this is implemented, and approved, this can be merged.  The server-side functionality is in beta.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@google-cloud/bigtable",
   "description": "Cloud Bigtable Client Library for Node.js",
-  "version": "0.13.0",
+  "version": "0.13.1",
```
stephenplusplus: We do releases in a separate PR.

sduskis:
There are too many changes in the PR.  This PR should only be for waitForReplication.

src/table.js
```diff
@@ -1341,6 +1341,7 @@ Table.prototype.sampleRowKeysStream = function(gaxOptions) {
 };
 
 /**
+<<<<<<< HEAD
```
sduskis: This looks like a bad merge.  Please delete `<<<<<<< HEAD`

src/table.js
```diff
@@ -1381,6 +1382,88 @@ Table.prototype.truncate = function(gaxOptions, callback) {
...
+      }, 10 * 60 * 1000);
+
+      // method to launch token consistency check
+      const launchCheck = () => {
+        const token = resp[0].consistencyToken;
+        this.checkConsistency(token, retryIfNecessary);
+      };
+
+      // method checks if retrial is required & init retrial with 5 sec delay
+      const retryIfNecessary = (err, res) => {
```
sduskis: The ordering is confusing.  Doesn't `retryIfNecessary` have to be in a line before `launchCheck`?

src/table.js
```diff
@@ -1381,6 +1382,88 @@ Table.prototype.truncate = function(gaxOptions, callback) {
...
+};
+
+/**
+ * Checks consistency for given ConsistencyToken
+ *
+ * @param {function(?error, ?boolean)} callback The callback function.
+ * @param {?Error} callback.err An error returned while making this request.
+ * @param {?Boolean} callback.resp Boolean value.
+ */
+Table.prototype.checkConsistency = function(token, callback) {
```
sduskis: If this is going to be a public method, then there needs to be another public method to retrieve a token via `generateConsistencyToken`

test/table.js
sduskis: @vijay-qlogic: what's the status of these tests?

src/table.js
vijay-qlogic: Yes, it returns a bool true/false as per /protos/google/bigtable/admin/v2/bigtable_table_admin.proto Ln. No. 372
Updating this

src/table.js
vijay-qlogic: Yes, methods reordered

src/table.js
vijay-qlogic: fixed

src/table.js
vijay-qlogic: Ok, 'll do this

sduskis:
The coverage needs to be at 100% before merging this.

src/table.js
vijay-qlogic: Done

test/table.js
```diff
@@ -1379,6 +1379,253 @@ describe('Bigtable/Table', function() {
...
+          assert.strictEqual(response, true);
+          assert.ifError(err);
+          done();
+        });
+
+        clock.runAll();
+      });
+
+      it('should retry checkConsistency', done => {
+        this.timeout(6000);
```
kolodny: This mocha directive shouldn't be needed since the tests isn't actually taking this long

test/table.js
```diff
@@ -1379,6 +1379,253 @@ describe('Bigtable/Table', function() {
...
+      it('should return true if consistent', function(done) {
+        table.checkConsistency({}, function(err, resp) {
+          assert.strictEqual(resp.consistent, true);
+          done();
+        });
+      });
+    });
+
+    describe('success when token is in-consistent', function() {
+      beforeEach(function() {
```
kolodny: Either reuse this beforeEach in both tests or inline them in the `it` block

test/table.js
```diff
@@ -1379,6 +1379,253 @@ describe('Bigtable/Table', function() {
...
+        assert.strictEqual(token, response.consistency_token);
+        callback(null, {consistent: true});
+      };
+
+      table.waitForReplication(() => {
+        done();
+      });
+    });
+
+    describe('retiries', () => {
```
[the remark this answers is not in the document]
vijay-qlogic: yes, that's "retries" 'll correct that shortly

src/table.js
```diff
@@ -1386,6 +1386,92 @@ Table.prototype.truncate = function(gaxOptions, callback) {
...
+      reqOpts: reqOpts,
+    },
+    callback
+  );
+};
+
+/**
+ * Checks consistency for given ConsistencyToken
+ *
+ * @param {function(?error, ?boolean)} callback The callback function.
```
stephenplusplus: We need the input parameters documented here as well.

sduskis:
@stephenplusplus and @kolodny, is this ready to be merged?

src/table.js
```diff
@@ -1386,6 +1386,94 @@ Table.prototype.truncate = function(gaxOptions, callback) {
...
+    }
+
+    // set timeout for 10 minutes
+    const timeoutAfterTenMinutes = setTimeout(() => {
+      callback(null, false);
+    }, 10 * 60 * 1000);
+
+    // method checks if retrial is required & init retrial with 5 sec delay
+    const retryIfNecessary = (err, res) => {
+      if (err || res.consistent === true) {
```
kolodny: This is wrong since if the error is set the next argument should be null or undefined. Change this block to something like this:

```js
if (err || res.consistent === true) {
  clearTimeout(timeoutAfterTenMinutes);
  if (err) {
    return callback(err);
  } else {
    return callback(null, true)
  }
}
```

Anything more is too clever IMO

test/table.js
```diff
@@ -1378,6 +1378,233 @@ describe('Bigtable/Table', function() {
...
+      });
+
+      it('should return true if token is consistent', function(done) {
+        responses = [
+          (config, callback) =>
+            callback(null, [{consistency_token: 'sample-token12345'}]),
+          (config, callback) => callback(null, {consistent: true}),
+        ];
+
+        table.bigtable.request = function(config, callback) {
```
kolodny: Set this in the beforeEach block

test/table.js
```diff
@@ -1378,6 +1378,233 @@ describe('Bigtable/Table', function() {
...
+          assert.strictEqual(response, true);
+          done();
+        });
+
+        clock.runAll();
+      });
+
+      it('should return false after 10 min if inconsistency repeats', done => {
+        this.timeout(10 * 61 * 1000); // 10 minutes
+        table.bigtable.request = function(config, callback) {
```
kolodny: Use the responses pattern the other tests are using

test/table.js
```diff
@@ -1378,6 +1378,233 @@ describe('Bigtable/Table', function() {
...
+      it('should return the error to the callback', function(done) {
+        table.checkConsistency(cToken, function(err) {
+          assert.strictEqual(err, error);
+          done();
+        });
+      });
+    });
+
+    describe('success', function() {
+      let responses = [
```
kolodny: Declare the variable here but set it's value in the beforeEach, especially since this value will get changed

sduskis:
@vijay-qlogic, there are still comments that need to be addressed.

test/table.js
vijay-qlogic: @kolodny here table.bigtable.request (with congig.method == checkConsistency) will be called unknown n-times, in-order to check, for continuous failure it must return false after 10 mins.

So can not user responses pattern.

test/table.js
vijay-qlogic: @kolodny if we set responses value in beforeEach(), .shift() on responses will be in-effective

sduskis: Does this mean that we should set `table.bigrable.request` in each test rather than doing it in the setup?

vijay-qlogic: No, that is for variable ```responses```

sduskis: Yes, `responses` is a variable used in `table.bigrable.request`.  I think that the variables and request setup should happen in the appropriate tests.

kolodny: Agreed with @sduskis, When I first reviewed there was only one test in this block, but that looks like it changed. The response value should be set at the start of each `it()` callback

vijay-qlogic:
@sduskis saw your last commit, looks like I misunderstood the changes asked by @kolodny.
I re based this branch. 
Kindly let me know if any more updates are required from my end.
We can also do a call if required.

sduskis:
@kolodny and @stephenplusplus, are all of the comments addressed?  If so can you please approve?

src/table.js
```diff
@@ -1383,6 +1383,99 @@ Table.VIEWS = {
...
+/**
+ * Generates Consistency-Token and check consistency for generated token
+ * In-case consistency check returns false, retrial is done in interval
+ * of 5 seconds till 10 minutes, after that it returns false.
+ *
+ * @param {function(?error, ?boolean)} callback The callback function.
+ * @param {?Error} callback.err An error returned while making this request.
+ * @param {?Boolean} callback.resp Boolean value.
+ */
+Table.prototype.waitForReplication = function(callback) {
```
kolodny: This will need to be changed to use the ES6 class syntax for methods

test/table.js
```diff
@@ -1378,6 +1378,248 @@ describe('Bigtable/Table', function() {
...
+          assert.ifError(err);
+          assert.strictEqual(response, true);
+          done();
+        });
+
+        clock.runAll();
+      });
+
+      it('should return false after 10 min if inconsistency repeats', done => {
+        this.timeout(10 * 61 * 1000); // 10 minutes
```
kolodny: All the `this.timeout` calls should be unnecessary and removed, they should be throwing errors since they're in fat arrow functions without a higher `this` scoped 🤷‍♂️

test/table.js
```diff
@@ -1378,6 +1378,248 @@ describe('Bigtable/Table', function() {
...
+
+        clock.runAll();
+      });
+
+      it('should return error if checkonsistency returns error', done => {
+        this.timeout(10 * 61 * 1000); // 10 minutes
+
+        let error = new Error('consistency-check error');
+
+        table.bigtable.request = function(config, callback) {
```
kolodny: Please use the responses pattern as above since it's self documenting which method should be called first. If you'd like you can add an assert for the method name in the callback

src/table.js
vijay-qlogic: Done https://github.com/googleapis/nodejs-bigtable/pull/82/commits/5ba4e36e0e7b47a136f301cf7d9a01473e68add6

src/table.js
```diff
@@ -1383,6 +1383,99 @@ Table.VIEWS = {
...
+/**
+ * Checks consistency for given ConsistencyToken
+ * @param {string} token consistency token
+ * @param {function(?error, ?boolean)} callback The callback function.
+ * @param {?Error} callback.err An error returned while making this request.
+ * @param {?Boolean} callback.resp.consistent Boolean value.
+ */
+Table.prototype.checkConsistency = function(token, callback) {
+  const reqOpts = {
+    name: this.id,
```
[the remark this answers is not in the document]
vijay-qlogic: no, not correct, added required updates https://github.com/googleapis/nodejs-bigtable/pull/82/commits/5ba4e36e0e7b47a136f301cf7d9a01473e68add6

test/table.js
vijay-qlogic: this.timeout removed https://github.com/googleapis/nodejs-bigtable/pull/82/commits/5ba4e36e0e7b47a136f301cf7d9a01473e68add6

test/table.js
vijay-qlogic: Done https://github.com/googleapis/nodejs-bigtable/pull/82/commits/5ba4e36e0e7b47a136f301cf7d9a01473e68add6

src/table.js
```diff
@@ -1368,6 +1368,99 @@ class Table {
...
+
+    // generate consistency-token
+    this.generateConsistencyToken(tokenHandler);
+  }
+
+  /**
+   * Generates Consistency-Token
+   * @param {function(?error, ?boolean)} callback The callback function.
+   * @param {?Error} callback.err An error returned while making this request.
+   * @param {?String} callback.resp.consistency_token The generated consistency token.
```
stephenplusplus: This is typically a spot where we would return just the token, and not make the user dig through the raw `resp` argument.

src/table.js
```diff
@@ -1368,6 +1368,99 @@ class Table {
...
+      callback
+    );
+  }
+
+  /**
+   * Checks consistency for given ConsistencyToken
+   * @param {string} token consistency token
+   * @param {function(?error, ?boolean)} callback The callback function.
+   * @param {?Error} callback.err An error returned while making this request.
+   * @param {?Boolean} callback.resp.consistent Boolean value.
```
stephenplusplus: This is another case where we would return just 👍 or 👎.

src/table.js
vijay-qlogic: Done https://github.com/googleapis/nodejs-bigtable/pull/82/commits/0b9d74f7edc672cbc91f912219a91c8f61bb549f

src/table.js
vijay-qlogic: Done https://github.com/googleapis/nodejs-bigtable/pull/82/commits/0b9d74f7edc672cbc91f912219a91c8f61bb549f

src/table.js
```diff
@@ -1368,6 +1368,111 @@ class Table {
...
+   * In-case consistency check returns false, retrial is done in interval
+   * of 5 seconds till 10 minutes, after that it returns false.
+   *
+   * @param {function(?error, ?boolean)} callback The callback function.
+   * @param {?Error} callback.err An error returned while making this request.
+   * @param {?Boolean} callback.resp Boolean value.
+   */
+  waitForReplication(callback) {
+    // handler for generated consistency-token
+    const tokenHandler = (err, resp) => {
```
stephenplusplus: We can just call this `token`, and save ourselves the alias line in `launchCheck`

src/table.js
```diff
@@ -1368,6 +1368,111 @@ class Table {
...
+
+    // generate consistency-token
+    this.generateConsistencyToken(tokenHandler);
+  }
+
+  /**
+   * Generates Consistency-Token
+   * @param {function(?error, ?boolean)} callback The callback function.
+   * @param {?Error} callback.err An error returned while making this request.
+   * @param {?String} callback.consistency_token The generated consistency token.
```
stephenplusplus: JS is traditionally camelCase, but still, I think we can just refer to this as `callback.token`

src/table.js
```diff
@@ -1381,6 +1381,111 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`
...
+    };
+
+    this.bigtable.request(
+      {
+        client: 'BigtableTableAdminClient',
+        method: 'generateConsistencyToken',
+        reqOpts: reqOpts,
+      },
+      (err, res) => {
+        if (err) {
```
stephenplusplus: This is just a code style comment that would match how we usually structure the code, so feel free to take it or leave it.

```js
if (err) {
  callback(err);
  return;
}

callback(null, res.consistencyToken);
```

The pattern is simply to handle errors in its own block, and let the success condition remain un-indented.

src/table.js
```diff
@@ -1381,6 +1381,111 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`
...
+    };
+
+    this.bigtable.request(
+      {
+        client: 'BigtableTableAdminClient',
+        method: 'checkConsistency',
+        reqOpts: reqOpts,
+      },
+      (err, res) => {
+        if (err) {
```
stephenplusplus: Same thought as above about error / success handling.

system-test/bigtable.js
```diff
@@ -372,6 +372,40 @@ describe('Bigtable', function() {
         done
       );
     });
+
+    it('should generate consistency token', done => {
```
stephenplusplus: Can you wrap these tests in a `describe('consistency tokens'` block?

system-test/bigtable.js
```diff
@@ -372,6 +372,40 @@ describe('Bigtable', function() {
...
+      TABLE.generateConsistencyToken(function(err, token) {
+        assert.notStrictEqual(token, undefined);
```
stephenplusplus: We don't need this line, since the next confirms it's a string.

Instead, can you add:

```js
assert.ifError(err);
```

system-test/bigtable.js
```diff
@@ -372,6 +372,40 @@ describe('Bigtable', function() {
...
+        assert.strictEqual(typeof token, 'string');
+        done();
+      });
+    });
+
+    it('should return error for checkConsistency of invalid token', done => {
+      TABLE.checkConsistency('dummy-token', function(err, res) {
+        assert.notStrictEqual(err, undefined);
```
stephenplusplus: We don't have to worry about testing the `res` argument, but we should do some type of checking that `err.message` is the message we expect.

system-test/bigtable.js
```diff
@@ -372,6 +372,40 @@ describe('Bigtable', function() {
...
+        assert.strictEqual(res, undefined);
+        done();
+      });
+    });
+
+    it('should return boolean for checkConsistency of token', done => {
+      TABLE.generateConsistencyToken(function(err, token) {
+        TABLE.checkConsistency(token, function(err, res) {
+          assert.strictEqual(err, null);
```
stephenplusplus:
```js
assert.ifError(err);
```

system-test/bigtable.js
```diff
@@ -372,6 +372,40 @@ describe('Bigtable', function() {
...
+          assert.strictEqual(typeof res, 'boolean');
+          done();
+        });
+      });
+    });
+
+    it('should return boolean for waitForReplication', done => {
+      TABLE.waitForReplication(function(err, res) {
+        assert.strictEqual(err, null);
```
stephenplusplus:
```js
assert.ifError(err);
```

src/table.js
vijay-qlogic: Done https://github.com/googleapis/nodejs-bigtable/pull/82/commits/02debc6039125a03b782c1a1a5216abb35e2ce84

src/table.js
vijay-qlogic: Done https://github.com/googleapis/nodejs-bigtable/pull/82/commits/02debc6039125a03b782c1a1a5216abb35e2ce84

src/table.js
vijay-qlogic: Done https://github.com/googleapis/nodejs-bigtable/pull/82/commits/02debc6039125a03b782c1a1a5216abb35e2ce84

src/table.js
vijay-qlogic: Done https://github.com/googleapis/nodejs-bigtable/pull/82/commits/02debc6039125a03b782c1a1a5216abb35e2ce84

sduskis:
@stephenplusplus, can you please do a final review?  It looks like all of the comments were addressed.

system-test/bigtable.js
```diff
@@ -374,6 +374,42 @@ describe('Bigtable', function() {
...
+      TABLE.generateConsistencyToken(function(err, token) {
+        assert.ifError(err);
+        assert.strictEqual(typeof token, 'string');
+        done();
+      });
+    });
+
+    it('should return error for checkConsistency of invalid token', done => {
+      TABLE.checkConsistency('dummy-token', function(err, res) {
+        assert.strictEqual(err.message, '3 INVALID_ARGUMENT: Invalid token');
```
stephenplusplus: Knowing our system test flakiness history too well, this is bound to be broken someday by an upstream change somewhere. To be a bit more future-proof, we could either check just the error code, or that the message includes simply "invalid token".

system-test/bigtable.js
```diff
@@ -374,6 +374,42 @@ describe('Bigtable', function() {
...
+        assert.strictEqual(res, undefined);
```
stephenplusplus: We can skip validating `res`. Just confirming the error is good enough.

test/table.js
```diff
@@ -1390,6 +1390,238 @@ describe('Bigtable/Table', function() {
...
+      table.generateConsistencyToken = function(callback) {
+        callback(null, consistencyToken);
+      };
+
+      table.checkConsistency = function(token, callback) {
+        assert.strictEqual(token, consistencyToken);
+        callback(null, true);
+      };
+
+      table.waitForReplication(() => {
```
stephenplusplus: `table.waitForReplication(done);` is good. It would catch an error should one appear.

test/table.js
```diff
@@ -1390,6 +1390,238 @@ describe('Bigtable/Table', function() {
...
+      table.bigtable.request = function(config) {
+        assert.strictEqual(config.client, 'BigtableTableAdminClient');
+        assert.strictEqual(config.method, 'generateConsistencyToken');
+        assert.strictEqual(config.reqOpts.name, table.name);
+        done();
+      };
+      table.generateConsistencyToken(assert.ifError);
+    });
+
+    it('should return a consistencyToken', function(done) {
```
stephenplusplus: One more test to check that it returns an error as well, please.

test/table.js
```diff
@@ -1390,6 +1390,238 @@ describe('Bigtable/Table', function() {
...
+
+      table.bigtable.request = function(config) {
+        assert.strictEqual(config.client, 'BigtableTableAdminClient');
+        assert.strictEqual(config.method, 'checkConsistency');
+        assert.strictEqual(config.reqOpts.name, table.name);
+        assert.strictEqual(config.reqOpts.consistencyToken, cToken);
+        done();
+      };
+
+      table.checkConsistency(cToken);
```
stephenplusplus: You can do `table.checkConsistency(cToken, assert.ifError);`, so that it catches an error.

test/table.js
```diff
@@ -1390,6 +1390,238 @@ describe('Bigtable/Table', function() {
...
+      const response = {
+        consistencyToken: cToken,
+      };
+
+      table.bigtable.request = function(config, callback) {
+        callback(null, response);
+      };
+
+      table.generateConsistencyToken(function(err, token) {
+        assert.strictEqual(token, cToken);
```
stephenplusplus: Above this check, do an `assert.ifError(err);`.

test/table.js
```diff
@@ -1390,6 +1390,238 @@ describe('Bigtable/Table', function() {
...
+          if (config.method === 'generateConsistencyToken') {
+            return callback(null, {consistencyToken: 'sample-token12345'});
+          }
+          if (config.method === 'checkConsistency') {
+            return callback(null, {consistent: false});
+          }
+        };
+
+        table.waitForReplication(function(err, response) {
+          setTimeoutSpy.called;
```
stephenplusplus: `assert.ifError(err);`

test/table.js
```diff
@@ -1391,6 +1391,250 @@ describe('Bigtable/Table', function() {
...
+          done();
+        });
+      });
+
+      it('should return false if not consistent', function(done) {
+        table.bigtable.request = function(config, callback) {
+          callback(null, {consistent: false});
+        };
+        table.checkConsistency('', function(err, resp) {
+          assert.strictEqual(resp, false);
```
stephenplusplus: `assert.ifError(err);`

test/table.js
```diff
@@ -1391,6 +1391,250 @@ describe('Bigtable/Table', function() {
...
+    });
+
+    describe('success', function() {
+      it('should return true if consistent', function(done) {
+        table.bigtable.request = function(config, callback) {
+          callback(null, {consistent: true});
+        };
+
+        table.checkConsistency('', function(err, resp) {
+          assert.strictEqual(resp, true);
```
stephenplusplus: `assert.ifError(err);`

sduskis:
@stephenplusplus, I added the `ifError`s

sduskis:
@stephenplusplus, can you please Approve?  I made the last two changes you requested.

stephenplusplus:
Thank you!

Issue #86, opened 2018-04-12
Drop support for node.js 4


JustinBeckwith:

sduskis:
This was fixed by #159 and #160

PR #87, opened 2018-04-13
Switch src/ to use ES classes

### LINKED ISSUE #85: Convert to ES classes
We have code like this:

```js
if (!(this instanceof ChunkTransformer)) {
    return new ChunkTransformer(options);
  }
```

That won't work with TypeScript :)

kolodny:
## Node v4 goes EOL 4/30/18

Fixes #85 

- [ ] Tests and linter pass

Note that instantiating a `Bigtable` still works via Proxies (See https://stackoverflow.com/a/35339748 ), yet all the other classes require the `new` keyword

This was generated by installing [lebab](https://github.com/lebab/lebab) and running:

```
./node_modules/.bin/lebab --replace test/ --transform class
```

and then fixing the newly broken tests.

I imagine we'll want to followup on the rest of the other transforms lebab offers.

### DIFF
--- a/src/app-profile.js
+++ b/src/app-profile.js
@@ -14,13 +14,11 @@
  * limitations under the License.
  */
 
-'use strict';
+const common = require('@google-cloud/common');
+const is = require('is');
+const snakeCase = require('lodash.snakecase');
 
-var common = require('@google-cloud/common');
-var is = require('is');
-var snakeCase = require('lodash.snakecase');
-
-var Cluster = require('./cluster.js');
+const Cluster = require('./cluster.js');
 
 /**
  * Create an app profile object to interact with your app profile.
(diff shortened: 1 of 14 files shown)


kolodny:
This PR is broken since it's still trying to build on node v4, once that's removed (#88 ) this should pass all tests

stephenplusplus:
This is pretty cool!

sduskis:
Is this ready to go now that node.js 4 was EOLed?

kolodny:
Should be. First thing would be to resolve #86. I can get working on that and then finish this one.

kolodny:
@alexander-fenster Is #86 something that needs to be done on your end or can this be done with some config file changes here?

sduskis:
@JustinBeckwith: What's the timeline for removing Node.js v4 support?

JustinBeckwith:
We are trying to keep node.js 4 support in all of our base npm modules (common, auth, etc).  For all others (like this library), you are free to drop Node.js 4, as long as you do a semver major bump.  

I would very much like to get the constructor pattern change in as well with that change.

kolodny:
@JustinBeckwith I have an almost ready to go version of this PR. The only blocker now is `util.promisifyAll` not working on ES classes. I have a fix out at  https://github.com/googleapis/nodejs-common/pull/150

That just needs to get merged and released then I can proceed with this PR

kolodny:
So this PR should be ready to go. I believe this requires a major semver bump. I figured the test files shouldn't be updated in the same PR. The only thing that really needs to be pointed out is that classes shouldn't be instaniatable without the new keyword:

```js
const client1 = new Bigtable()
const client1 = Bigtable() // Error: Class constructor Foo cannot be invoked without 'new'
```

I was on the fence about requiring people to use the `new` keyword or not, and chose to make it optional. I did this by using a [proxy](https://github.com/googleapis/nodejs-bigtable/pull/87/files#diff-1fdf421c05c1140f6d71444ea2b27638R802):

```js
Bigtable = new Proxy(Bigtable, {
  apply(target, thisArg, argumentsList) {
    return new target(...argumentsList);
  },
});
```

Let me know if this should be removed. I didn't add that to the semi-private classes since the common case of creating one is something like `instance.table(...)` and not `new Table(...)`

As before, most of this was done via lebab so there may be some issues so please do go through it again to make sure nothing terrible is being done.

Linting seems be having some issues so I'll be pushing more commits to fix that

kolodny:
This PR is acting weird. I'm closing this for now and opening a new PR with the same changes.

PR #88, opened 2018-04-13
Drop support for nodev4.

### LINKED ISSUE #86: Drop support for node.js 4

kolodny:
## Node 4 goes EOL 4/30/18

Fixes #86

- [x] Tests and linter pass

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,10 +3,6 @@ workflows:
   version: 2
   tests:
     jobs: &workflow_jobs
-      - node4:
-          filters:
-            tags:
-              only: /.*/
       - node6:
           filters:
             tags:
@@ -21,7 +17,6 @@ workflows:
               only: /.*/
       - lint:
           requires:
-            - node4
             - node6
             - node8
             - node9
@@ -30,7 +25,6 @@ workflows:
               only: /.*/
       - docs:
           requires:
-            - node4
             - node6
             - node8
             - node9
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -73,38 +67,6 @@ workflows:
               only: master
     jobs: *workflow_jobs
 jobs:
-  node4:
-    docker:
-      - image: 'node:4'
-        user: node
-    steps: &unit_tests_steps
```
stephenplusplus: We have to move this section to the next-lowest version supported. This is how the nightly tests work, I believe.

@alexander-fenster do we want to make this change one-off here (in other words, do we want this PR), or can the generator do it across all of our projects at once?

alexander-fenster: Both are fine (the generator generally refuses to make a change if it's already made, so nothing bad will happen if the node4 task is already deleted). But do we stop testing on node4 right now already? 

My understanding is that we want to wait at least for 4/30 when Node 4 is officially EOL. I'd like to add Node 10 into the same change so that we'd have one change across all repos, and not two.

In other words, I'd probably like to wait for 4/30 and then I'll make the change across all repos to disable node4 and add node10.

alexander-fenster: And, for sure, removal of this section is wrong as it's referenced from other jobs.

alexander-fenster:
@kolodny why do you want to remove node4 now? It's technically alive now (till 4/30).

kolodny:
@alexander-fenster My mistake, I thought EOL was 4/1 I'll close this PR for now and reopen after that time

PR #95, opened 2018-04-20
Check if Error is FamilyError for autoCreate.

### LINKED ISSUE #90: Auto create family does not work
The autoCreate option does not work for family

#### Environment details

OS: macos 10.13.3
Node.js version: 8.10.0
npm version: 5.6.0
yarn version: 1.3.2
google-cloud-node version: master branch (0.13)

#### Steps to reproduce

```js
        const table = bt.table('TestAutoCreate');
        await table.get({ autoCreate: true });

        const fam = table.family('AutoCreateFamily');
        await fam.get({ autoCreate: true });
```

```
 FamilyError: Column family not found: projects/teo-dev-de/instances/teo-dev-de/tables/TestAutoCreate/columnFamilies/AutoCreateFamily.
    at /Users/moander/dev/uzi/api/node_modules/@google-cloud/bigtable/src/family.js:355:17
    at /Users/moander/dev/uzi/api/node_modules/@google-cloud/bigtable/src/table.js:814:5
    at Immediate._onImmediate (/Users/moander/dev/uzi/api/node_modules/@google-cloud/bigtable/src/table.js:871:16)
```

kolodny:
Relying on the error code is fragile (and broken). Since the client library
manually controls when this error is thrown, it's the most reliable test
to use for autoCreate.

Fixes #90

- [x] Tests and linter pass
~- Appropriate docs were updated (if necessary)~

### DIFF
--- a/src/family.js
+++ b/src/family.js
@@ -296,7 +296,7 @@ Family.prototype.get = function(options, callback) {
 
   this.getMetadata(gaxOptions, function(err, metadata) {
     if (err) {
-      if (err.code === 5 && autoCreate) {
+      if (err instanceof FamilyError && autoCreate) {
         self.create({gaxOptions}, callback);
         return;
       }
--- a/system-test/bigtable.js
+++ b/system-test/bigtable.js
@@ -264,6 +264,14 @@ describe('Bigtable', function() {
         done();
       });
     });
+
+    it('should create a table if autoCreate is true', function(done) {
+      var table = INSTANCE.table(generateName('table'));
+      async.series(
+        [table.get.bind(table, {autoCreate: true}), table.delete.bind(table)],
+        done
+      );
+    });
   });
 
   describe('column families', function() {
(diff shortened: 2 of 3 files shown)


stephenplusplus:
Can you add a system test to verify this fixes #90?

kolodny:
@stephenplusplus Done

src/family.js
```diff
@@ -296,7 +296,7 @@ Family.prototype.get = function(options, callback) {
 
   this.getMetadata(gaxOptions, function(err, metadata) {
     if (err) {
-      if (err.code === 5 && autoCreate) {
```
moander: I'm seeing `err.code=404` using autoCreate. Will this sometimes be code 5 instead?

kolodny: `5` is the grpc code for 404. I'm not sure why we chose to use http sometimes, After this release there should be no more errors at all when using autoCreate (unless creation fails for some reason)

PR #100, opened 2018-04-23
Add dedicated method to truncate a table.

### LINKED ISSUE #84: Consider not deleting all rows when no prefix is defined
I suggest changing the behaviour of deleteRows to make it less easy to delete all rows in a table by mistake.

kolodny:
Fixes #84

- [x] Tests and linter pass

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -623,7 +623,7 @@ Table.prototype.truncate = function(gaxOptions, callback) {
  * Delete all rows in the table, optionally corresponding to a particular
  * prefix.
  *
- * @param {string} [prefix] Row key prefix. This field is required.
+ * @param {string} prefix Row key prefix.
  * @param {object} [gaxOptions] Request configuration options, outlined
  *     here: https://googleapis.github.io/gax-nodejs/CallSettings.html.
  * @param {function} callback The callback function.
--- a/src/table.js
+++ b/src/table.js
@@ -623,8 +623,9 @@ Table.prototype.truncate = function(gaxOptions, callback) {
  * Delete all rows in the table, optionally corresponding to a particular
  * prefix.
  *
- * @param {object} [options] Configuration object.
- * @param {string} [options.prefix] Row key prefix.
+ * @param {string} [prefix] Row key prefix. This field is required.
+ * @param {object} [gaxOptions] Request configuration options, outlined
+ *     here: https://googleapis.github.io/gax-nodejs/CallSettings.html.
  * @param {function} callback The callback function.
  * @param {?error} callback.err An error returned while making this request.
  * @param {object} callback.apiResponse The full API response.
@@ -644,30 +645,28 @@ Table.prototype.truncate = function(gaxOptions, callback) {
  *   }
  * };
  *
- * table.deleteRows({
- *   prefix: 'alincoln'
- * }, callback);
+ * table.deleteRows('alincoln', callback);
  *
  * //-
  * // If the callback is omitted, we'll return a Promise.
  * //-
- * table.deleteRows({ prefix: 'alincoln' }).then(function(data) {
+ * table.deleteRows('alincoln').then(function(data) {
  *   const apiResponse = data[0];
  * });
  */
-Table.prototype.deleteRows = function(options, callback) {
-  if (is.function(options)) {
-    callback = options;
-    options = {};
+Table.prototype.deleteRows = function(prefix, gaxOptions, callback) {
+  if (is.function(gaxOptions)) {
+    callback = gaxOptions;
+    gaxOptions = {};
   }
 
-  if (!options || !options.prefix) {
-    throw new Error('Use the truncate method to truncate a table.');
+  if (!prefix || typeof prefix === 'function') {
+    throw new Error('A prefix is required for deleteRows.');
   }
 
   var reqOpts = {
     name: this.id,
-    rowKeyPrefix: Mutation.convertToBytes(options.prefix),
+    rowKeyPrefix: Mutation.convertToBytes(prefix),
   };
 
   this.bigtable.request(
@@ -675,7 +674,7 @@ Table.prototype.deleteRows = function(options, callback) {
       client: 'BigtableTableAdminClient',
       method: 'dropRowRange',
       reqOpts: reqOpts,
-      gaxOpts: options.gaxOptions,
+      gaxOpts: gaxOptions,
     },
     callback
   );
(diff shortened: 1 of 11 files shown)


src/table.js
```diff
@@ -578,13 +578,53 @@ Table.prototype.delete = function(gaxOptions, callback) {
...
+};
+
 /**
  * Delete all rows in the table, optionally corresponding to a particular
  * prefix.
  *
  * @param {object} [options] Configuration object.
- * @param {string} [options.prefix] Row key prefix, when omitted all rows
- *     will be deleted.
+ * @param {string} [options.prefix] Row key prefix.
```
sduskis: Can you add a comment that this is required?

system-test/bigtable.js
```diff
@@ -1035,17 +1035,48 @@ describe('Bigtable', function() {
 
         row.delete(done);
       });
+    });
 
-      it('should delete all the rows', function(done) {
-        TABLE.deleteRows(function(err) {
-          assert.ifError(err);
+    describe('.truncate()', function() {
```
sduskis: Did you see the change in #101?

src/table.js
stephenplusplus: That's a good point-- Can the signature of the method just be a prefix and callback?

```js
table.deleteRows('prefix', callback)
```

sduskis: Don't we need to be able to pass in gaxOptions?

stephenplusplus: Another good point! Looks like I forgot to document that, so it slipped my mind.

However, in cases like this, we would have `gaxOptions` be a standalone argument, so that we use the most predictable path possible:

```js
// 99 percent of the time
table.deleteRows({ prefix: 'prefix' }, callback)

// 1% of the time
table.deleteRows({ prefix: 'prefix', gaxOptions: {...} }, callback)

// So why not
table.deleteRows('prefix', callback)
```

kolodny: Done

system-test/bigtable.js
kolodny: Yup, decorating the updated test with `skip`

src/table.js
```diff
@@ -578,13 +578,54 @@ Table.prototype.delete = function(gaxOptions, callback) {
...
+};
+
 /**
  * Delete all rows in the table, optionally corresponding to a particular
  * prefix.
  *
- * @param {object} [options] Configuration object.
- * @param {string} [options.prefix] Row key prefix, when omitted all rows
- *     will be deleted.
+ * @param {string} [prefix] Row key prefix. This field is required.
```
stephenplusplus: This can just be `prefix` now. The surrounding brackets indicate that it's optional.

stephenplusplus: (you can also remove `This field is required.` as that will be implied.

src/table.js
```diff
@@ -604,44 +645,36 @@ Table.prototype.delete = function(gaxOptions, callback) {
...
-  if (is.function(options)) {
-    callback = options;
-    options = {};
+Table.prototype.deleteRows = function(prefix, gaxOptions, callback) {
+  if (is.function(gaxOptions)) {
+    callback = gaxOptions;
+    gaxOptions = {};
+  }
+
+  if (!prefix || typeof prefix === 'function') {
```
stephenplusplus: To be consistent, we use the `is` module:

```js
is.fn(prefix)
```

src/table.js
kolodny: done

src/table.js
```diff
@@ -604,44 +645,36 @@ Table.prototype.delete = function(gaxOptions, callback) {
...
+    throw new Error('A prefix is required for deleteRows.');
```
stephenplusplus: We can note in the JSDocs block:

```js
@throws {error} If a prefix is not provided.
```

(reference: https://github.com/kolodny/nodejs-bigtable/blob/95e626dd9b7c91cca4f9f11ac3f8c0953c64e22b/src/table.js#L178)

src/table.js
kolodny: good call. Switched

src/table.js
kolodny: Done

test/table.js
```diff
@@ -1056,11 +1052,41 @@ describe('Bigtable/Table', function() {
       table.bigtable.request = function(config) {
         assert.strictEqual(config.reqOpts.rowKeyPrefix, fakePrefix);
         assert.strictEqual(spy.callCount, 1);
-        assert.strictEqual(spy.getCall(0).args[0], options.prefix);
+        assert.strictEqual(spy.getCall(0).args[0], prefix);
+        done();
+      };
+
+      table.deleteRows('a', assert.ifError);
```
stephenplusplus: I think you meant to pass in `prefix` here instead of `'a'`.

src/table.js
```diff
@@ -578,13 +578,56 @@ Table.prototype.delete = function(gaxOptions, callback) {
...
+ * table.truncate(function(err, apiResponse) {});
+ *
+ * //-
+ * // If the callback is omitted, we'll return a Promise.
+ * //-
+ * table.truncate().then(function(data) {
+ *   var apiResponse = data[0];
+ * });
+ */
+Table.prototype.truncate = function(gaxOptions, callback) {
```
stephenplusplus: This isn't my favorite request to make, but we try to keep our methods defined alphabetically so that it's easier to dig into the src/tests. Would you mind relocating it and the describe block in the tests?

test/table.js
```diff
@@ -1056,11 +1052,41 @@ describe('Bigtable/Table', function() {
...
+    });
+  });
+
+  describe('truncate', function() {
+    it('should provide the proper request options', function(done) {
+      table.bigtable.request = function(config, callback) {
+        assert.strictEqual(config.client, 'BigtableTableAdminClient');
+        assert.strictEqual(config.method, 'dropRowRange');
+        assert.strictEqual(config.reqOpts.name, TABLE_NAME);
+        callback();
```
stephenplusplus: We should have a check to confirm `config.gaxOpts` is `{}`.

test/table.js
```diff
@@ -1056,11 +1052,41 @@ describe('Bigtable/Table', function() {
...
+      }, /A prefix is required for deleteRows\./);
+    });
+  });
+
+  describe('truncate', function() {
+    it('should provide the proper request options', function(done) {
+      table.bigtable.request = function(config, callback) {
+        assert.strictEqual(config.client, 'BigtableTableAdminClient');
+        assert.strictEqual(config.method, 'dropRowRange');
+        assert.strictEqual(config.reqOpts.name, TABLE_NAME);
```
stephenplusplus: We also need to confirm `deleteAllDataFromTable` is set to `true`.

PR #110, opened 2018-04-29
Add type option to instance creation.

kolodny:
- [x] Tests and linter pass


test/index.js
```diff
@@ -388,6 +388,21 @@ describe('Bigtable', function() {
       bigtable.createInstance(INSTANCE_NAME, options, assert.ifError);
     });
 
+    it('should respect the type option', function(done) {
+      var options = {type: 'development'};
+      sinon.spy(FakeInstance, 'getTypeType_');
+
+      bigtable.request = function(config) {
+        assert.deepEqual(config.reqOpts.instance.type, 2);
```
stephenplusplus: This line tests functionality of `Instance.getTypeType_`. We need another test block for the new `Instance.getTypeType_` function to make sure that works as intended.

test/index.js
```diff
@@ -388,6 +388,21 @@ describe('Bigtable', function() {
       bigtable.createInstance(INSTANCE_NAME, options, assert.ifError);
     });
 
+    it('should respect the type option', function(done) {
+      var options = {type: 'development'};
+      sinon.spy(FakeInstance, 'getTypeType_');
```
stephenplusplus: We need to simulate what this method returns, so we can verify that's what ends up being sent to `request`.

For example, this pattern:

```js
var options = {type: 'development'}
var fakeTypeType = 99

FakeInstance.getTypeType_ = function(options) {
  assert.strictEqual(type, options.type)
  return fakeTypeType
}

bigtable.request = function(config) {
  assert.strictEqual(config.reqOpts.instance.type, fakeTypeType)
}

bigtable.createInstance(INSTANCE_NAME, options, assert.ifError)
```

test/index.js
kolodny: Added a test in https://github.com/googleapis/nodejs-bigtable/pull/110/files#diff-c5fd52f715b6d5d9a356e4b76caf7511

test/index.js
kolodny: Done

test/instance.js
```diff
@@ -127,6 +127,28 @@ describe('Bigtable/Instance', function() {
...
+      production: 1,
+      development: 2,
+    };
+
+    it('should default to unspecified', function() {
+      assert.strictEqual(Instance.getTypeType_(), types.unspecified);
+    });
+
+    it('should lowercase a type', function() {
+      assert.strictEqual(Instance.getTypeType_('production'), types.production);
```
stephenplusplus: Did you mean `'PRODUCTION'`?

test/instance.js
```diff
@@ -127,6 +127,28 @@ describe('Bigtable/Instance', function() {
...
 
+  describe('getTypeType_', function() {
+    var types = {
+      unspecified: 0,
+      production: 1,
+      development: 2,
+    };
+
+    it('should default to unspecified', function() {
+      assert.strictEqual(Instance.getTypeType_(), types.unspecified);
```
stephenplusplus: You could add an extra test here like:

```js
assert.strictEqual(Instance.getTypeType_('not-real-type'), types.unspecified)
```

test/instance.js
kolodny: Yes, good catch

test/instance.js
kolodny: Sounds good!

PR #114, opened 2018-05-01
chore: lock files maintenance

alexander-fenster:
🔨 update lock files to use the latest gRPC


stephenplusplus:
I think this is covered in #121

Issue #108, opened 2018-04-27
AppProfileID CRUD


sduskis:
The instance admin API allows the creation of AppProfileIds.  We need to be able to access that functionality from instance.js.

PR #121, opened 2018-05-03
bump version to 0.14.0

crwilcox:
# Changes
- MutateRow and MutateRows now set to use idempotent retry setting.
- x-goog-request-params header added to options for default requests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/bigtable",
-  "version": "0.13.1",
+  "version": "0.14.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


alexander-fenster:
By looking at the commit history, I see several more changes. @kolodny do any of them worth mentioning in the release notes? (the first message of this PR) I'm not sure about the current state of the code so would ask @kolodny / @stephenplusplus / @callmehiphop to sign off the release. Thanks!

kolodny:
For reference I believe this is the bulk of the commits
image

I would call out the truncate method which is a breaking change since deleteRows won't do that anymore. Defaulting to use client side timestamps as the default also seems to be noteworthy. If you also put in a bugfix section, there was the fixes with autocreation of families. That's all from my end @stephenplusplus or @sduskis may also want to chime in.

I'm going to create a PR for #108 in the next day or so in case that's something you want to hold off for

crwilcox:
Waiting on https://github.com/googleapis/nodejs-bigtable/pull/124

kolodny:
#124 is now done. We also should add #140 to the release notes

sduskis:
It looks like there are some conflicts to resolve.

PR #124, opened 2018-05-04
Add basic CRUDL operations for App Profiles.

### LINKED ISSUE #108: AppProfileID CRUD

kolodny:
Fixes #108

- [x] Tests and linter pass

### DIFF
--- a/src/app-profile.js
+++ b/src/app-profile.js
@@ -1,0 +1,391 @@
+/*!
+ * Copyright 2018 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+var common = require('@google-cloud/common');
+var is = require('is');
+var snakeCase = require('lodash.snakecase');
+
+var Cluster = require('./cluster.js');
+
+/**
+ * Create an app profile object to interact with your app profile.
+ *
+ * @class
+ * @param {Instance} instance The parent instance of this app profile.
+ * @param {string} name Name of the app profile.
+ *
+ * @example
+ * const Bigtable = require('@google-cloud/bigtable');
+ * const bigtable = new Bigtable();
+ * const instance = bigtable.instance('my-instance');
+ * const appProfile = instance.appProfile('my-app-profile');
+ */
+function AppProfile(instance, name) {
+  this.bigtable = instance.bigtable;
+  this.instance = instance;
+
+  var id = name;
+
+  if (id.indexOf('/') === -1) {
+    id = `${instance.id}/appProfiles/${name}`;
+  }
+
+  this.id = id;
+  this.name = id.split('/').pop();
+}
+
+/**
+ * Formats a app profile options object into proto format.
+ *
+ * @private
+ *
+ * @param {object} options The options object.
+ * @returns {object}
+ *
+ * @example
+ * Family.formatAppProfile_({
+ *   routing: 'any',
+ *   description: 'My App Profile',
+ * });
+ * // {
+ * //   multiClusterRoutingUseAny: {},
+ * //   description: 'My App Profile',
+ * // }
+ */
+AppProfile.formatAppProfile_ = function(options) {
+  const appProfile = {};
+
+  if (options.routing) {
+    if (options.routing === 'any') {
+      appProfile.multiClusterRoutingUseAny = {};
+    } else if (options.routing instanceof Cluster) {
+      appProfile.singleClusterRouting = {
+        clusterId: options.routing.name,
+      };
+      if (is.boolean(options.allowTransactionalWrites)) {
+        appProfile.singleClusterRouting.allowTransactionalWrites =
+          options.allowTransactionalWrites;
+      }
+    } else {
+      throw new Error(
+        'An app profile routing policy can only contain "any" or a `Cluster`.'
+      );
+    }
+  }
+
+  if (is.string(options.etag)) {
+    appProfile.etag = options.etag;
+  }
+
+  if (is.string(options.description)) {
+    appProfile.description = options.description;
+  }
+
+  return appProfile;
+};
+
+/**
+ * Create an app profile.
+ *
+ * @param {object} [options] See {@link Instance#createAppProfile}.
+ *
+ * @example
+ * const Bigtable = require('@google-cloud/bigtable');
+ * const bigtable = new Bigtable();
+ * const instance = bigtable.instance('my-instance');
+ * const appProfile = instance.appProfile('my-appProfile');
+ *
+ * appProfile.create(function(err, appProfile, apiResponse) {
+ *   if (!err) {
+ *     // The app profile was created successfully.
+ *   }
+ * });
+ *
+ * //-
+ * // If the callback is omitted, we'll return a Promise.
+ * //-
+ * appProfile.create().then(function(data) {
+ *   const appProfile = data[0];
+ *   const apiResponse = data[1];
+ * });
+ */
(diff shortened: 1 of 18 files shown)


src/app-profile.js
```diff
@@ -0,0 +1,391 @@
...
+    gaxOptions = {};
+  }
+
+  var reqOpts = {
+    appProfile: AppProfile.formatAppProfile_(metadata),
+    updateMask: {
+      paths: [],
+    },
+  };
+  reqOpts.appProfile.name = this.id;
```
kolodny: This is what is required when I was trying to get updates working. Putting the name as a property of `reqOpts` didn't do anything

src/instance.js
```diff
@@ -118,6 +129,101 @@ Instance.prototype.create = function(options, callback) {
   this.bigtable.createInstance(this.name, options, callback);
 };
 
+/**
+ * Create an app profile.
+ *
+ * @param {string} name The name to be used when referring to the new
+ *     app profile within its instance.
+ * @param {object} options AppProfile creation options.
+ * @param {'any'|Cluster} options.routing  The routing policy for all
```
sduskis: If this type is `Cluster` then `options.routing,name` is required.  I think that's important for the docs.  It might also be worth while using `options.routing.clusterId` instead of `name` so that the usage is clearer.

src/instance.js
```diff
@@ -118,6 +129,101 @@ Instance.prototype.create = function(options, callback) {
...
+ * @param {object} [options.gaxOptions]  Request configuration options, outlined
+ *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
+ * @param {boolean} [options.allowTransactionalWrites] Whether or not
+ *     CheckAndMutateRow and ReadModifyWriteRow requests are allowed by this
+ *     app profile. It is unsafe to send these requests to the same
+ *     table/row/column in multiple clusters. This is only used when the
+ *     routing value is a cluster.
+ * @param {string} [options.description] The long form description of the use
+ *     case for this AppProfile.
+ * @param {string} [options.etag] Strongly validated etag for optimistic
```
sduskis: Can we remove options.etag for now, since it's not implemented on the server-side?

src/instance.js
sduskis: oh.  I see.  This needs to be an instance of a `Cluster` object.  That's confusing.  Would `any|[a clusterId in the instance]` work  instead of a `Cluster` object?

src/instance.js
kolodny: Done

src/instance.js
kolodny: I believe the jsdoc system expects a format the way it is. If the user provides a cluster not in the instance, the API will reject the call

src/app-profile.js
```diff
@@ -0,0 +1,387 @@
...
+ * const appProfile = instance.appProfile('my-app-profile');
+ */
+function AppProfile(instance, name) {
+  this.bigtable = instance.bigtable;
+  this.instance = instance;
+
+  var id = name;
+
+  if (id.indexOf('/') === -1) {
+    id = `${instance.id}/appProfiles/${name}`;
```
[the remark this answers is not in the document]
sduskis: see #137 for more discussion.  I think that this topic shouldn't block AppProfiles, but we'll have to revisit this topic before beta.

src/app-profile.js
```diff
@@ -0,0 +1,391 @@
...
+
+  if (options.routing) {
+    if (options.routing === 'any') {
+      appProfile.multiClusterRoutingUseAny = {};
+    } else if (options.routing instanceof Cluster) {
+      appProfile.singleClusterRouting = {
+        clusterId: options.routing.name,
+      };
+      if (is.boolean(options.allowTransactionalWrites)) {
+        appProfile.singleClusterRouting.allowTransactionalWrites =
```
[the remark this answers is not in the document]
sduskis: looks good

src/app-profile.js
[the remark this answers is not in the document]
kolodny: Correct, we allow passing in the id or the name here, we call it `name` for the constructor. I'm following the same pattern we use for [cluster](https://github.com/googleapis/nodejs-bigtable/blob/master/src/cluster.js#L36-L47) and [instance](https://github.com/googleapis/nodejs-bigtable/blob/master/src/instance.js#L40-L51)

src/instance.js
```diff
@@ -76,6 +77,16 @@ Instance.getTypeType_ = function(type) {
   return types[type] || types.unspecified;
 };
 
+/**
+ * Get a reference to a Bigtable App Profile.
+ *
+ * @param {string} name The name of the cluster.
```
sduskis: Should this be the name of the appProfile?

src/instance.js
```diff
@@ -76,6 +77,16 @@ Instance.getTypeType_ = function(type) {
...
+ * @returns {Cluster}
```
sduskis: `@returns {AppProfile}`

src/instance.js
```diff
@@ -118,6 +129,99 @@ Instance.prototype.create = function(options, callback) {
...
+ * const Bigtable = require('@google-cloud/bigtable');
+ * const bigtable = new Bigtable();
+ * const instance = bigtable.instance('my-instance');
+ *
+ * const callback = function(err, appProfile, apiResponse) {
+ *   // `appProfile` is an AppProfile object.
+ * };
+ *
+ * const options = {
+ *   routing: instance,
```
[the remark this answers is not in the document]
sduskis: If this is for a Cluster based routing, can you please update the sample code appropriately?

Should this be: `routing: instance.cluster("my-cluster")`?

system-test/bigtable.js
```diff
@@ -57,7 +60,20 @@ describe('Bigtable', function() {
             {
               families: ['follows', 'traits'],
             },
-            done
+            function(err) {
```
sduskis: Can you please test the `routing: cluster` path?

src/app-profile.js
```diff
@@ -0,0 +1,387 @@
...
+/**
+ * Formats a app profile options object into proto format.
+ *
+ * @private
+ *
+ * @param {object} options The options object.
+ * @returns {object}
+ *
+ * @example
+ * Family.formatAppProfile_({
```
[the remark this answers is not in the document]
kolodny: Done

test/instance.js
```diff
@@ -176,6 +180,159 @@ describe('Bigtable/Instance', function() {
...
+
+      instance.bigtable.request = function(config) {
+        assert.strictEqual(config.gaxOpts, options.gaxOptions);
+        done();
+      };
+
+      instance.createAppProfile(APP_PROFILE_NAME, options, assert.ifError);
+    });
+
+    describe('should respect the routing option when', function() {
```
sduskis: The sentence is incomplete.

system-test/bigtable.js
kolodny: We test that in setMetadata https://github.com/googleapis/nodejs-bigtable/pull/124/files#diff-107134a4c52159839ec2ea9b9fb78460R211

src/instance.js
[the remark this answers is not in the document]
kolodny: Good catch, I had them reversed

src/instance.js
kolodny: Fixed

src/instance.js
kolodny: Fixed

test/instance.js
kolodny: This is hard to do with bdd frameworks, it turns into something like:

should respect the routing option with
  an 'any' value  √
  a cluster value √

sduskis: ack

system-test/bigtable.js
sduskis: ack

src/app-profile.js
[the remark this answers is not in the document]
sduskis: ack.  "name vs id" is broken, but it's not something that we need to fix for this PR.

src/app-profile.js
```diff
@@ -0,0 +1,403 @@
...
+ *
+ * @example
+ * const Bigtable = require('@google-cloud/bigtable');
+ * const bigtable = new Bigtable();
+ * const instance = bigtable.instance('my-instance');
+ * const appProfile = instance.appProfile('my-appProfile');
+ *
+ * const metadata = {
+ *   description: 'My Updated App Profile',
+ *   routing: myCluster,
```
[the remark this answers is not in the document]
kolodny: Fixed

PR #125, opened 2018-05-05
added sample code for instance methods

vijay-qlogic:
Fixes #<107> (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass


samples/instances.js
```diff
@@ -15,37 +15,305 @@
...
+  // command to create a DEVELOPMENT instance
+  .command(`create-dev-instance`, `Create Instance`, {}, createDevInstance)
+  .example(`node $0 create-dev-instance`, `Creates a DEVELOPMENT instance`)
+
+  // command to list all instances and check 'ssd-instance' is listed
+  .command(`list-instances`, `Lists all instances`, {}, listInstances)
   .example(`node $0 list`, `Lists all instances in the current project.`)
+
+  // command to get the instance 'ssd-instance'
+  .command(`get-instance`, `Get the Instance`, {}, getInstance)
```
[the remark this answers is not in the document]
vijay-qlogic: Sure 'll update these changes by tomorrow

sduskis:
@jabubake: It turns out that async/await is not allowed in samples, since node.js v6 does not support that feature.  We'll have to revert to using the older style.

JustinBeckwith:
I actually have no problem with using async/await in samples :)  I think most of our customers are running node.js 8 or greater (with the exception of GCF), so I say async away.

samples/instances.js
```diff
@@ -15,38 +15,202 @@
...
-  bigtable
-    .getInstances()
-    .then(results => {
-      const instances = results[0];
-      console.log('Instances:');
-      instances.forEach(instance => console.log(instance.id));
-    })
-    .catch(err => {
-      console.error('ERROR:', err);
+  console.log(`\nCreating a PRODUCTION Instance`);
```
kolodny: Don't use a newline on console.log to have empty line, use an empty call to `console.log()` instead

samples/instances.js
```diff
@@ -15,38 +15,202 @@
...
   // [END bigtable_list_instances]
+
+  console.log(`\nGet Instance`);
+  // [START bigtable_get_instance]
+  try {
+    const [instance] = await bigtable.instance(instanceName).get();
+    console.log(`Instance Name: ${instance.name}`);
+    console.log(`Instance Meta: ${JSON.stringify(instance.metadata)}`);
+  } catch (err) {
+    console.log(`Error getting instance ${err}`);
```
kolodny: Use `console.error` and send the error object as its own argument. See https://github.com/googleapis/nodejs-bigtable/pull/139#discussion_r190997148 for more details

samples/instances.js
```diff
@@ -15,38 +15,202 @@
...
 
-function listInstances() {
-  // [START bigtable_list_instances]
-  // Imports the Google Cloud client library
-  const Bigtable = require('@google-cloud/bigtable');
+// Imports the Google Cloud client library
+const Bigtable = require('@google-cloud/bigtable');
 
-  // Creates a client
+async function runInstanceOperations(instanceName, clusterName, clusterName2) {
```
kolodny: clusterName1 clusterName2

sduskis:
Please update the top-level README.md with a link to this sample.

sduskis:
@kolodny, I think that this is ready for review.

samples/instances.js
```diff
@@ -15,38 +15,209 @@
...
+  const instanceOptions = {
+    clusters: [
+      {
+        name: clusterName1,
+        nodes: 3,
+        location: 'us-central1-f',
+        storage: 'ssd',
+      },
+    ],
+    type: 'PRODUCTION', // Optional as default type is PRODUCTION
```
kolodny: @sduskis uppercase or lowercase doesn't matter here, should we prefer upper or lower in the samples?

samples/instances.js
```diff
@@ -15,38 +15,209 @@
...
+  // Create production instance with given options
+  try {
+    const [instance] = await bigtable.createInstance(
+      instanceName,
+      instanceOptions
+    );
+
+    console.log(`Created Instance: ${instance.name}`);
+  } catch (err) {
+    console.error(new Error('Error creating prod-instance'));
```
kolodny: Change all these to:

```js
console.error('Error creating prod-instance:', err);
```

samples/instances.js
sduskis: I think that upper case looks better for "constants"

sduskis:
@kolodny and @jabubake, I think that all of the comments were addressed.  PTAL

jabubake:
LGTM

PR #127, opened 2018-05-06
fix for framing location value

### LINKED ISSUE #126: Instance.js createCluster method sets location incorrectly
The location property is not correctly framed as required in createCluster method.

Throws following error:

> Error: 3 INVALID_ARGUMENT: Error in field 'cluster' : Error in field 'location' : When parsing 'projects/projects/grass-clump-479/locations/us-central1-b' : Location name expected in the form 'projects/<project_id>/locations/<zone_id>'.

vijay-qlogic:
Fixes #126  (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass

### DIFF
--- a/src/cluster.js
+++ b/src/cluster.js
@@ -65,6 +65,11 @@ Cluster.getLocation_ = function(project, location) {
     return location;
   }
 
+  // in-case project has '/', split and pick last component
+  if (project.indexOf('/') > -1) {
+    project = project.split('/').pop();
+  }
+
   return format('projects/{project}/locations/{location}', {
     project: project,
     location: location,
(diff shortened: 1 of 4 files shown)


sduskis:
LGTM.  I'm going to ask that @kolodny reviews as well.

@vijay-qlogic, can you please update the branch?

src/cluster.js
```diff
@@ -52,7 +52,7 @@ function Cluster(instance, name) {
  *
  * @private
  *
- * @param {string} project The project.
+ * @param {string} project The project ID.
```
sduskis: This is right.  Please also add a check in the code to make sure that there isn't a '/'.  If there is, please do a slice and get the last component.  That should be done for both location and project.

PR #129, opened 2018-05-07
Handle buffers as literal atoms when used as a Filter.

### LINKED ISSUE #128: Filter on binary data
I can write the buffer and retrieve it again using decode:false but I cannot figure out how to filter on the value.

```js
const buf = Buffer.from('a468c3a669', 'hex');

// Throws Can't convert to RegExp String from unknown type
{
   value: buf
}

// Returns zero rows instead of throwing
{
  value: [
    buf
  ]
}

// Using binary string also returns zero rows
{
   value: buf.toString('binary')
}
```

#### Environment details

  - `@google-cloud/bigtable` version: 0.13.1

kolodny:
Fixes #128 

- [x] Tests and linter pass

### DIFF
--- a/src/filter.js
+++ b/src/filter.js
@@ -811,7 +811,7 @@ Filter.prototype.toProto = function() {
  * // Or you can provide a Buffer or an array of Buffers if you wish to match
  * // against specfic binary value(s).
  * //-
- * var userInputedFaces = [Buffer.from('.|.'), Buffer.from(':-)')]
+ * var userInputedFaces = [Buffer.from('.|.'), Buffer.from(':-)')];
  * var filter = [
  *   {
  *     value: userInputedFaces
--- a/src/filter.js
+++ b/src/filter.js
@@ -19,6 +19,7 @@
 var arrify = require('arrify');
 var createErrorClass = require('create-error-class');
 var extend = require('extend');
+var escapeStringRegexp = require('escape-string-regexp');
 var is = require('is');
 
 var Mutation = require('./mutation.js');
@@ -100,7 +101,7 @@ Filter.convertToRegExpString = function(regex) {
   }
 
   if (is.array(regex)) {
-    return '(' + regex.join('|') + ')';
+    return '(' + regex.map(Filter.convertToRegExpString).join('|') + ')';
   }
 
   if (is.string(regex)) {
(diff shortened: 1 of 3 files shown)


src/filter.js
```diff
@@ -803,6 +808,17 @@ Filter.prototype.toProto = function() {
  * ];
  *
  * //-
+ * // Or you can provide a Buffer or an array of Buffers if you wish to match
+ * // against specfic binary value(s).
+ * //-
+ * var userInputedFaces = [Buffer.from('.|.'), Buffer.from(':-)')]
```
stephenplusplus: Trailing `;`.

kolodny: Fixed

src/filter.js
```diff
@@ -111,6 +112,10 @@ Filter.convertToRegExpString = function(regex) {
     return regex.toString();
   }
 
+  if (Buffer.isBuffer(regex)) {
+    return escapeStringRegexp(regex.toString());
```
moander: This will create an utf-8 string. Using `.toString('binary')` will preserve the actual bytes

test/filter.js
```diff
@@ -65,6 +65,13 @@ describe('Bigtable/Filter', function() {
...
 
+    it('should convert an Array of buffers to a single string', function() {
+      var faces = [Buffer.from('.|.'), Buffer.from('=|=')];
+      var str = Filter.convertToRegExpString(faces);
+
+      assert.strictEqual(str, '(\\.\\|\\.|=\\|=)');
+    });
+
     it('should not do anything to a string', function() {
       var str1 = 'hello';
```
moander: Should test some special characters to make sure strings are not escaped

src/filter.js
moander:
```js
Buffer.from(Buffer.from('e6','hex').toString())
// <Buffer ef bf bd>

Buffer.from(Buffer.from('e6','hex').toString('binary'))
// <Buffer c3 a6>

Buffer.from(Buffer.from('e6','hex').toString('binary'), 'binary')
// <Buffer e6>
```

moander: Maybe use `\xFF` escaping for non-ASCII bytes?

PR #136, opened 2018-05-17
Updated with param details in comments

muraliQlogic:
Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass

### DIFF
--- a/src/cluster.js
+++ b/src/cluster.js
@@ -101,6 +101,10 @@ Cluster.getStorageType_ = function(type) {
  * Create a cluster.
  *
  * @param {object} [options] See {@link Instance#createCluster}.
+ * @param {function} [callback] The callback function.
+ * @param {?error} callback.err An error returned while making this
+ *     request.
+ * @param {object} callback.apiResponse The full API response.
  *
  * @example
  * const Bigtable = require('@google-cloud/bigtable');
@@ -223,7 +227,12 @@ Cluster.prototype.exists = function(gaxOptions, callback) {
  *
  * @param {object} [gaxOptions] Request configuration options, outlined here:
  *     https://googleapis.github.io/gax-nodejs/CallSettings.html.
- *
+ * @param {function} callback The callback function.
+ * @param {?error} callback.err An error returned while making this
+ *     request.
+ * @param {object} callback.metadata The metadata.
+ * @param {object} callback.apiResponse The full API response.
+ * 
  * @example
  * cluster.get(function(err, cluster, apiResponse) {
  *   // The `cluster` data has been populated.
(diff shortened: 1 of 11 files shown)


src/cluster.js
```diff
@@ -223,6 +227,11 @@ Cluster.prototype.exists = function(gaxOptions, callback) {
  *
  * @param {object} [gaxOptions] Request configuration options, outlined here:
  *     https://googleapis.github.io/gax-nodejs/CallSettings.html.
+ * @param {function} callback The callback function.
+ * @param {?error} callback.err An error returned while making this
+ *     request.
+ * @param {object} callback.metadata The metadata.
```
stephenplusplus: I believe we can remove this line, per the code:

```js
callback(err, err ? null : self, metadata);
```

Technically, `apiResponse` and `metadata` are the same here. I would vote to keep `apiResponse` as the chosen name for it, as that would be consistent with the other documentation.

src/family.js
```diff
@@ -144,6 +144,11 @@ Family.formatRule_ = function(ruleObj) {
  * Create a column family.
  *
  * @param {object} [options] See {@link Table#createFamily}.
+ * @param {function} callback The callback function.
```
stephenplusplus: These are the response parameters: https://github.com/muraliQlogic/nodejs-bigtable/blob/9049097a93e294576dd3cf0fed639678e1fa7c9a/src/table.js#L197-L199

src/instance.js
```diff
@@ -82,6 +82,11 @@ Instance.getTypeType_ = function(type) {
  * @param {object} [options] See {@link Bigtable#createInstance}.
  * @param {object} [options.gaxOptions]  Request configuration options, outlined
  *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
+ * @param {function} callback The callback function.
```
stephenplusplus: These are the response parameters: https://github.com/muraliQlogic/nodejs-bigtable/blob/9049097a93e294576dd3cf0fed639678e1fa7c9a/src/index.js#L447-L453

src/table.js
```diff
@@ -140,7 +140,10 @@ Table.createPrefixRange_ = function(start) {
  * @param {object} [options] See {@link Instance#createTable}.
  * @param {object} [options.gaxOptions]  Request configuration options, outlined
  *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
- *
+ * @param {function} [callback] The callback function.
```
stephenplusplus: These are the response parameters: https://github.com/muraliQlogic/nodejs-bigtable/blob/9049097a93e294576dd3cf0fed639678e1fa7c9a/src/instance.js#L252-L255

PR #153, opened 2018-05-29
Removed deletedCells from the row.data object

### LINKED ISSUE #149: Try to delete cells of a column family but don't work
In my column family "tcolumn", I have these data for a row and I want to delete cell "tcell": 
```
{
 tcolumn: {
   tcell: [ {
    value: "myvalue",
    labels: [],
    timestamp: "thetimestamp"
  } ]
 }
}
```
I try to do row.deleteCells(["tcolumn:tcell"]) but I still see my cell "tcell" in the data

#### Environment details
  - Node.js version: 8.9.4
  - npm version:
  - `@google-cloud/bigtable` version: 0.13.1

Thanks!

### LINKED ISSUE #151: row.get() merges in stale columns
nodejs-bigtable `v13.1`

When rehydrating a row, the deleted columns will still show up since we merge the old object with the new object

https://github.com/googleapis/nodejs-bigtable/blob/44a117212428587bf586e0c11e716182d871dec8/src/row.js#L733

Steps to repro

```js
// Insert row with two cells.
const ROW_KEY = 'mykey';
const rowToInsert = {
  key: ROW_KEY,
  data: {
    tcolumn: {
      tcell: 'myvalue',
    },
    col2: {
      foo: 'bar',
    },
  },
};
await table.insert(rowToInsert);
const [row] = await table.row(ROW_KEY).get();
// assert(row.data.tcolumn.tcell)

await row.deleteCells(['tcolumn:tcell']);
const [updatedRow] = await row.get();
// assert(row.data.tcolumn.tcell) // This should be false

// Do note that fetching a newly created row works since we don't merge in a stale row:

{
  await row.deleteCells(['tcolumn:tcell']);
  const [updatedRow] = await table.row(ROW_KEY).get();
  // assert(! row.data.tcolumn.tcell) // does not exists which is the expected result
}
```

kolodny:
Fixes #151 Fixes #149

This PR will keep the row data in sync with how it's represented in bigtable as best as possible. Consider the current behavior:

```js
const row = table.row('myRow');
setInterval(() => console.log(is.empty(row.data)), 10)
await row.get()

// The above will log true a couple of times then start logging false.
```

The row object has it's state changed whenever it hydrates it's data. By the same token, when removing cells the `deleteCells` method should remove data that no longer exists in the data object.

- [ ] Tests and linter pass

### DIFF
--- a/src/row.js
+++ b/src/row.js
@@ -460,13 +460,28 @@ Row.prototype.deleteCells = function(columns, gaxOptions, callback) {
     gaxOptions = {};
   }
 
+  columns = arrify(columns);
+
   var mutation = {
     key: this.id,
-    data: arrify(columns),
+    data: columns,
     method: Mutation.methods.DELETE,
   };
 
-  this.table.mutate(mutation, gaxOptions, callback);
+  this.table.mutate(mutation, gaxOptions, err => {
+    if (!err) {
+      columns.map(Mutation.parseColumnName).forEach(column => {
+        if (this.data[column.family]) {
+          if (column.qualifier) {
+            delete this.data[column.family][column.qualifier];
+          } else {
+            delete this.data[column.family];
+          }
+        }
+      });
+    }
+    callback(err);
+  });
 };
 
 /**
--- a/test/row.js
+++ b/test/row.js
@@ -692,6 +692,38 @@ describe('Bigtable/Row', function() {
 
       row.deleteCells(columns, gaxOptions, done);
     });
+
+    it('should remove the column qualifier from the data object', function() {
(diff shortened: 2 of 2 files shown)


sduskis:
If we do this, we'd have to extend all of the other methods (`save`, `filter`, `increment`) to do something similar.  It's a non-backwards compatible change, but I think that it's worth doing before Beta.  IMHO, users should always use the data from the `get` callback rather than use row.data.

sduskis:
I'm closing this, since we need to have an alternative method of fixing this problem.

PR #158, opened 2018-06-01
Switch src/ to use ES classes

### LINKED ISSUE #85: Convert to ES classes

kolodny:
Fixes #85 

- [x] Tests and linter pass

---

This PR should be ready to go. I believe this requires a major semver bump. I figured the test files shouldn't be updated in the same PR. The only thing that really needs to be pointed out is that classes shouldn't be instantiatable without the new keyword:

```js
const client1 = new Bigtable()
const client2 = Bigtable() // Error: Class constructor Foo cannot be invoked without 'new'
```

I was on the fence about requiring people to use the `new` keyword or not, and chose to make it optional. I did this by using a proxy:

```js
Bigtable = new Proxy(Bigtable, {
  apply(target, thisArg, argumentsList) {
    return new target(...argumentsList);
  },
});
```

Let me know if this should be removed. I didn't add that to the semi-private classes since the common case of creating one is something like `instance.table(...)` and not `new Table(...)`

As before, most of this was done via lebab so there may be some issues so please do go through it again to make sure nothing terrible is being done.

### DIFF
--- a/src/app-profile.js
+++ b/src/app-profile.js
@@ -382,7 +382,7 @@ AppProfile.prototype.setMetadata = function(metadata, gaxOptions, callback) {
     {
       client: 'BigtableInstanceAdminClient',
       method: 'updateAppProfile',
-      reqOpts: reqOpts,
+      reqOpts,
       gaxOpts: gaxOptions,
     },
     callback
--- a/src/app-profile.js
+++ b/src/app-profile.js
@@ -14,11 +14,11 @@
  * limitations under the License.
  */
 
-var common = require('@google-cloud/common');
-var is = require('is');
-var snakeCase = require('lodash.snakecase');
+const common = require('@google-cloud/common');
+const is = require('is');
+const snakeCase = require('lodash.snakecase');
 
-var Cluster = require('./cluster.js');
+const Cluster = require('./cluster.js');
 
 /**
  * Create an app profile object to interact with your app profile.
@@ -37,7 +37,7 @@ function AppProfile(instance, name) {
   this.bigtable = instance.bigtable;
   this.instance = instance;
 
-  var id = name;
+  let id = name;
 
   if (id.indexOf('/') === -1) {
     id = `${instance.id}/appProfiles/${name}`;
@@ -251,7 +251,7 @@ AppProfile.prototype.exists = function(gaxOptions, callback) {
  * });
  */
 AppProfile.prototype.get = function(gaxOptions, callback) {
-  var self = this;
+  const self = this;
 
   if (is.fn(gaxOptions)) {
     callback = gaxOptions;
@@ -286,7 +286,7 @@ AppProfile.prototype.get = function(gaxOptions, callback) {
  * });
  */
 AppProfile.prototype.getMetadata = function(gaxOptions, callback) {
-  var self = this;
+  const self = this;
 
   if (is.fn(gaxOptions)) {
     callback = gaxOptions;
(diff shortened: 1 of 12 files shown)


sduskis:
@kolodny, it looks like there are conflicts

kolodny:
Forced push an updated branch.

sduskis:
@kolodny, it looks like there are conflicts again.

sduskis:
ping @alexander-fenster @stephenplusplus @JustinBeckwith, can we please get a review on this?

src/app-profile.js
```diff
@@ -14,13 +14,11 @@
...
+const common = require('@google-cloud/common');
+const is = require('is');
+const snakeCase = require('lodash.snakecase');
 
-var common = require('@google-cloud/common');
-var is = require('is');
-var snakeCase = require('lodash.snakecase');
-
-var Cluster = require('./cluster.js');
+const Cluster = require('./cluster.js');
```
JustinBeckwith: don't bother with the `.js`

src/app-profile.js
```diff
@@ -35,359 +33,361 @@ var Cluster = require('./cluster.js');
...
+   * //-
+   * // If the callback is omitted, we'll return a Promise.
+   * //-
+   * appProfile.get().then(function(data) {
+   *   var appProfile = data[0];
+   *   var apiResponse = data[1];
+   * });
+   */
+  get(gaxOptions, callback) {
+    const self = this;
```
JustinBeckwith: Lets file a `type: cleanup` issue to come back and switch to arrow functions and drop the self stuff

src/cluster.js
```diff
@@ -14,11 +14,9 @@
  * limitations under the License.
  */
 
-'use strict';
-
-var common = require('@google-cloud/common');
-var format = require('string-format-obj');
-var is = require('is');
+const common = require('@google-cloud/common');
+const format = require('string-format-obj');
```
JustinBeckwith: You probably don't need this with string interpolation :)

src/family.js
```diff
@@ -14,17 +14,15 @@
  * limitations under the License.
  */
 
-'use strict';
-
-var common = require('@google-cloud/common');
-var createErrorClass = require('create-error-class');
-var is = require('is');
+const common = require('@google-cloud/common');
+const createErrorClass = require('create-error-class');
```
JustinBeckwith: You probably don't need this, since you can just extend the Error class.  Feel free to throw in a follow up issue.

src/index.js
```diff
@@ -805,6 +801,14 @@ Bigtable.Cluster = Cluster;
  */
 Bigtable.Instance = Instance;
 
+// Allow creating a `Bigtable` instance without using the `new` keyword.
+/* eslint-disable-next-line no-class-assign */
+Bigtable = new Proxy(Bigtable, {
```
JustinBeckwith: That is clever.  @ofrobots does this get us out of jail with the constructor change!?

src/app-profile.js
JustinBeckwith: Feel free to file a follow up issue

src/cluster.js
JustinBeckwith: again, cool to do as a followup

src/index.js
ofrobots: This is super cool!

I am on the fence 🤺though. If we are going to allow non-constructor calling, then we should be consistent about it and support it everywhere. So we ought to do it for `Table` etc. in this library, and in all our other libraries. Many – but not all – of these API classes are rarely constructed, so the performance overhead of the proxy must be considered.

This is a semver major anyway. Let's not try to be cute, as Ryan recently [lamented](https://www.youtube.com/watch?v=M3BM9TB-8yA) us as well.

JustinBeckwith: For this library, which is in alpha, I tend to agree.  For others that are GA...  I may take this opt-out approach :)

PR #159, opened 2018-06-01
Drop support for nodev4.

kolodny:
- [ ] Tests and linter pass

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,12 +3,10 @@ workflows:
   version: 2
   tests:
     jobs: &workflow_jobs
-      - node4:
+      - node6:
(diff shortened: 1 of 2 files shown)


stephenplusplus:
@JustinBeckwith @alexander-fenster @crwilcox should we automate this and send PRs to all of the repos?

JustinBeckwith:
👋  Usually, I'd be in favor of that.  In this case though, I'm hoping to lump together other breaking changes at the same time (like switching to es classes).  There's no real rush here to explicitly drop the v4 support, so I'd rather do this one-at-a-time as it makes sense. 

That having been said, a few of us are meeting today to figure this out.

PR #146, opened 2018-05-22
Issue 117 fix, updated setMetadata to use partialInstanceUpdate.

muraliQlogic:
Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

PR #167, opened 2018-06-05
Fix for Issue 117, setMetadata to use partialInstance Update

### LINKED ISSUE #117: Support labels on instances, and use PartialUpdateInstance

muraliQlogic:
Fixes #117  (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass


muraliQlogic:
This pull request contains changes for all review comments on [PR#146](https://github.com/googleapis/nodejs-bigtable/pull/146)

sduskis:
@kolodny: this PR replaces another one you reviewed previously.  PTAL.

src/instance.js
```diff
@@ -970,12 +971,26 @@ Instance.prototype.setMetadata = function(metadata, gaxOptions, callback) {
...
+    instance: extend({name: this.id}, metadata),
+    updateMask: {
+      paths: [],
+    },
+  };
+
+  const fieldsForMask = ['displayName', 'type', 'labels'];
+
+  fieldsForMask.forEach(field => {
+    if (reqOpts.instance[field]) {
```
kolodny: Hmm, setting the description to an empty string will probably fail since it's considered falsely. Looks like I did that in App Profile. Can you change that here? Test instead for  
`field in reqOpts.instance`

test/instance.js
```diff
@@ -1015,12 +1015,15 @@ describe('Bigtable/Instance', function() {
 
   describe('setMetadata', function() {
     it('should provide the proper request options', function(done) {
-      var metadata = {a: 'b'};
-      var expectedMetadata = extend({name: instance.id}, metadata);
+      var metadata = {displayName: 'updateDisplayName'};
+      var expectedMetadata = {
+        instance: extend({name: instance.id}, metadata),
```
kolodny: This is somewhat of a testing anti-pattern, just create the object that you expect, in this case  

```js
{
  name: instance.name,
  metadata: metadata,
}
```

muraliQlogic:
There is already a system test for instance update using setmetadata method. Do you want me to add something specific to update mask?

sduskis:
@muraliQlogic, can you please resolve the conflicts?

src/instance.js
```diff
@@ -17,6 +17,7 @@
 const common = require('@google-cloud/common');
 const extend = require('extend');
 const is = require('is');
+var snakeCase = require('lodash.snakecase');
```
kolodny: const

src/instance.js
```diff
@@ -938,12 +939,25 @@ class Instance {
       callback = gaxOptions;
       gaxOptions = {};
     }
+    var reqOpts = {
```
kolodny: const

kolodny:
@sduskis  I think that was addressed with https://github.com/googleapis/nodejs-bigtable/pull/167#issuecomment-395846526

src/instance.js
```diff
@@ -938,12 +939,25 @@ class Instance {
...
+      }
+    });
 
     this.bigtable.request(
       {
         client: 'BigtableInstanceAdminClient',
-        method: 'updateInstance',
-        reqOpts: extend({name: this.name}, metadata),
+        method: 'partialUpdateInstance',
+        reqOpts: reqOpts,
```
kolodny: Use shorthand for this, this line should change to just

```js
reqOpts,
```

PR #168, opened 2018-06-06
resolved name-vs-id issue for table.js

### LINKED ISSUE #137: name vs. id?
There are two concepts for which the terms "id" and "name" are used:

1. **Unique Identifier**: A string that uniquely identifies a resource within some namespace.  Examples:  "my-project", "my-instance" or "my-table".
2. **Fully Qualified Identifier**: A fully qualified identifier that uniquely describes a resource which includes the full name space.  Examples: "/project/my-project",  "/project/my-project/instances/my-instance" and  "/project/my-project/instances/my-instance/tables/my-table"

The Cloud Bigtable protobuf objects use the term "id" to describe the Unique Identifier, and the term "name" to describe the Fully Qualified Identifier.

The node.js client uses inconsistent nomenclature: 

- Projects use "id" is the unique identifier and "name" as the fully qualified identifier
- In instances, "name" to describes the Unique Identifier and "id" describes the Fully Qualified Identifier.  ([Instance Code](https://github.com/googleapis/nodejs-bigtable/blob/master/src/instance.js#L40), [Table Code](https://github.com/googleapis/nodejs-bigtable/blob/master/src/table.js#L58)).

When converting node.js variables to proto messages, the [code](https://github.com/googleapis/nodejs-bigtable/blob/master/src/table.js#L1355) looks like:

```
  var reqOpts = {
    -> name: this.id <-,
    deleteAllDataFromTable: true,
  };
```

This type of confusion, I believe, lead to bug #126.

I would like to use the protobuf as the standard for naming.

@kolodny, @stephenplusplus, @theacodes: do you have any thoughts on this subject?

vijay-qlogic:
Partially Fixes #137 (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass

### DIFF
--- a/src/instance.js
+++ b/src/instance.js
@@ -342,7 +342,7 @@ class Instance {
    * @see [Designing Your Schema]{@link https://cloud.google.com/bigtable/docs/schema-design}
    * @see [Splitting Keys]{@link https://cloud.google.com/bigtable/docs/managing-tables#splits}
    *
-   * @throws {error} If a name is not provided.
+   * @throws {error} If a id is not provided.
    *
    * @param {string} id Unique identifier of the table.
    * @param {object} [options] Table creation options.
--- a/src/instance.js
+++ b/src/instance.js
@@ -343,9 +343,9 @@ Instance.prototype.createCluster = function(name, options, callback) {
  * @see [Designing Your Schema]{@link https://cloud.google.com/bigtable/docs/schema-design}
  * @see [Splitting Keys]{@link https://cloud.google.com/bigtable/docs/managing-tables#splits}
  *
- * @throws {error} If a name is not provided.
+ * @throws {error} If a id is not provided.
  *
- * @param {string} name The name of the table.
+ * @param {string} id unique identifier for a table.
  * @param {object} [options] Table creation options.
  * @param {object|string[]} [options.families] Column families to be created
  *     within the table.
@@ -419,11 +419,11 @@ Instance.prototype.createCluster = function(name, options, callback) {
  *   const apiResponse = data[1];
  * });
  */
-Instance.prototype.createTable = function(name, options, callback) {
+Instance.prototype.createTable = function(id, options, callback) {
   var self = this;
 
-  if (!name) {
-    throw new Error('A name is required to create a table.');
+  if (!id) {
+    throw new Error('An id is required to create a table.');
   }
 
   options = options || {};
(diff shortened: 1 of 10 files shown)


sduskis:
Any place that uses `table.id` or `table.name` has to be changed as well.  Specifically, you have to update row.js in order to make this change work correctly.

sduskis:
Also, we need tests in row.js to make sure that requests are set up correctly.

sduskis:
I did another search of the codebase, and found that `family.js` and `instance.js` also have references to table names and ids.

`family.js` also references to `table.id` which needs to be changed to `table.name`.  It also has a variable called `tableName`, which is actually right, so you don't have to change that.

`instance.js` has a few references to table name and id.  The `createTable()` method has a parameter called `name` which should actually be `id`. `getTable()` has a `var name` which should actually be `var id`

vijay-qlogic:
changes in `family.js` and `instance.js` was planned to be done in separate pull-request but as you said this is breaking, I think we must do it in same pull-request.
Shall I?

sduskis:
I think that a breakdown is required.  The breakdown ought to be by object, and not by file.  The goal here ought to be to use the right variables for the Table object, where ever it's used.  If you just change the table.js file, then a whole bunch of things will fail.

src/instance.js
```diff
@@ -480,7 +480,7 @@ Instance.prototype.createTable = function(name, options, callback) {
     },
     function(...args) {
       if (args[1]) {
-        var table = self.table(args[1].name);
+        var table = self.table(args[1].id);
```
sduskis: `args` is a list of `Table` proto objects.  Table only has [`name`](https://github.com/googleapis/googleapis/blob/master/google/bigtable/admin/v2/table.proto#L99), not `id`.  I think that you need to use `args[1].name.split('/').pop()`

src/instance.js
```diff
@@ -886,8 +886,8 @@ Instance.prototype.getTables = function(options, callback) {
     function(...args) {
       if (args[1]) {
         args[1] = args[1].map(function(tableObj) {
-          var name = tableObj.name.split('/').pop();
-          var table = self.table(name);
+          var id = tableObj.id;
```
sduskis: `tableObj` is a proto object that only has name, and not id.  This should be:

```
 var id = tableObj.name.split('/').pop();
 var table = self.table(id);
```

vijay-qlogic:
@sduskis shall we also change constructor definition for `table.js` to accept `id` instead of `name`
As per changes done in `instance.js` `table` method

sduskis:
@vijay-qlogic That sounds like a good idea.  Please do so.

sduskis:
Mostly looks good.  It looks like there's a failure related to a node6 test that's unrelated to these changes.  We need to rerun the tests, somehow.

sduskis:
@vijay-qlogic, can you please resolve the conflicts?

PR #169, opened 2018-06-06
switch id and name in app-profile

muraliQlogic:
Partial fix for #137  (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass


test/app-profile.js
```diff
@@ -81,16 +81,16 @@ describe('Bigtable/AppProfile', function() {
     });
 
     it('should expand name into full resource path', function() {
-      assert.strictEqual(appProfile.id, APP_PROFILE_ID);
+      assert.strictEqual(appProfile.name, APP_PROFILE_ID);
```
sduskis: Please change `APP_PROFILE_ID` to `APP_PROFILE_NAME`

test/app-profile.js
```diff
@@ -155,7 +155,7 @@ describe('Bigtable/AppProfile', function() {
         options_,
         callback
       ) {
-        assert.strictEqual(name, appProfile.name);
+        assert.strictEqual(name, appProfile.id);
```
sduskis: please change the `name` field to be `id`

muraliQlogic: @sduskis  :Changes committed as per comments.

sduskis:
App Profiles did not exist in previous versions, otherwise, this would be a breaking change.

test/instance.js
```diff
@@ -773,9 +773,9 @@ describe('Bigtable/Instance', function() {
 
       instance.getAppProfiles(function(err, appProfiles, apiResponse) {
```
sduskis: Please also change

```
   var APP_PROFILE_NAME = 'my-app-profile';
```

to 
```
   var APP_PROFILE_ID = 'my-app-profile';
```

and also update any usages of that variable.

src/app-profile.js
```diff
@@ -35,18 +35,16 @@ var Cluster = require('./cluster.js');
...
  * const appProfile = instance.appProfile('my-app-profile');
  */
-function AppProfile(instance, name) {
+function AppProfile(instance, id) {
   this.bigtable = instance.bigtable;
   this.instance = instance;
 
-  var id = name;
-
   if (id.indexOf('/') === -1) {
```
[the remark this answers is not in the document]
kolodny: I don't have a strong opinion either way. However if we do keep it, I would suggest we compare the first bit to the current value of `instance.id` to make sure the user didn't mix up accounts or something, like `assert(this.id.includes(instance.id))`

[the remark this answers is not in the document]
sduskis: That check sounds reasonable.

system-test/bigtable.js
```diff
@@ -200,7 +200,7 @@ describe('Bigtable', function() {
       APP_PROFILE.getMetadata(function(err, metadata) {
         assert.strictEqual(
           metadata.name,
-          APP_PROFILE.id.replace('{{projectId}}', bigtable.projectId)
```
sduskis: Please also change `APP_PROFILE_NAME` to `APP_PROFILE_ID` in this file.

sduskis:
@muraliQlogic, can you please resolve the conflicts?

sduskis:
@muraliQlogic, can you please resolve the conflicts?

sduskis:
@muraliQlogic, can you please add a comment saying "I signed it!"

muraliQlogic:
I signed it!

PR #172, opened 2018-06-08
change id vs. name in Cluster

### LINKED ISSUE #137: name vs. id?

AVaksman:
Fixes #137  for cluster

### DIFF
--- a/src/app-profile.js
+++ b/src/app-profile.js
@@ -88,7 +88,7 @@ class AppProfile {
         appProfile.multiClusterRoutingUseAny = {};
       } else if (options.routing instanceof Cluster) {
         appProfile.singleClusterRouting = {
-          clusterId: options.routing.name,
+          clusterId: options.routing.name.split('/').pop(),
         };
         if (is.boolean(options.allowTransactionalWrites)) {
           appProfile.singleClusterRouting.allowTransactionalWrites =
--- a/src/app-profile.js
+++ b/src/app-profile.js
@@ -88,7 +88,7 @@ class AppProfile {
         appProfile.multiClusterRoutingUseAny = {};
       } else if (options.routing instanceof Cluster) {
         appProfile.singleClusterRouting = {
-          clusterId: options.routing.name.split('/').pop(),
+          clusterId: options.routing.id,
         };
         if (is.boolean(options.allowTransactionalWrites)) {
           appProfile.singleClusterRouting.allowTransactionalWrites =
--- a/src/cluster.js
+++ b/src/cluster.js
@@ -38,7 +38,7 @@ class Cluster {
 
     var name;
 
-    if (id.indexOf('/') > -1) {
+    if (id.includes('/')) {
       if (id.includes(`${instance.name}/clusters/`)) {
         name = id;
       } else {
--- a/src/cluster.js
+++ b/src/cluster.js
@@ -23,7 +23,7 @@ const is = require('is');
  *
  * @class
  * @param {Instance} instance The parent instance of this cluster.
- * @param {string} name Name of the cluster.
+ * @param {string} id Id of the cluster.
  *
  * @example
  * const Bigtable = require('@google-cloud/bigtable');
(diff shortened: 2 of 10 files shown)


src/app-profile.js
```diff
@@ -91,7 +91,7 @@ AppProfile.formatAppProfile_ = function(options) {
       appProfile.multiClusterRoutingUseAny = {};
     } else if (options.routing instanceof Cluster) {
       appProfile.singleClusterRouting = {
-        clusterId: options.routing.name,
+        clusterId: options.routing.name.split('/').pop(),
```
sduskis: why not just `name.id`?  `name` is a `Cluster`

sduskis:
The following test is failing here and elsewhere:

```
  1) Bigtable
       request
         makeRequestStream
           should destroy the stream with GAX error:
```

We need to fix that before merging anything else.

sduskis:
@AVaksman, can you please resolve the conflicts?

kolodny:
@AVaksman the `should destroy the stream with GAX error:` error is now fixed

sduskis:
Yes, assuming that the conflicts are resolved.

sduskis:
@AVaksman, code coverage dropped.  Can you please add more tests?  I think that the error you throw in the cluster constructor needs a new test

src/cluster.js
```diff
@@ -32,18 +32,17 @@ const is = require('is');
...
     this.bigtable = instance.bigtable;
     this.instance = instance;
 
-    let id = name;
-
-    if (!id.includes('/')) {
-      id = `${instance.name}/clusters/${name}`;
+    if (id.indexOf('/') > -1) {
+      throw new Error(
+        'Cluster Id is not formatted correclty. Just `mycluster` rather than `projects/myproject/instances/myinstance/clusters/mycluster`'
```
stephenplusplus: Across our different APIs, we accept the dual format. If the user is over-specific, what's the harm?

sduskis: One way or another, there needs to be error checking.  We'd have to check that `id.contains(instance.name)` here.

AVaksman: Perhaps error prone, a lot more room to make a typo while spelling out
`"projects/myproject/instances/myinstance/clusters/mycluster"`
rather then just passing `"mycluster"` and let the client convert it into fully qualified name?

stephenplusplus: That's a fine idea. Allowing the upstream API to communicate the error at the time of an operation is usually an okay fallback. But in this case, I would agree that since you have the `instance.name` available, you might as well catch the error asap.

stephenplusplus: The user should be allowed to provide any input that we can make sense of. We can handle both the name or ID. We can encourage the easy one, but we shouldn't penalize the user for being verbose.

sduskis: Ack. We'll change the logic to include a correctness check that makes sure that the parent and full name are in sync.

sduskis:
Code coverage still needs to be updated.

sduskis:
AppVeyor's queue seems to be stuck.  I'm going to merge this anyway.

Issue #178, opened 2018-06-11
node.js v6 cleanup


kolodny:
This is just a checklist of items that should be done in a followup for #158 

- [x] Remove `.js.` on the require calls. Double check with @ofrobots if we want to do this since Ryan lamented not being explicit with `require`s in his video
- [x] Remove any mention of `const self = this` by switching to arrow functions
- [x] Remove `string-format-obj` as a dep
- [x] Remove `create-error-class` as a dep

---

- [ ] Migrate tests to use ES6 targeting nodev6

ofrobots:
> Remove .js. on the require calls. Double check with @ofrobots if we want to do this since Ryan lamented not being explicit with requires in his video

That ship 🛳 has long sailed :). Omitting the extension is now necessary to switch a project to typescript, so I have no objections 👍.

sduskis:
@kolodny, is there more to come, or is this bug done?

kolodny:
The only item left is to migrate the tests over to es6.

sduskis:
@kolodny, I'm going to close this.  Pleas reopen if this is still a concern.

PR #180, opened 2018-06-11
Cleanup

kolodny:
Addresses parts of #178

- [ ] Tests and linter pass


src/filter.js
```diff
@@ -15,19 +15,21 @@
...
-const Mutation = require('./mutation.js');
+const Mutation = require('./mutation');
 
 /**
  * @private
  */
-const FilterError = createErrorClass('FilterError', function(filter) {
-  this.message = `Unknown filter: ${filter}.`;
-});
+class FilterError extends Error {
```
stephenplusplus: The one issue with this is we lose the `.name` property on the function, which breaks this (possibly documented) behavior:

```js
if (err.name === 'FilterError') {
 // handle filter error
}
```

kolodny: Fixed, good catch. I've updated the other ones too

PR #186, opened 2018-06-13
resolved name vs. id issue for family.js

vijay-qlogic:
Fixes partially #137 (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass

### DIFF
--- a/src/row.js
+++ b/src/row.js
@@ -188,7 +188,7 @@ class Row {
     options = options || {};
 
     families.forEach(function(family) {
-      const familyData = (data[family.name] = {});
+      const familyData = (data[family.id] = {});
 
       family.columns.forEach(function(column) {
         const qualifier = Mutation.convertFromBytes(column.qualifier);
--- a/src/row.js
+++ b/src/row.js
@@ -16,22 +16,25 @@
 
 const arrify = require('arrify');
 const common = require('@google-cloud/common');
-const createErrorClass = require('create-error-class');
 const dotProp = require('dot-prop');
 const extend = require('extend');
 const flatten = require('lodash.flatten');
 const is = require('is');
 
-const Filter = require('./filter.js');
-const Mutation = require('./mutation.js');
+const Filter = require('./filter');
+const Mutation = require('./mutation');
 
 /**
  * @private
  */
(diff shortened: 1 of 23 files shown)


src/row.js
```diff
@@ -188,7 +188,7 @@ class Row {
     options = options || {};
 
     families.forEach(function(family) {
-      const familyData = (data[family.name] = {});
+      const familyData = (data[family.id] = {});
```
sduskis: I'd revert this one, for now.  This changes the way rows are read.  The other changes changes aspects of table and family administration.

Once we're ready for this change, this would be a backwards incompatible change for row reads, which is a bit more tricky to convey to users.

src/table.js
```diff
@@ -699,18 +699,18 @@ class Table {
...
-   * @param {string} name The family name.
+   * @param {string} id The family unique identifier.
    * @returns {Family}
    *
    * @example
    * const family = table.family('my-family');
    */
-  family(name) {
-    if (!name) {
+  family(id) {
```
sduskis: Please also change any uses of `family()` to only get the id instead of the full name.

system-test/bigtable.js
```diff
@@ -388,18 +388,19 @@ describe('Bigtable', function() {
         assert.ifError(err);
         assert.strictEqual(families.length, 3);
         assert(families[0] instanceof Family);
-        assert.strictEqual(families[0].name, FAMILY.name);
+        assert.strictEqual(families[2].id, FAMILY.id);
```
sduskis: Why was there a change from `families[0]` to `families[2]`?

test/row.js
```diff
@@ -447,7 +447,7 @@ describe('Bigtable/Row', function() {
 
     var families = [
       {
-        name: 'test-family',
+        id: 'test-family',
```
sduskis: This should be reverted so that we can do a separate PR for families in ReadRows

system-test/bigtable.js
vijay-qlogic: because 2 families("follow", "traits") are already created above(ln. no. 61) in before section. And 3rd family("presidents") is created here, so index should be 2.

Issue #171, opened 2018-06-07
utf-8/encoding bug 


moander:
Please see my comments here https://github.com/googleapis/nodejs-bigtable/pull/129#pullrequestreview-118364984

Problem 1:
Binary values in key will get unexpected values using binary values in buffers. 

Problem 2:
Binary value  filters will not return any matches

Problem 3:
TransformError from getRows after inserting unique keys 0xE6 and 0xF8

#### Environment details

  - OS: macOS 10.13.4
  - Node.js version: v8.11.2
  - npm version: 5.6.0
  - `@google-cloud/bigtable` version: 0.14.0

#### Steps to reproduce

```js
async function test() {
    await bt.createTable('test', {
        families: [
            'c'
        ]
    }).catch(swallowCode(6));

    const table = bt.table('test');

    await table.mutate([
        {
            key: 'a',
            method: 'insert',
            data: {
                c: {
                    test: Buffer.from('x', 'binary')
                }
            }
        },
        {
            key: 'b',
            method: 'insert',
            data: {
                c: {
                    test: Buffer.from('æ', 'binary')
                }
            }
        },
        {
            key: Buffer.from('æ'),
            method: 'insert',
            data: {
                c: {
                    test: Buffer.from('æ', 'binary')
                }
            }
        },
        {
            key: Buffer.from('æ', 'binary'),
            method: 'insert',
            data: {
                c: {
                    test: Buffer.from('æ', 'binary')
                }
            }
        },

    ]);

    function printRow(row) {
        console.log(row.id, Buffer.from(row.id), row.data.c.test[0].value);
    }

    let rows;

    console.log('## no filter');
    [rows] = await table.getRows({
        decode: false,
        filter: [
            {
                column: {
                    cellLimit: 1
                }
            },
        ]
    });
    rows.forEach(printRow);

    console.log('## æ utf-8 filter');
    [rows] = await table.getRows({
        decode: false,
        filter: [
            {
                column: {
                    cellLimit: 1
                }
            },
            {
                value: 'æ'
            }
        ]
    });
    rows.forEach(printRow);


    console.log('## æ binary filter');
    [rows] = await table.getRows({
        decode: false,
        filter: [
            {
                column: {
                    cellLimit: 1
                }
            },
            {
                value: Buffer.from('æ', 'binary')
            }
        ]
    });
    rows.forEach(printRow);

    console.log('## x binary filter');
    [rows] = await table.getRows({
        decode: false,
        filter: [
            {
[...]

PR #187, opened 2018-06-13
Handle non utf8 buffer filters

kolodny:
Addresses #171

- [x] Tests and linter pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "google-auto-auth": "^0.10.0",
     "google-gax": "^0.16.0",
     "is": "^3.0.1",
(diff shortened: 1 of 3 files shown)


kolodny:
@moander patching in this PR seems to fix the filtering issue you were experiencing

LMK if this solves your use case. Unfortunately we need to attempt to convert the filter to a string since we need to escape non and special chars in the buffer. Also note that while a string can be latin1, that doesn't mean that it isn't valid utf8 in which case I suspect this issue will happen again. We can potentially expose an option to `rawBuffer` and not try to escape any characters

moander:
Thanks @kolodny I test it tomorrow :)

PR #193, opened 2018-06-13
Update to google-gax 0.17.1

kolodny:
- [x] Tests and linter pass

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -17,9 +17,9 @@
 const arrify = require('arrify');
 const common = require('@google-cloud/common');
 const extend = require('extend');
+const {GrpcClient} = require('google-gax');
 const GrpcService = require('@google-cloud/common-grpc').Service;
 const googleAuth = require('google-auto-auth');
-const gax = require('google-gax');
 const is = require('is');
 const retryRequest = require('retry-request');
 const streamEvents = require('stream-events');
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -19,7 +19,7 @@ const common = require('@google-cloud/common');
 const extend = require('extend');
 const GrpcService = require('@google-cloud/common-grpc').Service;
 const googleAuth = require('google-auto-auth');
-const grpc = require('google-gax').grpc().grpc;
+const gax = require('google-gax');
```
JustinBeckwith: Not sure if you use gax elsewhere in the file, but you could shorten this all to `const {GrpcClient} = require('google-gax');`

sduskis:
TypeError: gax.GrpcClient is not a constructor

src/index.js
kolodny: Good call, changed

sduskis:
@kolodny, did you manually change generated code?

kolodny:
@sduskis I did, @JustinBeckwith will that be a problem when the code is regenerated?

stephenplusplus:
@JustinBeckwith @crwilcox how would we get this change into the generator? I'm blocked on this for https://github.com/googleapis/nodejs-spanner/pull/220 as well.

JustinBeckwith:
@alexander-fenster has the latest on this :)

PR #194, opened 2018-06-13
Adding clusterId format verification and unit tests

### LINKED ISSUE #189: New errors for name / id checks in constructors
Constructors will generally do things like this:

```
  constructor(parent, id) {
    this.parent = parent;

    let name = id;
    if (!name.includes('/')) {
      name = `${parent.name}/type/${id}`;
    }

    this.id = name.split('/').pop();
    this.name = name;
  }
````

If `name.includes('/')`, we should check to make sure that we add the following check, and throw an error if this fails:
```
name.startsWith(`${parent.name}/type/`)
```

This logic (and test case) should be added for at least the following: instance, cluster, table, row, family.

### PR DESCRIPTION
**Pull request by AVaksman**
Fixes #189 for cluster

### DIFF
--- a/test/cluster.js
+++ b/test/cluster.js
@@ -158,16 +158,15 @@ describe('Bigtable/Cluster', function() {
     });
 
     it('should accept fully expanded name as clusterId', function() {
-      let id_ = `${INSTANCE.name}/clusters/${CLUSTER_ID}`;
-      var cluster2 = new Cluster(INSTANCE, id_);
-      assert.strictEqual(cluster2.name, id_);
+      var cluster2 = new Cluster(INSTANCE, CLUSTER_NAME);
+      assert.strictEqual(cluster2.name, CLUSTER_NAME);
       assert.strictEqual(cluster2.id, CLUSTER_ID);
     });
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by sduskis** [comment]
We should really have this for all types, and not just clusters, as per #189.  @kolodny removed the cluster check in #188.

**Comment by AVaksman** [comment]
I wanted verify that unit tests look good in this PR to follow this example for all types

File: test/cluster.js [hunk 1]
```diff
@@ -157,6 +157,20 @@ describe('Bigtable/Cluster', function() {
       cluster.create(options, done);
     });
 
+    it('should accept fully expanded name as clusterId', function() {
+      let id_ = `${INSTANCE.name}/clusters/${CLUSTER_ID}`;
```
sduskis: shouldn't this just be `id =`?  Do we need the `_`?

sduskis: Can't we just use CLUSTER_NAME instead of _id?  Also, why not use `var` like everywhere else?

AVaksman: done

sduskis:
This was superceded by #198

AVaksman:
@sduskis yes, thank you

PR #202, opened 2018-06-21
New errors for name / id checks in family constructor

muraliQlogic:
Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass

### DIFF
--- a/test/family.js
+++ b/test/family.js
@@ -85,19 +85,18 @@ describe('Bigtable/Family', function() {
       var family = new Family(TABLE, FAMILY_ID);
       assert.strictEqual(family.name, FAMILY_NAME);
     });
-  });
 
-  describe('formatName_', function() {
-    it('should format the column family name', function() {
-      var formatted = Family.formatName_(TABLE.name, FAMILY_ID);
-
-      assert.strictEqual(formatted, FAMILY_NAME);
+    it('should leave full family names unaltered and localize the id from the name', function() {
+      var family = new Family(TABLE, FAMILY_NAME);
+      assert.strictEqual(family.name, FAMILY_NAME);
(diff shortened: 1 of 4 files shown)


test/table.js
```diff
@@ -373,11 +374,11 @@ describe('Bigtable/Table', function() {
       };
 
       table.family = function(name) {
```
sduskis: This should be function(id)

test/family.js
```diff
@@ -85,19 +85,18 @@ describe('Bigtable/Family', function() {
...
+      assert.strictEqual(family.name, FAMILY_NAME);
+      assert.strictEqual(family.id, FAMILY_ID);
     });
 
-    it('should not re-format the name', function() {
-      var formatted = Family.formatName_(TABLE.name, FAMILY_ID);
-
-      assert.strictEqual(formatted, FAMILY_NAME);
+    it('should throw if family id in wrong format', function() {
+      var id = `columnFamilies/${FAMILY_ID}`;
```
sduskis: The id here should just be FAMILY_ID

muraliQlogic: @sduskis  can you please clarify this comment ,id here is a sample of wrongly formatted family id for which we are expecting an Error.

test/table.js
muraliQlogic: Changed.

test/family.js
sduskis: I spoke with Alex about this.  Can you please chase to to something like: `bad/format/${FAMILY_ID}` or `/project/bad-project/instances/bad-instance/columnFamiles/${FAMILY_ID}` so that the intention is clearer?

PR #203, opened 2018-06-22
Fix rowKey being force stringified to utf8

### LINKED ISSUE #185: Don't autostringify the rowKey when decode is false
See https://github.com/googleapis/nodejs-bigtable/issues/171 for more context

The issue at hand is that the library aggressively converts the rowKey buffer to a string even when the user explicitly sets decode to false. This causes issues where the encoding scheme the user is using doesn't match the default scheme of utf8

kolodny:
Fixes #185 

The key should be treated as a black box of bytes and not as a utf8 encoded string. 
 
`Buffer.from('ø', 'binary').toString() === Buffer.from('æ', 'binary').toString()`  
evaluates to true so two consecutive keys like this will cause the chunkTransformer to throw an error. This PR also allows choosing a custom encoding since the `row.data.family` is a hashmap and needs to stringify the value so we can't just leave it as a buffer with `decode = false`

- [x] Tests and linter pass

### DIFF
--- a/src/chunktransformer.js
+++ b/src/chunktransformer.js
@@ -132,7 +132,7 @@ class ChunkTransformer extends Transform {
    * @private
    */
   reset() {
-    this.prevRowKey = '';
+    this.prevRowKey = null;
     this.family = {};
     this.qualifiers = [];
     this.qualifier = {};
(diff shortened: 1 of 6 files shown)


moander:
Master branch now works as expected :) Thank you very much!

PR #208, opened 2018-06-23
fix #196 should not call replaceProjectID token when projectid is passed in option

### LINKED ISSUE #196: Replace {{projectId}} with actual project.
The nodejs-bigtable client uses {{projectId}} as a placeholder.  Ideally, we use `process.env.GOOGLE_CLOUD_PROJECT` or `process.env.GCLOUD_PROJECT` for our system tests, and otherwise force users to set projectId.

ajaaym:
Fixes #196  (it's a good idea to open an issue first for discussion)

- [x] Tests and linter pass

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -690,6 +690,7 @@ describe('Bigtable', function() {
           [CONFIG.method]: common.util.noop,
         };
       });
+
       it('should replace the project ID token', function(done) {
         var replacedReqOpts = {};
 
@@ -700,6 +701,7 @@ describe('Bigtable', function() {
 
           return replacedReqOpts;
         };
+
         bigtable.api[CONFIG.client][CONFIG.method] = {
           bind: function(gaxClient, reqOpts) {
             assert.strictEqual(reqOpts, replacedReqOpts);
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -663,6 +663,33 @@ describe('Bigtable', function() {
...
+      beforeEach(function() {
+        bigtable = new Bigtable();
+        bigtable.getProjectId_ = function(callback) {
+          callback(null, PROJECT_ID);
+        };
+
+        bigtable.api[CONFIG.client] = {
+          [CONFIG.method]: common.util.noop,
+        };
+      });
```
stephenplusplus: Can you leave a new line to separate all `beforeEach`, `describe`, `it`, etc. blocks?

test/index.js
```diff
@@ -673,7 +700,6 @@ describe('Bigtable', function() {
 
           return replacedReqOpts;
         };
-
```
stephenplusplus: Please leave the removed new lines in.

test/index.js
sduskis: @stephenplusplus, I fixed the new lines.  PTAL.

PR #222, opened 2018-06-27
Added @example tags to src/instance.js

### LINKED ISSUE #157: @example Comments should refer to sample code

```
/**
 ...
*/
```

vijay-qlogic:
Fixes #157 for Instance.js (it's a good idea to open an issue first for discussion)

- [x] Tests and linter pass


sduskis:
@jabubake: Do we need tests for these snippets?

jabubake:
@sduskis : Yes, ensuring these are tested will help maintain them.

Since these snippets are for the API methods, one option to minimize the number of tests is to ensure that the [integration tests](https://github.com/googleapis/nodejs-bigtable/blob/master/system-test/bigtable.js) for the library leverage these snippets as test cases.

stephenplusplus:
Samples actually have their own testing files and conventions: https://github.com/googleapis/nodejs-bigtable/tree/master/samples. See the `system-test` directory.

There is already a file, `instances.js`. Maybe these new tests should be moved into that file, then tested from `system-test/instances.test.js`?

jabubake:
@stephenplusplus I agree there is overlap here : in this case, `samples/instance.js` provides a clean way to run through basic use cases in a single format. The snippets are intended to focus on the the different ways of  calling the API w/ promise, callback etc. This is a little unique to node.js, so i'm fine with the additional source file here vs cluttering `samples/instance.js` : we can maybe remove the cli from the `document-snippets/instance.js` and just have tests written against it, so that developers do not run into weird states running the CLI  commands out of order.

samples/document-snippets/instance.js
```diff
@@ -0,0 +1,382 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const Bigtable = require('@google-cloud/bigtable');
+const bigtableClient = new Bigtable();
+
+function newInstance(instanceId) {
```
sduskis: Please call these functions something similar to the tag.  In this case, please use `createInstance`.

samples/document-snippets/instance.js
```diff
@@ -0,0 +1,382 @@
...
+
+  // creates a new Instance
+  instance
+    .create(options)
+    .then(result => {
+      let newInstance = result[0];
+      // let operations = result[1];
+      // let apiResponse = result[2];
+
+      console.log(`Created Instance: ${newInstance.name}`);
```
sduskis: Please use `id` here.

samples/document-snippets/instance.js
```diff
@@ -0,0 +1,382 @@
...
+    storage: 'ssd',
+  };
+
+  instance
+    .createCluster(clusterId, options)
+    .then(result => {
+      const newCluster = result[0];
+      // const operations = result[1];
+      // const apiResponse = result[2];
+      console.log(`Cluster created: ${newCluster.name}`);
```
sduskis: Please use `id` here.

samples/document-snippets/instance.js
```diff
@@ -0,0 +1,382 @@
...
+
+      console.log(`Created Instance: ${newInstance.name}`);
+    })
+    .catch(err => {
+      console.error('Error creating prod-instance:', err);
+    });
+  // [END bigtable_create_instance]
+}
+
+function newCluster(clusterId, instanceId) {
```
sduskis: Please call this `createCluster`.

samples/document-snippets/instance.js
```diff
@@ -0,0 +1,382 @@
...
+      // const apiResponse = result[2];
+      console.log(`Cluster created: ${newCluster.name}`);
+    })
+    .catch(err => {
+      console.error('Error creating cluster: ', err);
+    });
+  // [END bigtable_create_cluster]
+}
+
+function newAppProfile(appProfileId, clusterId, instanceId) {
```
sduskis: Please call this `createAppProfile`.  Also, please reverse the order of parameters.  This should be `instanceId, clusterId, appProfileId`.

samples/document-snippets/instance.js
sduskis: Also, please start the parameters from the "greatest" thing.  In this case, please use `instanceId, clusterId`

samples/document-snippets/instance.js
```diff
@@ -0,0 +1,382 @@
...
+    .catch(err => {
+      console.error('Error geting AppProfiles: ', err);
+    });
+  // [END bigtable_get_app_profiles]
+}
+
+function getMetaData(instanceId) {
+  const instance = bigtableClient.instance(instanceId);
+
+  // [START bigtable_get_imeta]
```
sduskis: Please use the full name in the tag, specifically, `bigtable_get_instance_metadata`.

samples/document-snippets/instance.js
```diff
@@ -0,0 +1,382 @@
...
+  //   autoPaginate: true
+  // };
+
+  instance
+    .getTables(options)
+    .then(result => {
+      console.log(`Tables:`);
+      let tables = result[0];
+      tables.forEach(t => {
+        console.log(t.name);
```
sduskis: Please use `t.id`

samples/document-snippets/instance.js
```diff
@@ -0,0 +1,382 @@
...
+      });
+    })
+    .catch(err => {
+      console.error('Error geting Tables: ', err);
+    });
+  // [END bigtable_get_tables]
+}
+
+function setMetaData(instanceId) {
```
sduskis: Please call this method `updateInstance`

vijay-qlogic: Done
https://github.com/googleapis/nodejs-bigtable/pull/222/commits/174f67cadb820ec8b26349131f69e66a403fc720

sduskis:
@stephenplusplus, do you know how to generate API documentation?  That's important in testing this PR.

stephenplusplus:
That's all done behind the scenes from me. @jmdobry can probably help.

vijay-qlogic:
I searched for generating API docs that pull in examples from other files, but no where in **JSDoc** guide found `region_tag` being used to for same.
@jmdobry can you please help here.

jmdobry:
If your changes were on a branch of the nodejs-bigtable repo itself, I could stage the API ref docs to test their use of `region_tag`. I don't have a way to test when the changes are on a fork.

sduskis:
@jmdobry, can you potentially use  `vijay-qlogic:tag-example-comments` as the branch?

vijay-qlogic:
@jmdobry I have referred [nodejs-spanner](https://github.com/googleapis/nodejs-spanner/blob/72efac10bcab961732cca234312908fa1b5bc3dd/src/batch-transaction.js#L116) to include `region_tag`, But I see Document generation in `nodejs-spanner` repository as well

sduskis:
@jmdobry, I created a new branch on `nodejs-bigtable` called `docs-test`.  Can you please try it out?

vijay-qlogic:
Hi Jason,

Can you please verify the "example" section while generating the API Docs.
Because in my case "region_tag" do not pulls the example code.

Sharing the screen-shot here.

sduskis:
@vijay-qlogic, this was verified.

PR #250, opened 2018-07-18
test: use strictEqual in tests

JustinBeckwith:


sduskis:
@kinwa91, this broke the system tests: https://circleci.com/gh/googleapis/nodejs-bigtable/7361.  Do you have any idea on how we can fix that?

PR #251, opened 2018-07-18
Fixing the system test.

sduskis:


system-test/bigtable.js
```diff
@@ -554,8 +554,8 @@ describe('Bigtable', function() {
         assert.ifError(err);
         var maxAge = metadata.gcRule.maxAge;
 
-        assert.strictEqual(maxAge.seconds, rule.age.seconds);
-        assert.strictEqual(maxAge.nanas, rule.age.nanas);
+        assert.strictEqual(maxAge.seconds, rule.age.seconds.toString());
+        assert.strictEqual(maxAge.nanos, rule.age.nanos.toString());
```
stephenplusplus: These tests fail locally for me. I added some logging:

```js
typeof maxAge.seconds string
typeof maxAge.nanos number
typeof rule.age.seconds number
typeof rule.age.nanos number
```

Can we make `maxAge.seconds` a Number?

sduskis: @stephenplusplus, maxAge is a protobuf Duration object, which should be a number according to the [protobuf definition](https://github.com/google/protobuf/blob/master/src/google/protobuf/duration.proto#L108).  I'm not sure why it's returning as a string.

PR #253, opened 2018-07-23
Use assert.deepStrictEqual instead of deepEqual.

sduskis:
`deepEqual` is causing some issues in PR #247, where the linter seems to prefers `deepStrinctEqual`.  Also, this fixes a strange System test problem from PR #251.


sduskis:
Most lint issues are resolved.  The only one left is:

>  37:2  error  Async functions are not supported until Node.js 7.6.0. The configured version range is '>=6.0.0'  node/no-unsupported-features/es-syntax

@stephenplusplus or @kolodny, do you have any ideas on how to fix it?

PR #284, opened 2018-08-28
Added example region-tags for cluster object

### LINKED ISSUE #157: @example Comments should refer to sample code

vijay-qlogic:
Fixes #157 partially for cluster.js

- [ ] Tests and linter pass


sduskis:
@dpebot merge on green

PR #288, opened 2018-08-29
Added example region-tags for app-profile object

### LINKED ISSUE #157: @example Comments should refer to sample code

vijay-qlogic:
Fixes #157 partially for app-profile object (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass


samples/document-snippets/app-profile.js
```diff
@@ -0,0 +1,136 @@
...
+const snippets = {
+  create: (instanceId, appProfileId) => {
+    // [START bigtable_create_app_profile]
+    const Bigtable = require('@google-cloud/bigtable');
+    const bigtable = new Bigtable();
+    const instance = bigtable.instance(instanceId);
+    const appProfile = instance.appProfile(appProfileId);
+
+    appProfile
+      .create()
```
sduskis: Will this work without setting the appProfile type?

vijay-qlogic: I feel it should work. 
I saw the constructor for appProfile, 'type' is not required

sduskis: I feel like it won't work. `type` is not required for `getMetadata()`, but `create()` shouldn't work  without it.

sduskis:
@vijay-qlogic, can you please fix the `create()` logic?  @AVaksman, FYI

JustinBeckwith:
@vijay-qlogic @sduskis this PR is getting a little moldy.  Can we close this out?  Is there more work to be done here?

JustinBeckwith:
@vijay-qlogic @sduskis this PR is getting a little moldy.  Can we close this out?  Is there more work to be done here?

PR #337, opened 2018-10-23
Fix "hello world" samples

hegemonic:
+ `index.js`: Use the API correctly (I introduced this bug in 13e3d769962f9301ad73253968d2ae394f7d9766)
+ `index.v6.js`: Make some of the changes from 13e3d769962f9301ad73253968d2ae394f7d9766 (but don't introduce the new bug)

The sample isn't creating the column family for some reason, which means it doesn't work yet, but this change gets it closer to working. Solomon is looking into the column family issue separately.

- [x] Tests and linter pass

### DIFF
--- a/samples/hello-world/index.js
+++ b/samples/hello-world/index.js
@@ -39,8 +39,7 @@ const getRowGreeting = row => {
 (async () => {
   try {
     // [START connecting_to_bigtable]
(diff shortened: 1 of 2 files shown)


hegemonic:
@sduskis Here's the PR I promised!

PR #339, opened 2018-10-24
Fixing the Hello world

sduskis:
The json for family name in the system test was incorrect.  It was "id", but should have been "name"


sduskis:
@googleapis/yoshi-nodejs, I can use some help here.  It looks like node11 tests can't install grpc, so the tests fail.  Everything else seems to pass.

stephenplusplus (approved):

PR #353, opened 2018-11-04
Fix for Issue#343

### LINKED ISSUE #343: Empty column qualifiers should not be rejected
Relevant code is here for reads: https://github.com/googleapis/nodejs-bigtable/blob/13e3d769/src/chunktransformer.js#L251

It's completely valid to use a column like "fam:" which has no qualifier

muraliQlogic:
Fixes #343  

- [ ] Tests and linter pass

### DIFF
--- a/src/chunktransformer.js
+++ b/src/chunktransformer.js
@@ -248,15 +248,6 @@ class ChunkTransformer extends Transform {
         return;
       }
     }
-    if (chunk.familyName && !chunk.qualifier) {
-      this.destroy(
-        new TransformError({
-          message: 'A qualifier must be specified',
-          chunk,
-        })
-      );
-      return;
-    }
     this.validateResetRow(chunk);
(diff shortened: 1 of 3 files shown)


sduskis:
@muraliQlogic: The requirement is to have a check that requires a family with a qualifier.  The qualifier could be the empty string.  This PR doesn't look right.

sduskis:
fam":"" is fine, but "fam":[null] is not ok. I believe the if statement checks for "fam":[null] rather than "fam":"". We should add a unit test to confirm that.  @muraliQlogic, can you please change the the chunktransformer.js test for those requirements?

muraliQlogic:
@sduskis : I modified if condition to check for null or undefined qualifier.

sduskis:
@muraliQlogic: Does `ChunkTransformer. validateNewRow` also need the check for null or undefined?

PR #398, opened 2018-12-26
refactor: modernize the samples a bit

JustinBeckwith:
So this was bad.  None of the tests in the `document_snippet` directory were actually getting run, so 100% of the samples are untested.  The sample tests are also all broken. As a first step, I'm moving files around and updating a few things to get us in a better position to actually get all of the tests passing.


samples/package.json
```diff
@@ -16,17 +20,10 @@
...
-  "nyc": {
-    "exclude": [
-      "**/*.test.js"
-    ]
+    "doctoc": "^1.3.1"
   },
   "scripts": {
     "doctoc": "doctoc README.md */README.md",
-    "test": "mocha system-test/*.test.js --timeout=600000"
+    "test": "mocha system-test --timeout=60000"
```
sduskis: 60 seconds might be problematic.  Table create / modify / delete took 20+ seconds at some points.

PR #403, opened 2019-01-20
fix: Prevent error when `decode: false` option is set in createReadStream.

### LINKED ISSUE #394: validateRowInProgress throws TransformError with decode: false
#### Environment details

  - OS: macOS Sierra 10.12.6
  - Node.js version: v8.14.0
  - npm version: 6.4.1
  - `@google-cloud/bigtable` version: 1.0.0

#### Steps to reproduce

  1. Have a table with multiple cells in the same row
  2. Use `table.createReadStream` with `decode=false` option.
  3. You get a `TransformError: A commit is required between row keys (...)`.

Example:
```js
const options = {
    prefix: 'test',
    filter: [
        {
            column: ['a', 'b']
        },
        {
            column: {
                cellLimit: 1
            }
        }
    ],
    decode: false
}

table.createReadStream(options)
    .on('error', err => {
        console.error(err);
    })
    .on('data', row => {
        console.log(row.id.toString(), row.data)
    })
    .on('end', () => {
        console.log('end');
    });
```

As the key is not being decoded, it will compare two different buffer instances and it will throw at:

https://github.com/googleapis/nodejs-bigtable/blob/ad95d12a5312fe25a1193b9c49ec80fd3e02b0b6/src/chunktransformer.js#L235-L248

zrosenbauer:
Prevent error when `decode: false` option is set in createReadStream.

Fixes #394

- [ ] Tests and linter pass


zrosenbauer:
I signed it!

JustinBeckwith:
@sduskis who's the best person to review bigtable PRs for nodejs?

PR #426, opened 2019-02-27
refactor(ts): convert to typescript

JustinBeckwith:


package.json
```diff
@@ -8,13 +8,11 @@
...
-  "main": "./src/index.js",
+  "main": "build/src/index.js",
   "files": [
-    "protos",
-    "src",
-    "AUTHORS",
-    "CONTRIBUTORS",
-    "LICENSE"
+    "build/protos",
+    "build/proto",
```
callmehiphop: Is this going to be an iterative process like the other repos? We might want to ignore the declaration/map files for now.

package.json
```diff
@@ -29,18 +27,22 @@
...
     "snippet-test": "mocha samples/document-snippets/tests/*.js --timeout 600000",
     "fix": "eslint --fix '**/*.js'",
     "docs-test": "linkinator docs -r --skip www.googleapis.com",
-    "predocs-test": "npm run docs"
+    "predocs-test": "npm run docs",
+    "clean": "gts clean",
+    "compile": "tsc -p . && cp -r src/v2 build/src/ && cp -r proto* build/",
+    "prepare": "npm run compile",
+    "pretest": "npm run compile",
+    "proto": "mkdir -p proto && pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/bigtable/v2/bigtable.proto | pbts -o proto/bigtable.d.ts -"
```
callmehiphop: I think we want to build from `protos` instead of `node_modules/google-proto-files`

src/v2/bigtable_client.js
```diff
@@ -19,7 +19,7 @@ const gax = require('google-gax');
 const merge = require('lodash.merge');
 const path = require('path');
 
-const VERSION = require('../../package.json').version;
+const VERSION = require('../../../package.json').version;
```
callmehiphop: Will this get overwritten by synthtool?

system-test/read-rows.ts
```diff
@@ -88,9 +101,9 @@ describe('Bigtable/Table', () => {
       responses = null;
       rowKeysRead = [];
       requestedOptions = [];
-      stub = sinon.stub(bigtable, 'request').callsFake(cfg => {
-        const reqOpts = cfg.reqOpts;
-        const requestOptions = {};
+      stub = (sinon as any).stub(bigtable, 'request').callsFake(cfg => {
```
callmehiphop: Should install `@types/sinon` for this and use `sinon.SinonStub`

package.json
JustinBeckwith: Yes, and good call.  Updated!

system-test/read-rows.ts
JustinBeckwith: I did!  There is some weird typing bug I didn't want to try digging into here, and figured I would pick it up later.

src/v2/bigtable_client.js
JustinBeckwith: woooops good find

PR #439, opened 2019-03-14
refactor: drop dependency on extend

JustinBeckwith:


src/index.ts
```diff
@@ -683,7 +682,7 @@ export class Bigtable {
           this.api[config.client] = gaxClient;
         }
 
-        let reqOpts = extend(true, {}, config.reqOpts);
```
callmehiphop: I'm not sure we want to make this change, `extend(true, ...)` creates a deep copy of an object, and `Object.assign` makes a shallow copy.

JustinBeckwith: I will admit to not fully understanding the difference here.  I'm mostly putting it out there as a conversation starter :) 

When you say "deep" copy, what I thought that meant is that it wouldn't copy past the first set of properties.  Yet...

```
> Object.assign({}, {a: { b: { c: 5 } } });
{ a: { b: { c: 5 } } }
```

Can you show me how this breaks down different than extend?

callmehiphop: I guess when I say deep, I mean recursive. To illustrate

```js
const a = {yes: true};
const b = {a};
const c = Object.assign({}, b);

b.a.yes = false;
console.log(c.a.yes);
// => false
```

JustinBeckwith: Got it. Thank you so much for illustrating!

callmehiphop: My pleasure!

bcoe: a trick I've used in yargs, to avoid introducing a dependency, is the following:

`const copy = JSON.parse(JSON.stringify(obj)) `, if you know your object is going to only have simple types.

PR #440, opened 2019-03-14
refactor(ts): add proto generated files

### LINKED ISSUE #427: Convert library to TypeScript

joecervino:
Fixes #427 
- [x] Tests and linter pass

Just adding the proto generated files & script before making PR's for the rest of the files 👍


package.json
```diff
@@ -44,7 +45,10 @@
     "compile": "tsc -p . && cp -r src/v2 build/src/ && cp -r proto* build/ && cp test/*.js build/test",
     "prepare": "npm run compile",
     "pretest": "npm run compile",
-    "proto": "mkdir -p proto && pbjs -t static-module -w commonjs -p protos google/bigtable/v2/bigtable.proto | pbts -o proto/bigtable.d.ts -"
+    "proto": "npm run proto:bigtable && npm run proto:bigtable_instance_admin && npm run proto:bigtable_table_admin",
+    "proto:bigtable": "mkdir -p proto && pbjs -t static-module -w commonjs -p protos google/bigtable/v2/bigtable.proto | pbts -i long -o proto/bigtable.d.ts -",
+    "proto:bigtable_instance_admin": "mkdir -p proto && pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/bigtable/admin/v2/bigtable_instance_admin.proto | pbts -i long -o proto/bigtable_instance_admin.d.ts -",
```
callmehiphop: Any reason why we are building from node_modules instead of the protos directory?

proto/bigtable.d.ts
```diff
@@ -1,4 +1,5 @@
 import * as $protobuf from "protobufjs";
+import * as long from "long";
```
callmehiphop: This doesn't work, we'll need to remove the `-i` parameter and unfortunately hardcode the correct import for now.

```ts
import * as Long from 'long';
```

proto/bigtable_instance_admin.d.ts
```diff
@@ -0,0 +1,8646 @@
+import * as $protobuf from "protobufjs";
+import * as long from "long";
```
callmehiphop: Same as above

proto/bigtable_table_admin.d.ts
```diff
@@ -0,0 +1,7759 @@
+import * as $protobuf from "protobufjs";
+import * as long from "long";
```
callmehiphop: Same as above.

proto/bigtable.d.ts
joecervino: Yikes okay :(

package.json
joecervino: Hm, yeah good question and no excuse for you to be honest. I was keeping the script consistent across Datastore and Spanner and didn't realize we were sourcing the latter two files differently. Lesson learned - don't copy, paste, forget 😅 Fixed!

package.json
```diff
@@ -44,7 +45,7 @@
     "compile": "tsc -p . && cp -r src/v2 build/src/ && cp -r proto* build/ && cp test/*.js build/test",
     "prepare": "npm run compile",
     "pretest": "npm run compile",
-    "proto": "mkdir -p proto && pbjs -t static-module -w commonjs -p protos google/bigtable/v2/bigtable.proto | pbts -o proto/bigtable.d.ts -"
+    "proto": "mkdir -p proto && pbjs -t static-module -w commonjs -p protos google/bigtable/v2/bigtable.proto | pbts -o proto/bigtable.d.ts - && pbjs -t static-module -w commonjs -p protos google/bigtable/admin/v2/bigtable_instance_admin.proto | pbts -o proto/bigtable_instance_admin.d.ts - && pbjs -t static-module -w commonjs -p protos google/bigtable/admin/v2/bigtable_table_admin.proto | pbts -o proto/bigtable_table_admin.d.ts -"
```
callmehiphop: This is a very big command! Could we either

1. Bundle the types together into a single file (this is untested but I think it works)
    ```sh
    $ mkdir -p proto && pbjs -t static-module -w commonjs -p protos google/bigtable/v2/bigtable.proto google/bigtable/admin/v2/bigtable_instance_admin.proto google/bigtable/admin/v2/bigtable_table_admin.proto | pbts -o proto/bigtable.d.ts -
    ```
1. Have separate commands per file

While we're at it lets install `mkdirp` as a dev dependency and use that instead of `mkdir -p`.

joecervino: I thought that might've been what you meant, but I guess I took the safe route - it works! Pushing now 👍

PR #443, opened 2019-03-20
refactor(ts): enable noImplicitAny on index.ts

### LINKED ISSUE #427: Convert library to TypeScript

joecervino:
Fixes #427 (1/21 files) :)

- [x] Tests and linter pass


src/index.ts
```diff
@@ -38,16 +38,19 @@
 
 import * as arrify from 'arrify';
 import {replaceProjectIdToken} from '@google-cloud/projectify';
+import {Operation as GaxOperation} from 'google-gax/build/src/longrunning';
```
JustinBeckwith: Do you need to reach into the full path here, or is this type exported from the index of `google-gax`?

src/index.ts
```diff
@@ -86,6 +89,94 @@ const {grpc} = new gax.GrpcClient();
  * @property {Constructor} [promise] Custom promise module to use instead of
  *     native Promises.
  */
+export interface Client {
+  //tslint: disable-next-line: no-any
+  [k: string]: any;
```
JustinBeckwith: I mean, this pretty much make this object an `any`

src/index.ts
```diff
@@ -86,6 +89,94 @@ const {grpc} = new gax.GrpcClient();
...
+  servicePath?: string;
+  sslCreds?: ChannelCredentials;
+  port?: number;
+}
+export interface ClientConfig {
+  [k: string]: string|{}|Function|Client|undefined;
```
JustinBeckwith: I would start this one over, and extend `GoogleAuthOptions` as it will have most oof these fields

src/index.ts
```diff
@@ -86,6 +89,94 @@ const {grpc} = new gax.GrpcClient();
...
+  gaxOptions?: CallOptions;
+}
+export interface CreateInstanceRequest {
+  parent?: string|null;
+  instanceId?: string|null;
+  clusters?: ({[k: string]: ICluster}|null);
+  instance?: IInstance|null;
+}
+//tslint:disable-next-line interface-name
+export interface IInstance {
```
JustinBeckwith: I don't really like the `I` prefix for interfaces, and it isn't something we've used elsewhere.  Is there a proper Instance type somewhere else in the project?

src/index.ts
```diff
@@ -86,6 +89,94 @@ const {grpc} = new gax.GrpcClient();
...
+  name?: string|null;
+  displayName?: string|null;
+  state?: InstanceState|null;
+  type?: InstanceType|null;
+  labels?: {[k: string]: string}|null|string;
+}
+export interface RequestConfig {
```
JustinBeckwith: @callmehiphop I feel like this interface has to be defined somewhere else already

src/index.ts
callmehiphop: This is my fault I think, lol. For exporting generated response types I was at a loss for what to call them.

src/index.ts
callmehiphop: There is the [`ProtoOpts`](https://github.com/googleapis/nodejs-common-grpc/blob/master/src/service.ts#L38) interface in `grpc-common`, but we'll need to extend it (I think). For gax/handwritten clients, the request method is not inherited from anything, so that's probably why it isn't defined in any core libs.

joecervino:
@JustinBeckwith @callmehiphop Sounds good - I'll get `npm run fix` working properly,  I'll refine `index.ts`, and I'll submit a PR for something a bit more foundational. Thanks for taking the time to double-team this :)

src/index.ts
joecervino: I hear ya - I extended `ProtoOpts` in OPR 1.2 but had some trouble with it so I added `ProtoOpts` properties into `RequestConfig`. I can slim it down if needed 👍

src/index.ts
joecervino: I renamed it `ManualIInstance` if that's okay for now 🤷‍♂️

src/index.ts
joecervino: 🤦‍♂️ Yeah, the precise index signature would've been even more offensive. I removed it for now to see if later PR's can help solve the issue (it's born strictly out of enabling noImplicitAny)

JustinBeckwith:
This has sat dormont for a while, and I think its the last thing that should land.  I'm closing this PR out for now, but keeping the branch around.

PR #448, opened 2019-03-25
refactor(ts): enable noImplicitAny on mutation.ts

### LINKED ISSUE #427: Convert library to TypeScript

joecervino:
Fixes #427 (2/21ish files)

Notes:
1. I branched this off of the `run-fix` branch to make sure I'm linting properly. Hope that's cool.
2. This was a tricky one, no online docs and little in-repo docs, so looking forward to learning :)

- [x] Tests and linter pass

### DIFF
--- a/src/mutation.ts
+++ b/src/mutation.ts
@@ -17,7 +17,6 @@
 import * as arrify from 'arrify';
 import * as is from 'is';
 import * as Long from 'long';
-<<<<<<< HEAD
 
 import {google as btTypes} from '../proto/bigtable';
 
@@ -62,8 +61,6 @@ export interface SetCellObj {
 export interface ValueObj {
   [k: string]: Buffer|Value|ValueObj;
 }
-=======
->>>>>>> ecb47fc0233bcb2068509696c1f4a3fccb9dd222
 
 /**
  * Formats table mutations to be in the expected proto format.
@@ -276,14 +273,9 @@ export class Mutation {
    * //   deleteFromRow: {}
    * // }
    *
-   * //-
    * // It's also possible to specify a time range when deleting specific
-<<<<<<< HEAD
    * // columns.
-=======
-   * columns.
->>>>>>> ecb47fc0233bcb2068509696c1f4a3fccb9dd222
-   * //-
+   *
    * Mutation.encodeDelete([
    *   {
    *     column: 'follows:gwashington',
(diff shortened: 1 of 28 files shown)


joecervino:
Boo docs, boo 👎

src/mutation.ts
```diff
@@ -14,10 +14,49 @@
  * limitations under the License.
  */
 
+import {Metadata} from '@google-cloud/common';
```
callmehiphop: Pretty sure we should never use this type in grpc clients.

src/mutation.ts
```diff
@@ -242,7 +285,8 @@ export class Mutation {
    * ]);
    * @private
    */
-  static encodeDelete(data?) {
+  static encodeDelete(data?: Metadata): EncodeDeleteDataResponse
```
callmehiphop: Usually `encode` methods are just formatting request objects. I think you could use [`IDeleteFromColumn`](https://github.com/googleapis/nodejs-bigtable/blob/master/proto/bigtable.d.ts#L3336) here, but moving forward I think we should avoid using `Response` for these kinds of methods.

src/mutation.ts
```diff
@@ -154,8 +196,9 @@ export class Mutation {
    * // ]
    * @private
    */
-  static encodeSetCell(data) {
-    const mutations: any[] = [];
+  static encodeSetCell(data: Metadata): EncodeSetCellResponse[] {
```
callmehiphop: Can we use [`ISetCell`](https://github.com/googleapis/nodejs-bigtable/blob/master/proto/bigtable.d.ts#L3228) here?

src/mutation.ts
```diff
@@ -292,7 +336,7 @@ export class Mutation {
    * @returns {object}
    * @private
    */
-  static parse(mutation) {
+  static parse(mutation: Mutation): JsonObj {
```
callmehiphop: Should this be [`IMutateRowRequest`](https://github.com/googleapis/nodejs-bigtable/blob/master/proto/bigtable.d.ts#L709) ?

src/mutation.ts
```diff
@@ -314,7 +358,7 @@ export class Mutation {
    * // }
    * @private
    */
-  static parseColumnName(column) {
+  static parseColumnName(column: string): ParseColumnNameResponse {
```
callmehiphop: Can we rename this to something like `Column`?

src/mutation.ts
```diff
@@ -329,7 +373,8 @@ export class Mutation {
    * @returns {object}
    * @private
    */
-  toProto() {
+  toProto(): JsonObj {
+    // tslint:disable-next-line no-any
```
callmehiphop: I think this should be [`IMutateRowRequest`](https://github.com/googleapis/nodejs-bigtable/blob/master/proto/bigtable.d.ts#L709), should be able to drop the `no-any` comment too.

src/mutation.ts
joecervino: Just curious, want to learn this right - is there something specific about `static` methods that mean I should avoid a `Response` return annotation or is it something more general about `Response`s?

callmehiphop: More general for `Response`, IMO we should only use that word to describe actual response payloads, where this method I think is just something we use internally to transform an object into the expected proto format.

joecervino: Makes sense, cool. Noted 👍

JustinBeckwith:
@callmehiphop can this get another look?

PR #492, opened 2019-05-31
docs(samples): add conditional, batch, and incremental write samples

billyjacobson:
- [ ] Tests and linter pass


billyjacobson:
I think it might be better to reuse the instance through the test rather than creating and deleting. Might encounter race conditions, or instances that don't delete and then we'd be hogging bigtable resources.

PR #501, opened 2019-06-21
refactor: drop dependency on nodejs-repo-tools

JustinBeckwith:


crwilcox:
Samples and systems are also failing on the release pr. I think quota is low.

crwilcox:
Samples test failure seems real?
```
1) instances
       should list zones:
     Error: Command failed: node instances.js "nodejs-bigtable-samples-369c21"
instances.js <command>
```

PR #507, opened 2019-07-10
fix: cancel streaming grpc request when user ends stream

### LINKED ISSUE #506: Cancelling a read row stream doesn't cancel the underlying grpc stream
This is copied from a stackoverflow question:

https://stackoverflow.com/questions/56918799/how-to-manually-end-stream-created-by-tablecreatereadstream

> I want to process data from BigTable table in a stream, but end stream early. I am using this code for Node.js (slightly modified https://github.com/googleapis/nodejs-bigtable/blob/master/samples/document-snippets/table.js#L215):

```javascript
const stream = new BigTable({projectId, keyFilename})
  .instance(instanceId)
  .table(tableName)
  .createReadStream();
stream.on('data', () => {
  stream.end();
});
stream.on('end', () => {
  console.log('end');
  stream.destroy();
});
stream.on('close', () => {
  console.log('close');
});
```

Output is:
```javascript
end
close
```

But the process is not yet finished. So, i believe that some resources are not closed. How should i stop stream and close relevant resources?

stephenplusplus:
Fixes #506

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "through2": "^3.0.0"
   },
   "devDependencies": {
+    "@google-cloud/common": "^2.0.4",
     "@types/escape-string-regexp": "^1.0.0",
     "@types/extend": "^3.0.1",
     "@types/is": "0.0.21",
--- a/package.json
+++ b/package.json
@@ -69,7 +69,6 @@
     "through2": "^3.0.0"
   },
   "devDependencies": {
-    "@google-cloud/common": "^2.0.4",
     "@types/escape-string-regexp": "^1.0.0",
     "@types/extend": "^3.0.1",
     "@types/is": "0.0.21",
--- a/src/table.ts
+++ b/src/table.ts
@@ -32,6 +32,8 @@ import {ChunkTransformer} from './chunktransformer';
 // See protos/google/rpc/code.proto
 // (4=DEADLINE_EXCEEDED, 10=ABORTED, 14=UNAVAILABLE)
 const RETRYABLE_STATUS_CODES = new Set([4, 10, 14]);
+// (1=CANCELLED)
+const IGNORED_STATUS_CODES = new Set([1]);
 
 /**
  * Create a Table object to interact with a Cloud Bigtable table.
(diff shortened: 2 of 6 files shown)


src/table.ts
```diff
@@ -321,6 +325,14 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
     }
 
     const userStream = through.obj();
+    const end = userStream.end.bind(userStream);
+    userStream.end = () => {
```
stephenplusplus: @callmehiphop what do you think about overriding `end()` here, and cancelling the request stream? (`abort()` maps to the grpc stream's `cancel()`). Is there a better way?

Note that we show in the sample documentation that we will end a stream early when the user calls `this.end()` from a data event handler.

system-test/read-rows-acceptance-tests.ts
```diff
@@ -87,6 +87,9 @@ describe('Read Row Acceptance tests', function() {
           objectMode: true,
         });
 
+        /* tslint:disable-next-line */
+        (stream as any).abort = function() {};
```
JustinBeckwith: I believe nodejs-common has an `AbortableDuplex` interface that should work here.

src/table.ts
callmehiphop: Makes sense to me.

I feel like we should have a discussion about making a breaking change to move away from `end()`. IMO its a little confusing because `end()` is a writable method that indicates no more data should be written to the stream. I think if we went all in on destroy instead we could just implement a `_destroy` method that would cancel the request when called. That or something custom like `cancel` or `abort`. 🤷‍♂

stephenplusplus: Yep, abort or cancel are ideal names to me. Although taking out the naming collision, "end()" makes sense, too. I guess I will keep end for now and open an issue about a switch in the future?

callmehiphop: SGTM!

system-test/read-rows-acceptance-tests.ts
stephenplusplus: @JustinBeckwith could you take a look if I went about this the right way? I had to add a devDependency on common, and cast the PassThrough streams to `Duplex` streams. It started to feel wrong :)

package.json
```diff
@@ -69,6 +69,7 @@
     "through2": "^3.0.0"
   },
   "devDependencies": {
+    "@google-cloud/common": "^2.0.4",
```
JustinBeckwith: Ugh.  This will cause problems.  Back the way it was!  The `any` will do fine.

callmehiphop:
LGTM if @JustinBeckwith is happy with it!

PR #509, opened 2019-07-17
fix: do not log warning for ignored grpc options

### LINKED ISSUE #508: Calling any methods logs warning on the console in version 2.0.1
#### Environment details

  - OS: macOS
  - Node.js version: v10.11.0
  - npm version: 6.5.0
  - `@google-cloud/bigtable` version: 2.0.1

#### Steps to reproduce

  1. Instantiate BigTable
  2. Call any method that uses grpc underneath
  3. Logs warning to the console:

```
Unrecognized channel argument 'max_send_message_length' will be ignored.
Unrecognized channel argument 'max_receive_message_length' will be ignored.
```

Looks like it caused by the `'grpc.max_send_message_length': -1` and       `'grpc.max_receive_message_length': -1` default configs in [src/index.ts](https://github.com/googleapis/nodejs-bigtable/blob/061a28b86638231d724caef456876bd79c02aacb/src/index.ts#L405)

mkls:
Fixes #508

Not sure this is the correct fix for the problem, but gets rid of the warnings

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -401,9 +401,7 @@ export class Bigtable {
       {
         libName: 'gccl',
         libVersion: PKG.version,
-        scopes,
-        'grpc.max_send_message_length': -1,
-        'grpc.max_receive_message_length': -1,
+        scopes
       },
       options
     );


JustinBeckwith:
@stephenplusplus mind taking a look?

mkls:
This was actually recently fixed in #528, will close this PR

PR #527, opened 2019-08-02
fix(types): Include TypeScript types with module

### LINKED ISSUE #526: Emit typescript types on build
**Is your feature request related to a problem? Please describe.**
While using this library there is absolutely 0 typing information available on import. There is a constant need to have the docs open or div into the source code itself to be able to use this library. 

Most of the code is written in typescript why wouldn't types be emitted?

 **Describe the solution you'd like**
Typescript definition files should be emitted on build.
 **Describe alternatives you've considered**
I suppose I could generate my own and commit them to DT but I feel that wouldn't be the right approach.

csidell-earny:
Fixes #526

- [ X ] Tests and linter pass
- [ X ] Code coverage does not decrease (if any source code was changed)
- [ X ] Appropriate docs were updated (if necessary)

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "!**/*.map"
   ],
   "main": "build/src/index.js",
+  "types": "build/src/index.d.ts",
   "scripts": {
     "check": "gts check",
     "clean": "gts clean",
--- a/protos/protos.json
+++ b/protos/protos.json
@@ -1,0 +1,2681 @@
+{
+  "nested": {
+    "google": {
+      "nested": {
+        "bigtable": {
+          "nested": {
+            "admin": {
+              "nested": {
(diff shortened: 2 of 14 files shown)


csidell-earny:
@googlebot I signed it!

csidell-earny:
@JustinBeckwith I wouldn't know what it is upstream, I assume it's the same filename?

JustinBeckwith:
It should look exactly like the line I pointed to above :)

csidell-earny:
Well in that case that change has been made!

csidell-earny:
Should be ready to merge, undid the type fixes for some of the exists() calls as to not have this do more than the original PR title. @JustinBeckwith

PR #554, opened 2019-10-11
refactor: drop dependency on @google-cloud/common-grpc

JustinBeckwith:
This is more a conversation starter than anything.  The `@google-cloud/common-grpc` module creates a few problems for us:
- `google-gax`, `@google-cloud/common`, and `@google-cloud/common-grpc` all reference `google-auth-library`, which is causing diamond dependency issues when we update things
- We are looking to make broad updates to the transport layers, and I'd like to reduce the number of common core libs we try to use

There is some code copying in here in `decorateStatus` I don't feel great about.  Hoping @alexander-fenster will have suggestions on if/how/where in gax this could live.

With minimal effort, we could also drop the dependency on `@google-cloud/common`.  The error util libraries seem to be the only bits we're using.  We should have a single set of Error classes we're using, and have those potentially available in their own shared single use library.

### DIFF
--- a/src/decorateStatus.ts
+++ b/src/decorateStatus.ts
@@ -53,7 +53,6 @@ export function decorateStatus(response) {
   if (response && GRPC_ERROR_CODE_TO_HTTP[response.code]) {
     const defaultResponseDetails = GRPC_ERROR_CODE_TO_HTTP[response.code];
     let message = defaultResponseDetails.message;
-
     if (response.message) {
       // gRPC error messages can be either stringified JSON or strings.
       try {
@@ -62,12 +61,14 @@ export function decorateStatus(response) {
         message = response.message;
       }
     }
-
     return extend(true, obj, response, {
       code: defaultResponseDetails.code,
       message,
     });
   }
-
   return null;
+}
+
+export function shouldRetryRequest(r: {code: number}) {
+  return [429, 500, 502, 503].includes(r.code);
 }
--- a/src/decorateStatus.ts
+++ b/src/decorateStatus.ts
@@ -1,0 +1,73 @@
+/*!
+ * Copyright 2019 Google LLC. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as extend from 'extend';
+
+/**
+ * @const {object} - A map of protobuf codes to HTTP status codes.
+ * @private
+ */
+const GRPC_ERROR_CODE_TO_HTTP = [
+  {code: 200, message: 'OK'},
+  {code: 499, message: 'Client Closed Request'},
+  {code: 500, message: 'Internal Server Error'},
+  {code: 400, message: 'Bad Request'},
+  {code: 504, message: 'Gateway Timeout'},
+  {code: 404, message: 'Not Found'},
+  {code: 409, message: 'Conflict'},
+  {code: 403, message: 'Forbidden'},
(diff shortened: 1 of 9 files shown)


src/decorateStatus.ts
```diff
@@ -0,0 +1,73 @@
...
+ * limitations under the License.
+ */
+
+import * as extend from 'extend';
+
+/**
+ * @const {object} - A map of protobuf codes to HTTP status codes.
+ * @private
+ */
+const GRPC_ERROR_CODE_TO_HTTP = [
```
alexander-fenster: I hate this array so much. It's a plain mapping from gRPC error codes to HTTP error codes which does not make much sense. `@grpc/grpc-js` has its own mapping from error codes to text descriptions:

```
> require('@grpc/grpc-js').status
{
  '0': 'OK',
  '1': 'CANCELLED',
  '2': 'UNKNOWN',
  '3': 'INVALID_ARGUMENT',
  '4': 'DEADLINE_EXCEEDED',
  '5': 'NOT_FOUND',
  '6': 'ALREADY_EXISTS',
  '7': 'PERMISSION_DENIED',
  '8': 'RESOURCE_EXHAUSTED',
  '9': 'FAILED_PRECONDITION',
  '10': 'ABORTED',
  '11': 'OUT_OF_RANGE',
  '12': 'UNIMPLEMENTED',
  '13': 'INTERNAL',
  '14': 'UNAVAILABLE',
  '15': 'DATA_LOSS',
  '16': 'UNAUTHENTICATED',
  OK: 0,
  CANCELLED: 1,
  UNKNOWN: 2,
  INVALID_ARGUMENT: 3,
  DEADLINE_EXCEEDED: 4,
  NOT_FOUND: 5,
  ALREADY_EXISTS: 6,
  PERMISSION_DENIED: 7,
  RESOURCE_EXHAUSTED: 8,
  FAILED_PRECONDITION: 9,
  ABORTED: 10,
  OUT_OF_RANGE: 11,
  UNIMPLEMENTED: 12,
  INTERNAL: 13,
  UNAVAILABLE: 14,
  DATA_LOSS: 15,
  UNAUTHENTICATED: 16
}
```

and I feel it's much better to just provide this explanation rather than try to map it to HTTP code (e.g. `15 DATA_LOSS` does not really translate to `500 Internal Server Error`, and `11 OUT_OF_RANGE` cannot be replaced with `400 Bad Request` since it loses important details.

So, let's get rid of this mapping. If it makes it to be a semver major change, I'm totally fine with it.

src/index.ts
```diff
@@ -749,7 +747,7 @@ export class Bigtable {
             currentRetryAttempt: 0,
             noResponseRetries: 0,
             objectMode: true,
-            shouldRetryFn: (Service as any).shouldRetryRequest_,
+            shouldRetryFn: r => [429, 500, 502, 503].indexOf(r.code) > -1,
```
alexander-fenster: We should refactor this to use gax for handling retries, it's perfectly capable of doing it.  (I didn't read this code so it might be harder than it sounds)

JustinBeckwith: Hrumph.  The comments in the code point to here, which makes me think maybe we don't support retries natively here yet?
https://github.com/googleapis/gax-nodejs/blob/master/src/streamingCalls/streamDescriptor.ts#L55

src/decorateStatus.ts
JustinBeckwith: While I agree we should change this, I kinda don't want to touch too much here at once.  Happy to create a cleanup item to swing back around.

src/decorateStatus.ts
```diff
@@ -0,0 +1,73 @@
...
+      // gRPC error messages can be either stringified JSON or strings.
+      try {
+        message = JSON.parse(response.message).description;
+      } catch (e) {
+        message = response.message;
+      }
+    }
+
+    return extend(true, obj, response, {
+      code: defaultResponseDetails.code,
```
bcoe: do we have tests exercising the `decorateStatus` logic, even one smoke test that exercised the lookup table would be great.

src/index.ts
bcoe: likewise, I care less about this logic being pulled in from `grpc-common`, and care more that we make sure there's one test that exercise the retry logic.

src/decorateStatus.ts
JustinBeckwith: Tests added!  The existing ones were less than comprehensive, so I wrote new ones.

JustinBeckwith: added!

PR #576, opened 2019-11-13
fix: safe integer bounds conversion

### LINKED ISSUE #575: One-off issue in conversion of INT64 to number
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

  - Search the issues already opened: https://github.com/googleapis/nodejs-bigtable/issues
  - Search the issues on our "catch-all" repository: https://github.com/googleapis/google-cloud-node
  - Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js

#### Environment details

  - OS: MacOS 10.4, Windows 10
  - Node.js version: !2.3.1
  - npm version: 6.10.2
  - `@google-cloud/bigtable` version: 1.0.1

#### Steps to reproduce

  1. Set an Integer column to Number.MIN_SAFE_INTEGER  or Number.MAX_SAFE_INTEGER, via some other method
  2. Attempt to read the value. Using this library
  3. See that the value returned is a string.

Source: https://github.com/googleapis/nodejs-bigtable/blob/1cb65a47553ab93572bfd95825074ab362ecd9f8/src/mutation.ts#L114
This line should read 
```
if (Number.MIN_SAFE_INTEGER <= num && num <= Number.MAX_SAFE_INTEGER) {
```
to include the SAFE_INTEGER bounds.

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

laljikanjareeya:
Instead of checking `MIN_SAFE_INTEGER` and `MAX_SAFE_INTEGER` condition, Added embedded function `Number.isSafeInteger`  to check safe integer.

Fixes #575

### DIFF
--- a/src/mutation.ts
+++ b/src/mutation.ts
@@ -111,7 +111,7 @@ export class Mutation {
       // tslint:disable-next-line no-any
       const num = Long.fromBytes(buf as any).toNumber();
 
-      if (Number.MIN_SAFE_INTEGER < num && num < Number.MAX_SAFE_INTEGER) {
+      if (Number.isSafeInteger(num)) {
         return num;
       }
     }
--- a/test/mutation.ts
+++ b/test/mutation.ts
@@ -81,6 +81,30 @@ describe('Bigtable/Mutation', function() {
         assert.strictEqual(num, decoded);
       });
 
+      it('should not convert a base64 encoded smaller than MIN_SAFE_INTEGER number when true', function() {
+        const num = Number.MIN_SAFE_INTEGER - 100;
+        const encoded = Buffer.from(Long.fromNumber(num).toBytesBE()).toString(
+          'base64'
+        );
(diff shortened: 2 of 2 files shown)


AVaksman:
@laljikanjareeya could you please also add a unit test(s) where `{isPossibleNumber: true}` true is passed but value to be decoded is outside of bounds of `Number` (smaller than min, larger then max).

laljikanjareeya:
> @laljikanjareeya could you please also add a unit test(s) where `{isPossibleNumber: true}` is passed, but value to be decoded is outside of bounds of `Number` (smaller than min, larger then max).

@AVaksman Added more unit tests for outside of bounds.

AVaksman:
LGTM

PR #579, opened 2019-11-14
fix(types): overload for bigtable.getInstances

### LINKED ISSUE #564: getInstances type is incorrect: returns a Promise when callback is unspecified
#### Environment details

  - OS: macOS 10.14
  - Node.js version: v12.12.0
  - npm version: 6.11.3
  - `@google-cloud/bigtable` version: 2.2.1

#### Steps to reproduce

I have code that calls `getInstances` without defining a callback:

```ts
const [instances] = await bigtable.getInstances();
```

`getInstances` behaves according to its docstring:

> If the callback is omitted, we'll return a Promise.

...but this behavior is not reflected in its generated type, which has return type `void` and no overloads:

```ts
getInstances(gaxOptions?: any, callback?: any): void;
```

This mismatch causes eslint to complain because `void` is a non-Promise.

Screen Shot 2019-11-07 at 15 44 55

#### Proposal

Declare an overload type:

```typescript
getInstances(gaxOptions?: any, callback: any): void;
getInstances(gaxOptions?: any): Promise<Instance[]>;
```

Thanks!

laljikanjareeya:
While accessing `bigtable.getInstances` with `async` function, it is giving error of return type.
Added overload and related document changes.

Fixes #564

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -73,9 +73,9 @@ export interface GetInstancesCallback {
   ): void;
 }
 export type GetInstancesResponse = [
-  Instance[] | null,
+  Instance[],
   {} | null,
-  google.bigtable.admin.v2.IListInstancesResponse | null
+  google.bigtable.admin.v2.IListInstancesResponse
 ];
 /**
  * @typedef {object} ClientConfig
@@ -603,21 +603,9 @@ export class Bigtable {
     );
   }
 
+  getInstances(gaxOptions?: CallOptions): Promise<GetInstancesResponse>;
   getInstances(callback: GetInstancesCallback): void;
   getInstances(gaxOptions: CallOptions, callback: GetInstancesCallback): void;
-  getInstances(gaxOptions?: CallOptions): Promise<GetInstancesResponse>;
-  /**
-   * Query object for listing instances.
-   *
-   * @typedef {object} CallOptions
-   * @property {boolean} [autoPaginate=true] Have pagination handled
-   *     automatically.
-   * @property {number} [maxApiCalls] Maximum number of API calls to make.
-   * @property {number} [maxResults] Maximum number of items to return.
-   * @property {number} [pageSize] Maximum number of results per page.
-   * @property {string} [pageToken] A previously-returned page token
-   *     representing part of the larger set of results to view.
-   */
   /**
    * @typedef {array} GetInstancesResponse
    * @property {Instance[]} 0 Array of {@link Instance} instances.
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -62,6 +64,19 @@ const PKG = require('../../package.json');
...
+export interface GetInstancesCallback {
+  (
+    err: ServiceError | null,
+    result?: Instance[] | null,
+    nextQuery?: {} | null,
+    response?: google.bigtable.admin.v2.IListInstancesResponse | null
+  ): void;
+}
+export type GetInstancesResponse = [
+  Instance[] | null,
```
callmehiphop: I don't think this will ever be null, if the request fails the promise would get rejected.

src/index.ts
```diff
@@ -62,6 +64,19 @@ const PKG = require('../../package.json');
...
+  {} | null,
+  google.bigtable.admin.v2.IListInstancesResponse | null
```
callmehiphop: same as above!

src/index.ts
```diff
@@ -588,16 +603,39 @@ export class Bigtable {
     );
   }
 
+  getInstances(callback: GetInstancesCallback): void;
+  getInstances(gaxOptions: CallOptions, callback: GetInstancesCallback): void;
+  getInstances(gaxOptions?: CallOptions): Promise<GetInstancesResponse>;
```
callmehiphop: This has always been kinda fuzzy for me - is there any harm in declaring this AFTER the callback overloads? I know the official docs say to order them from most specific to least, but I don't really know if it matters here.

src/index.ts
```diff
@@ -588,16 +603,39 @@ export class Bigtable {
...
+  /**
+   * Query object for listing instances.
+   *
+   * @typedef {object} CallOptions
```
callmehiphop: I would advocate that we do not include a JSDoc typedef for this, just because if this is likely to become stale and overlooked. If anything lets use either an [`@external` tag](https://jsdoc.app/tags-external.html) or just link to the upstream docs.

src/index.ts
```diff
@@ -610,17 +648,37 @@ export class Bigtable {
    * });
    *
    * //-
+   * // To control how many API requests are made and page through the results
+   * // manually, set `autoPaginate` to `false`.
+   * //-
```
callmehiphop: This is nothing you did, but this comment style was used for a really old version of our docs. Since we're touching it we might as well update it to use `@example <caption>...</caption>`

PR #586, opened 2019-11-26
feat(samples): adds read and filter snippets

billyjacobson:

### DIFF
--- a/samples/__snapshots__/filters.js
+++ b/samples/__snapshots__/filters.js
@@ -1,0 +1,501 @@
+exports['filters should filter with row regex 1'] = `
+Reading data for phone#4c410523#20190501:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+\tos_build: PQ2A.190405.003 @1559361600000000
+Column Family cell_plan
+\tdata_plan_01gb: true @1559361600000000
+\tdata_plan_01gb: false @1559188800000000
+\tdata_plan_05gb: true @1559361600000000
+
+Reading data for phone#5c10102#20190501:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+\tos_build: PQ2A.190401.002 @1559361600000000
+Column Family cell_plan
+\tdata_plan_10gb: true @1559361600000000
+
+
+`;
+
+exports['filters should filter with cells per col 1'] = `
+Reading data for phone#4c410523#20190501:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+\tos_build: PQ2A.190405.003 @1559361600000000
+Column Family cell_plan
+\tdata_plan_01gb: true @1559361600000000
+\tdata_plan_01gb: false @1559188800000000
+\tdata_plan_05gb: true @1559361600000000
+
+Reading data for phone#4c410523#20190502:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+\tos_build: PQ2A.190405.004 @1559361600000000
+Column Family cell_plan
+\tdata_plan_05gb: true @1559361600000000
+
+Reading data for phone#4c410523#20190505:
+Column Family stats_summary
+\tconnected_cell: 0 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+\tos_build: PQ2A.190406.000 @1559361600000000
+Column Family cell_plan
+\tdata_plan_05gb: true @1559361600000000
+
+Reading data for phone#5c10102#20190501:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+\tos_build: PQ2A.190401.002 @1559361600000000
+Column Family cell_plan
+\tdata_plan_10gb: true @1559361600000000
+
+Reading data for phone#5c10102#20190502:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 0 @1559361600000000
+\tos_build: PQ2A.190406.000 @1559361600000000
+Column Family cell_plan
+\tdata_plan_10gb: true @1559361600000000
+
+
+`;
+
+exports['filters should filter with cells per row 1'] = `
+Reading data for phone#4c410523#20190501:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+
+Reading data for phone#4c410523#20190502:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+
+Reading data for phone#4c410523#20190505:
+Column Family stats_summary
+\tconnected_cell: 0 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+
+Reading data for phone#5c10102#20190501:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+
+Reading data for phone#5c10102#20190502:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 0 @1559361600000000
+
+
+`;
+
+exports['filters should filter with cells per row offset 1'] = `
+Reading data for phone#4c410523#20190501:
+Column Family stats_summary
+\tos_build: PQ2A.190405.003 @1559361600000000
+Column Family cell_plan
+\tdata_plan_01gb: true @1559361600000000
+\tdata_plan_01gb: false @1559188800000000
+\tdata_plan_05gb: true @1559361600000000
+
+Reading data for phone#4c410523#20190502:
+Column Family stats_summary
+\tos_build: PQ2A.190405.004 @1559361600000000
+Column Family cell_plan
+\tdata_plan_05gb: true @1559361600000000
+
+Reading data for phone#4c410523#20190505:
+Column Family stats_summary
+\tos_build: PQ2A.190406.000 @1559361600000000
+Column Family cell_plan
+\tdata_plan_05gb: true @1559361600000000
+
+Reading data for phone#5c10102#20190501:
+Column Family stats_summary
+\tos_build: PQ2A.190401.002 @1559361600000000
+Column Family cell_plan
+\tdata_plan_10gb: true @1559361600000000
+
+Reading data for phone#5c10102#20190502:
+Column Family stats_summary
+\tos_build: PQ2A.190406.000 @1559361600000000
+Column Family cell_plan
+\tdata_plan_10gb: true @1559361600000000
+
+
+`;
+
+exports['filters should filter with col family regex 1'] = `
+Reading data for phone#4c410523#20190501:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+\tos_build: PQ2A.190405.003 @1559361600000000
+
+Reading data for phone#4c410523#20190502:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+\tos_build: PQ2A.190405.004 @1559361600000000
+
+Reading data for phone#4c410523#20190505:
+Column Family stats_summary
+\tconnected_cell: 0 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+\tos_build: PQ2A.190406.000 @1559361600000000
+
+Reading data for phone#5c10102#20190501:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+\tos_build: PQ2A.190401.002 @1559361600000000
+
+Reading data for phone#5c10102#20190502:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 0 @1559361600000000
+\tos_build: PQ2A.190406.000 @1559361600000000
+
+
+`;
+
+exports['filters should filter with col qualifier regex 1'] = `
+Reading data for phone#4c410523#20190501:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+
+Reading data for phone#4c410523#20190502:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+
+Reading data for phone#4c410523#20190505:
+Column Family stats_summary
+\tconnected_cell: 0 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+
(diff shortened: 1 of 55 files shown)


samples/filterSnippets.js
```diff
@@ -0,0 +1,351 @@
+/**
```
JustinBeckwith: I know this is annoying, but could I trouble you to rebase and copy one of the existing license headers?  We made all of them consistent :)

samples/filterSnippets.js
```diff
@@ -0,0 +1,351 @@
...
+  // [START bigtable_filters_limit_value_regex]
+  // [START bigtable_filters_limit_timestamp_range]
+  // [START bigtable_filters_limit_block_all]
+  // [START bigtable_filters_limit_pass_all]
+  // [START bigtable_filters_modify_strip_value]
+  // [START bigtable_filters_modify_apply_label]
+  // [START bigtable_filters_composing_chain]
+  // [START bigtable_filters_composing_interleave]
+  // [START bigtable_filters_composing_condition]
+  const Bigtable = require('@google-cloud/bigtable');
```
JustinBeckwith: I'm a little surprised we don't need to destructure this with `const {Bigtable} = require('@google-cloud/bigtable');`

samples/filterSnippets.js
```diff
@@ -0,0 +1,351 @@
...
+  // [END bigtable_filters_limit_block_all]
+  // [END bigtable_filters_limit_pass_all]
+  // [END bigtable_filters_modify_strip_value]
+  // [END bigtable_filters_modify_apply_label]
+  // [END bigtable_filters_composing_chain]
+  // [END bigtable_filters_composing_interleave]
+  // [END bigtable_filters_composing_condition]
+
+  // [START bigtable_filters_limit_row_sample]
+  async function filterRowSample() {
```
JustinBeckwith: at a glance, I don't see why this method would have to be async.

samples/filterSnippets.js
```diff
@@ -0,0 +1,351 @@
...
+    const filter = {
+      row: {
+        sample: 0.75,
+      },
+    };
+    readWithFilter(filter);
+  }
+  // [END bigtable_filters_limit_row_sample]
+  // [START bigtable_filters_limit_row_regex]
+  async function filterRowRegex() {
```
JustinBeckwith: Here and below - no need for these methods to be async unless they do an `await` inside.

samples/filterSnippets.js
```diff
@@ -0,0 +1,351 @@
...
+    await table
+      .createReadStream({
+        filter,
+      })
+      .on('error', err => {
+        // Handle the error.
+        console.log(err);
+      })
+      .on('data', row => printRow(row.id, row.data))
+      .on('end', function() {
```
JustinBeckwith: We can use an arrow function here, `() => { ...`

samples/filterSnippets.js
```diff
@@ -0,0 +1,351 @@
...
+    console.log(`Reading data for ${rowkey}:`);
+
+    for (const columnFamily of Object.keys(rowData)) {
+      const columnFamilyData = rowData[columnFamily];
+      console.log(`Column Family ${columnFamily}`);
+
+      for (const columnQualifier of Object.keys(columnFamilyData)) {
+        const col = columnFamilyData[columnQualifier];
+
+        for (let i = 0; i < col.length; i++) {
```
JustinBeckwith: unless you need the index, you could also do a `for (const cell of col) { ...` to make it more consistent

samples/readSnippets.js
```diff
@@ -0,0 +1,219 @@
+/**
```
JustinBeckwith: Here and below, same ask of using the new consistent license header

samples/readSnippets.js
```diff
@@ -0,0 +1,219 @@
...
+  // [END bigtable_reads_rows]
+  // [END bigtable_reads_row_range]
+  // [END bigtable_reads_row_ranges]
+  // [END bigtable_reads_prefix]
+  // [END bigtable_reads_filter]
+
+  // [START bigtable_reads_row]
+  async function readRow() {
+    const rowkey = 'phone#4c410523#20190501';
+
```
JustinBeckwith: No need to double space the code :)

samples/test/filters.js
```diff
@@ -0,0 +1,777 @@
...
+
+const INSTANCE_ID = `nodejs-bigtable-samples-keepme`;
+const TABLE_ID = `mobile-time-series-${uuid.v4()}`.substr(0, 30); // Bigtable naming rules
+
+describe('filters', async () => {
+  const bigtable = Bigtable();
+  const instance = bigtable.instance(INSTANCE_ID);
+  let table;
+  const TIMESTAMP = new Date();
+  let TIMESTAMP_MINUS_HR = new Date(TIMESTAMP);
```
JustinBeckwith: WHY ARE THESE ALL YELLING AT ME

samples/test/filters.js
```diff
@@ -0,0 +1,777 @@
...
+  TIMESTAMP_MINUS_HR = new Date(
+    TIMESTAMP_MINUS_HR.setHours(TIMESTAMP_MINUS_HR.getHours() - 1)
+  );
+  const TIMESTAMP_SECS = TIMESTAMP.getTime() * 1000;
+  const TIMESTAMP_MINUS_HR_SECS = TIMESTAMP_MINUS_HR.getTime() * 1000;
+
+  before(async function() {
```
JustinBeckwith: Arrow function `async () => {` should do nicely

samples/test/filters.js
```diff
@@ -0,0 +1,777 @@
...
+    await table.delete().catch(console.error);
+  });
+
+  it('should filter with row sample', async () => {
+    const stdout = execSync(
+      `node filterSnippets ${INSTANCE_ID} ${TABLE_ID} filterRowSample`
+    );
+    console.log('stdout', stdout);
+    const result = `Reading data for`;
+    assert.equal(stdout.includes(result), true);
```
JustinBeckwith: Instead of this line and the prior, with chai you can do:
```js
assert.include(stdout, 'Reading data for');
```

samples/test/filters.js
```diff
@@ -0,0 +1,777 @@
...
+  });
+
+  it('should filter with row regex', async () => {
+    const stdout = execSync(
+      `node filterSnippets ${INSTANCE_ID} ${TABLE_ID} filterRowRegex`
+    );
+    const result = `Reading data for phone#4c410523#20190501:
+Column Family stats_summary
```
JustinBeckwith: This is 🍌🍌🍌.  @bcoe don't you have a fancy way of testing output you wanted us to start using for situations like this?

samples/test/reads.js
```diff
@@ -0,0 +1,321 @@
...
+const TABLE_ID = `mobile-time-series-${uuid.v4()}`.substr(0, 30); // Bigtable naming rules
+
+describe('reads', async () => {
+  const bigtable = Bigtable();
+  const instance = bigtable.instance(INSTANCE_ID);
+  let table;
+  const TIMESTAMP = new Date();
+  const TIMESTAMP_SECS = TIMESTAMP.getTime() * 1000;
+
+  before(async function() {
```
JustinBeckwith: arrow function

samples/writeBatch.js
```diff
@@ -70,7 +70,9 @@ function main(instanceId = 'YOUR_INSTANCE_ID') {
     await table.insert(rowsToInsert);
 
     console.log(
-      `Successfully wrote 2 rows: ${rowsToInsert[0].key} and ${rowsToInsert[1].key}`
+      `Successfully wrote 2 rows: ${rowsToInsert[0].key} and ${
```
JustinBeckwith: You may need to do a `rm -rf node_modules package-lock.json && npm install && npm run fix`

samples/filterSnippets.js
billyjacobson: should I do that? didn't run into an issue with it, but let me know what would be preferred

samples/writeBatch.js
billyjacobson: did that, not sure if that fixed what you wanted

samples/document-snippets/instance.js
```diff
@@ -204,6 +204,67 @@ const snippets = {
...
+    const policy = {
+      bindings: [
+        {
+          role: 'roles/bigtable.viewer',
+          members: ['user:mike@example.com', 'group:admins@example.com'],
+        },
+      ],
+    };
+
+    instance
```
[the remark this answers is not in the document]
billyjacobson: wait, this isn't even from my PR, this is from something else that got merged in somehow. Not sure how to get it out of the PR

billyjacobson:
I'm unable to pass lint because of how I am calling the functions

samples/test/filters.js
bcoe: Yeah, this is pretty rough; If we know the data is not going to change between tests, I suggest that we'd use:

https://www.npmjs.com/package/snap-shot-it

It will capture the output on the first run of tests, and then fail tests if the output happens to change in the future (at which point you can eyeball using a diff what has actually changed).

Alternatively, since this is just a smoke test, I would just test with a much smaller regex, e.g.,

```js
/Reading Data for phone#[0-9]c[0-9]{6}#[0-9]{9}/
```

I would definitely avoid putting this large of a fixture inline.

bcoe: For what it's worth, I think snapshots are a great pattern, and have been itching to try them out if the need arises in our client libraries:


https://jestjs.io/docs/en/snapshot-testing

billyjacobson: Added the snapshots

samples/document-snippets/instance.js
```diff
@@ -204,6 +204,67 @@ const snippets = {
     // [END bigtable_get_clusters]
   },
 
+  getIamPolicy: instanceId => {
+    // [START bigtable_get_instance_Iam_policy]
+    const Bigtable = require('@google-cloud/bigtable');
+    const bigtable = new Bigtable();
+    const instance = bigtable.instance(instanceId);
+
+    instance
```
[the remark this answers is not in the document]
billyjacobson: not from my PR

samples/__snapshots__/reads.js
```diff
@@ -0,0 +1,147 @@
...
+Reading data for phone#4c410523#20190501:
+Column Family stats_summary
+\tconnected_cell: 1 @1559361600000000
+\tconnected_wifi: 1 @1559361600000000
+\tos_build: PQ2A.190405.003 @1559361600000000
+
+
+`;
+
+exports['reads should read part of one row 1'] = `
```
bcoe: I really appreciate how this abstracts the fixtures from the data, thank you :+1:

samples/readSnippets.js
```diff
@@ -0,0 +1,218 @@
...
+  // [END bigtable_reads_row]
+  // [END bigtable_reads_row_partial]
+  // [END bigtable_reads_rows]
+  // [END bigtable_reads_row_range]
+  // [END bigtable_reads_row_ranges]
+  // [END bigtable_reads_prefix]
+  // [END bigtable_reads_filter]
+
+  eval(`${read}()`);
```
bcoe: I would avoid using `eval()` in general it's frowned upon by the community because it can lead to vulnerabilities in code, as an example, this would allow me to run:

```
node readSnippets.js instance-id table-id "require('rimraf').sync('/nuke/my/computer')"
```

This is also resulting in `linting` issues since it sees the method names defined but never called. As an alternative implementation, I would just do a `switch` statement ... this will correct the linting error, and make security people's eyes bulge less :laughing: 

```js
switch (read) {
  caset "printRow":
    printRow();
    break;
}
```

I know, way less cool.

billyjacobson: Alright! I actually had it that way originally, but was requested to do it as functions in Java, so I was just trying to keep things consistent. And yeah, I know about the one sample per file, I wanted to reuse the print functions and other logic since it is so repetitive here.

samples/__snapshots__/filters.js
```diff
@@ -0,0 +1,501 @@
+exports['filters should filter with row regex 1'] = `
```
JustinBeckwith: I hate this question, but do we need license headers on snapshots @bcoe?

samples/readSnippets.js
```diff
@@ -0,0 +1,217 @@
...
+  // [END bigtable_reads_rows]
+  // [END bigtable_reads_row_range]
+  // [END bigtable_reads_row_ranges]
+  // [END bigtable_reads_prefix]
+  // [END bigtable_reads_filter]
+  switch (readType) {
+    // [START bigtable_reads_row]
+    case 'readRow': {
+      const rowkey = 'phone#4c410523#20190501';
+
```
JustinBeckwith: Can we delete the unnecessary newlines here?

samples/readSnippets.js
```diff
@@ -0,0 +1,217 @@
...
+      // [END bigtable_reads_row_partial]
+      break;
+    }
+
+    // [START bigtable_reads_rows]
+    case 'readRows': {
+      const rowKeys = ['phone#4c410523#20190501', 'phone#4c410523#20190502'];
+
+      const [rows] = await table.getRows({keys: rowKeys});
+      for (let i = 0; i < rows.length; i++) {
```
JustinBeckwith: This would be more readable if expressed as:

```js
for (const row of rows) {
  printRow(row.id, row.data);
}
```

samples/readSnippets.js
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  // [START bigtable_reads_row]
+  // [START bigtable_reads_row_partial]
+  // [START bigtable_reads_rows]
+  // [START bigtable_reads_row_range]
+  // [START bigtable_reads_row_ranges]
+  // [START bigtable_reads_prefix]
+  // [START bigtable_reads_filter]
+  function printRow(rowkey, rowData) {
```
JustinBeckwith: This could absolutely be correct, so please don't block on this comment.  Are we effectively trying to iterate over the entire object structure, and just log everything?  If so, something like this could suffice:

```js
console.log(util.inspect(row, { depth: 10 }));
```

Again - what you did is probably better, just throwing an alternative approach out there
```

File: samples/package.json [hunk 23]
```diff
@@ -20,7 +20,8 @@
   "devDependencies": {
     "@google-cloud/nodejs-repo-tools": "^3.0.0",
     "chai": "^4.2.0",
-    "mocha": "^6.0.0"
+    "mocha": "^6.0.0",
+    "snap-shot-it": "^7.9.1"
   },
   "scripts": {
     "test": "mocha --timeout=60000"
```
bcoe: we should add the command here:

```
"test:snap": "SNAPSHOT_UPDATE=1 npm test",
```

So that when someone inevitably breaks the output, they can easily update it.

samples/__snapshots__/filters.js
bcoe: I advised @billyjacobson against this, the `__snapshots_` are generated automatically using the command `npm run test:snap`; I think it's fine to ignore the license header check in this PR, but we should add the ignore rule to `synthtool`'s templates.

samples/readSnippets.js
billyjacobson: Can I just do a forEach?

samples/readSnippets.js
billyjacobson: I'm trying to print it in a particular way to give people a clear reference of how the object is laid out and how to find the data they are looking for

samples/readSnippets.js
billyjacobson: I like to separate the variable declarations from the API call, but I deleted the line between the call and print so it would be consistent across the samples

samples/readSnippets.js
JustinBeckwith: I have no horse in this race 😆 @bcoe we should have prescriptive guidance on how we want this done in the samples rubric tho

PR #597, opened 2019-12-24
build: use c8 for coverage

JustinBeckwith:


bcoe:
merging for now, but we're going to want to do some research in to the new year, as to why so few of these files are executed.

PR #609, opened 2020-01-31
fix: row.exists out of memory errors

### LINKED ISSUE #608: row.exists doesn't limit cell versions, can cause out of memory exceptions

  - [x] Search the issues already opened: https://github.com/googleapis/nodejs-bigtable/issues
  - [x] Search the issues on our "catch-all" repository: https://github.com/googleapis/google-cloud-node
  - [x] Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js

#### Environment details

  - OS: macos 10.14.6, linux (on GKE)
  - Node.js version: all (discovered while on `v8.16.2`)
  - npm version: all (discovered while on `6.4.1`)
  - `@google-cloud/bigtable` version: all (discovered via `1.0.1`)

#### Steps to reproduce

  1. write a very large number of increments to a single row (a simple loop that sends increment mutations should work)
  2. run `row.exists()` against the row, monitor memory usage

#### Expected result

`row.exists()` should query enough information to accurately determine if the row exists, but not pull all versions of all cells within the row.

#### Actual result

With very active rows which are creating many versions due to increment operations, `row.exists()` fetches every version of the cell, consuming a massive amount of memory (potentially crashing node) and taking a long time to execute (while waiting for the versions) before finally resolving with the boolean result. This can cause server failures and API call timeouts.

#### Analysis

`row.exists()` fetches the entire row, including all versions of all cells and doesn’t accept an options object, so it cannot be made to limit the amount of data fetched.

Specifically, `row.exists`:
* accepts no parameters/options
* forces `gaxOptions` to an empty object
* calls `row.getMetadata`
* which calls `table.getRows`
* which calls `createReadStream`, then pipes all results in-memory to be concatenated in a single array

jschmidt-metacx:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-bigtable/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Appropriate docs were updated (if necessary) - N/A

Fixes #608 🦕

Notes:
* Contributor License Agreement process has been started (via Corporate CLA).
* `npm test` passed locally ✅ 
* `npm run samples-test` passed locally ✅ 
* `npm run system-test` failed, but in same manner as on `master` branch - is this okay?
* `npm run fix` executed before commit ✅ 
* Added test to verify new options passed
* Refactored existing tests (run against original code and new code) to not check full options object, as default contents were being changed and it would break existing tests.


jschmidt-metacx:
@googlebot I signed it!

jschmidt-metacx:
@googlebot I signed it!

kolea2:
Hi @jschmidt-metacx, thanks for the PR! I wanted to highlight @igorbernstein2's recent comment on the issue: https://github.com/googleapis/nodejs-bigtable/issues/608#issuecomment-581608230

jschmidt-metacx:
Hi @kolea2 thanks for pointing out the comment. I see the settings/filters mentioned are in `bigtable.proto` and `data.proto`, but I'm not sure how to apply that to `row.exists()`. I'm happy to change the implementation if there's a preferred approach.

If I'm reading things correctly, would this mean adding a second filter of
```javascript
          {
            value: {
            }
            column: {
              cellLimit: 1,
            },
          }
```

jschmidt-metacx:
@googlebot I signed it!

billyjacobson:
> Hi @kolea2 thanks for pointing out the comment. I see the settings/filters mentioned are in `bigtable.proto` and `data.proto`, but I'm not sure how to apply that to `row.exists()`. I'm happy to change the implementation if there's a preferred approach.
> 
> If I'm reading things correctly, would this mean adding a second filter of
> 
> ```js
>           {
>             value: {
>               strip: true
>             }
>           }
> ```

I think the filter suggested by @igorbernstein2  would look like this:

```
      const filter = [
        {
          row: {
            cellLimit: 1,
          },
        },
        value: {
          strip: true,
        }
      ];

```

Get only 1 cell for the entire row because that's all you need to ensure it exists, and then use the strip filter so that you don't get any data back from the cell which could prevent issues with very large cells taking a while to return and reduce overall latency. Sorry our filter documentation isn't very fleshed out at the moment, we are working on improving it :)

laljikanjareeya:
Closing this PR due to inactivity.
Created PR [Here](https://github.com/googleapis/nodejs-bigtable/pull/676)

PR #631, opened 2020-03-06
feat: move to Typescript code generation

xiaozhenliu-gg5:


xiaozhenliu-gg5:
Send out cl/299452868 for fixing the timeout setting in `admin/v2` which blocks system-test. Other than that, all tests are green.

synth.py
```diff
@@ -5,46 +5,50 @@
...
+    "main-service": f"bigtable"
+    },
+    proto_path=f'/google/bigtable/{version}',
+    extra_proto_files=['google/cloud/common_resources.proto'],
+  )
 
 s.copy(
   v2_library,
-  excludes=['package.json', 'README.md', 'src/index.js', 'src/v2/index.js']
+  excludes=['package.json', 'README.md', 'src/index.ts', 'src/v2/index.ts', 'tsconfig.json', 'tslint.json']
```
JustinBeckwith: If we truly need to ignore `index.ts`, `tsconfig`, etc, we need comment blocks explaining why please. 

Also, I don't want to ignore `tsconfig.json` and `tslint.json`

steffnay:
@xiaozhenliu-gg5 @bcoe This looks good to me. I'm actually not super familiar with this library, though, so I don't think I can be much help with the tests

synth.py
xiaozhenliu-gg5: `bigtable` is a little bit special, that it combines `admin/v2` and `v2` client files into `src/v2`. So there's manual work in `src/v2/index.ts` and `src/index.ts` to add AdminClients.

And for `tsconfig.json` & `tslint.json` it has enabled some setting like ` "noImplicitAny": false` to make sure the manual layer `src/*.ts` pass the lint check. 
I can surely remove these files from `excluded = []` and manually modify json files in synth.py, but that may increase many more lines in synth.py. Thank you!

JustinBeckwith: Oh gross! 🙈  Skipping these for now is fine.  Can you please make sure there's a followup bug in the repo to fix the types and disable this though?  Maybe add a comment saying as much in this doc, referencing the issue?

xiaozhenliu-gg5: yes, sure! I added some comments in synth.py, and I will also create bugs for them for better tracking. Thanks for the suggestions!

package.json
```diff
@@ -69,38 +68,45 @@
     "through2": "^3.0.0"
   },
   "devDependencies": {
+    "@types/mocha": "^7.0.2",
```
alexander-fenster: nit: would be great to keep it sorted :)

alexander-fenster:
@xiaozhenliu-gg5 Please remove `src/v2/.DS_Store ` from the change :)

synth.py
```diff
@@ -5,46 +5,53 @@
...
 
 logging.basicConfig(level=logging.DEBUG)
-
-gapic = gcp.GAPICGenerator()
-v2_library = gapic.node_library(
+version='v2'
+gapic = gcp.GAPICMicrogenerator()
+v2_library = gapic.typescript_library(
   'bigtable',
   'v2',
```
alexander-fenster: if you have `version='v2'`, let's use it here :)

xiaozhenliu-gg5: sure!

package.json
xiaozhenliu-gg5: of course, sorted it by `npx sort-package-json` :)

Issue #635, opened 2020-03-09
enable "noImplicitAny" check in tsconfig.json


xiaozhenliu-gg5:
Clean up synth.py.
we don't wanna exclude `tsconfig.json` in generated client library.
reference: #https://github.com/googleapis/nodejs-bigtable/pull/631

JustinBeckwith:
Yup that's a thing I did!

PR #641, opened 2020-03-15
fix: improve typescript types for Table class

JustinBeckwith:
Incrementally inching towards #635.  This ensures proper types in `table.ts`.

### DIFF
--- a/test/table.ts
+++ b/test/table.ts
@@ -95,7 +95,7 @@ describe('Bigtable/Table', function() {
   let TABLE_NAME;
 
   let Table: typeof tblTypes.Table;
-  let table: tblTypes.Table;
+  let table: any;
 
   before(function() {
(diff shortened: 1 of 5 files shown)


test/table.ts
```diff
@@ -95,7 +95,7 @@ describe('Bigtable/Table', function() {
   let TABLE_NAME;
 
   let Table: typeof tblTypes.Table;
-  let table: tblTypes.Table;
+  let table: any;
```
alexander-fenster: don't we need eslint-ignore here? how does it work?

JustinBeckwith: Yikes, good call out.  For some reason we don't...  will fix this after we have user land types working properly (more important that tests)

kolea2:
LGTM thank you!

PR #642, opened 2020-03-17
fix(docs): cluster option name must be id

### LINKED ISSUE #640: docs mention cluster option "name", must be "id"
Example code in the docs (such as [here](https://googleapis.dev/nodejs/bigtable/2.3.1/Instance.html#create-examples)) suggests setting a `name` property in a cluster options object. That's wrong, should be `id`. Otherwise the library tries to create a cluster with id `"undefined"` which I think is also fishy.

laljikanjareeya:
Fixes #640


bcoe:
@laljikanjareeya this seems to be breaking a sample test:

```
Failed to read: projects/{1046198160504}/instances/nodejs-bigtable-samples-keepme/tables/mobile-time-series
```

PR #643, opened 2020-03-17
build: remove duplicate proto generation

JustinBeckwith:
Following up on feedback from https://github.com/googleapis/nodejs-bigtable/pull/641.


alexander-fenster:
If it just works, we're lucky :)  (It's FAR from being that simple in Spanner)

PR #646, opened 2020-03-18
fix: improve TypeScript types

JustinBeckwith:


src/filter.ts
```diff
@@ -156,7 +160,7 @@ export class Filter {
    * //   startTest2Exclusive: 'value3'
    * // }
    */
-  static createRange(start, end, key) {
+  static createRange(start, end, key: string) {
```
alexander-fenster: looks like... Pascal. What are the types for `start` and `end`? :)

src/filter.ts
```diff
@@ -894,23 +892,22 @@ export class Filter {
    *   }
    * ];
    */
-  value(value) {
+  value(value): void {
```
alexander-fenster: missing type for the parameter

src/index.ts
```diff
@@ -917,5 +948,4 @@ promisifyAll(Bigtable, {
 module.exports = Bigtable;
```
alexander-fenster: This is unrelated, but I don't like this :)  We can change this to proper TypeScript exports.

JustinBeckwith:
Trying to do this bit by bit :) I find the larger the PR the less likely it is to work 😂 But yeah, noImplicitAny coming in like a 5 PR set.

src/filter.ts
JustinBeckwith: Added some cheap well know ones first.  This library is a trip man.  Will come back around for completeness again.

PR #650, opened 2020-03-19
fix: improve TypeScript types

JustinBeckwith:


src/chunktransformer.ts
```diff
@@ -46,7 +46,7 @@ export interface Qualifier {
 }
 export interface Row {
   key?: Value;
-  data?: Data;
+  data?: any;
```
alexander-fenster: :)

src/chunktransformer.ts
```diff
@@ -354,10 +354,10 @@ export class ChunkTransformer extends Transform {
...
     this.validateNewRow(chunk, newRowKey);
     if (chunk.familyName && chunk.qualifier) {
-      const row = this.row;
-      row!.key = newRowKey;
-      row!.data = {} as Data;
-      this.family = row!.data![chunk.familyName.value] = {} as Family;
+      const row = this.row!;
+      row.key = newRowKey;
+      row.data = {};
+      this.family = row.data[chunk.familyName.value] = {} as Family;
```
alexander-fenster: am I dreaming, or am I really seeing the good old `a = b = {}` in the real-life JavaScript code? o_O

src/filter.ts
```diff
@@ -372,32 +416,31 @@ export class Filter {
...
-      column = {
-        name: column,
+  column(column: RegExp | string | {}): void {
+    let col: Column;
+    if (typeof column !== 'object') {
+      col = {
+        name: column as string,
       };
+    } else {
+      col = column as Column;
```
alexander-fenster: Achtung! It does not work for `RegExp` as you expect.

```
> const is = require('is')
undefined
> r = new RegExp(/.../)
/.../
> is.object(r)
false
> typeof r
'object'
> is.regexp(r)
true
```

src/filter.ts
```diff
@@ -658,30 +701,31 @@ export class Filter {
...
-    if (!is.object(row)) {
-      row = {
+  row(row: Row | string | RegExp | string[]): void {
+    let r: Row;
+    if (typeof row !== 'object') {
+      r = {
         key: row,
       };
+    } else {
+      r = row as Row;
```
alexander-fenster: same comment about `RegExp`

src/chunktransformer.ts
JustinBeckwith: Trying not to change the functionality here, but yeah....

src/chunktransformer.ts
JustinBeckwith: FINE

src/filter.ts
JustinBeckwith: Super nice catch.  Fixed!

src/filter.ts
JustinBeckwith: Put it back

JustinBeckwith:
@alexander-fenster ready for another look (this time with tests passing!)

src/chunktransformer.ts
alexander-fenster: You know what, I'm taking this back! I just realized that it might be what they actually wanted to assign _the same_ object to both `this.family` and `row.data[chunk.familyName.value]`. I'm afraid of this code :(

PR #654, opened 2020-03-19
fix: cluster id store as undefined in createInstance request

### LINKED ISSUE #652: id save as undefined if not provided in createInstance should throw an error
#### Environment details

  - OS: any
  - Node.js version: any
  - npm version: any
  - `@google-cloud/bigtable` version: any

#### Steps to reproduce
```ts
function createInstance() {
    const bigtable = new Bigtable();
    const instance = bigtable.instance('test-instance');
    const instanceOptions = {
        clusters: [
            {
                nodes: 3,
                location: 'us-central1-f',
                storage: 'ssd',
            }
        ],
        labels: { 'prod-label': 'prod-label' },
    };
    instance.create(instanceOptions, (err, response) => {
        if (err) {
            console.log(err);
        }
        console.log(`Created Instance successfully.`);
    });
}
```

laljikanjareeya:
Fixes #652

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -623,6 +623,9 @@ export class Bigtable {
     const callback =
       typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
 
+    if (arrify(options.clusters).some(cluster => !cluster.id)) {
+      throw new Error('An id of cluster is required to create an instance.');
+    }
     const reqOpts: any = {
       parent: this.projectName,
       instanceId: id,
--- a/src/index.ts
+++ b/src/index.ts
@@ -590,7 +590,9 @@ export class Bigtable {
       callback = options;
       options = {};
     }
-
+    if (arrify(options.clusters).some(cluster => !cluster.id)) {
+      throw new Error('id of cluster is required to create an instance.');
+    }
     const reqOpts: any = {
       parent: this.projectName,
       instanceId: id,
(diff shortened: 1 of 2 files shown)


test/index.ts
```diff
@@ -487,6 +487,31 @@ describe('Bigtable', function() {
...
+            nodes: 3,
+            location: 'us-central1-f',
+            storage: 'ssd',
+          },
+        ],
+        labels: {'prod-label': 'prod-label'},
+      };
+
+      bigtable.createInstance = function(id, option, callback) {
+        callback(error);
```
stephenplusplus: This test looks to be overriding the behavior of the function we're trying to test. The test should call `bigtable.createInstance` wrapped in a try/catch to assert that the expected error is thrown.

laljikanjareeya: Done!!

src/index.ts
```diff
@@ -623,6 +623,9 @@ export class Bigtable {
     const callback =
       typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
 
+    if (arrify(options.clusters).some(cluster => !cluster.id)) {
```
stephenplusplus: Could you move this check/throw down to inside of the `reduce`? We're already arrifying and iterating the clusters down there, so we might as well keep it together. I have an example below, and I made a small change to the wording of the error. Sorry for not noticing sooner!

```js
reqOpts.clusters = arrify(options.clusters!).reduce((clusters, cluster) => {
  if (!cluster.id) {
    throw new Error('A cluster was provided without an `id` property defined.');
  }

  clusters[cluster.id!] = {
    location: Cluster.getLo
```

stephenplusplus:
LGTM. Thank you!

PR #676, opened 2020-03-26
fix: row.exists out of memory errors

### LINKED ISSUE #608: row.exists doesn't limit cell versions, can cause out of memory exceptions

laljikanjareeya:
Fixes #608

### DIFF
--- a/test/row.ts
+++ b/test/row.ts
@@ -21,6 +21,7 @@ import {Mutation} from '../src/mutation.js';
 import * as rw from '../src/row';
 import {Table, Entry} from '../src/table.js';
 import {Chunk} from '../src/chunktransformer.js';
+import {CallOptions} from 'google-gax';
 
 const sandbox = sinon.createSandbox();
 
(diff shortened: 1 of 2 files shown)


test/row.ts
```diff
@@ -766,21 +767,60 @@ describe('Bigtable/Row', () => {
...
       row.exists(gaxOptions, assert.ifError);
     });
 
+    it('should pass gaxOptions to getMetadata', done => {
+      const gaxOptions = {
+        testProperty: true,
+      } as CallOptions;
+
+      sandbox.stub(row, 'getMetadata').callsFake(gaxOptions_ => {
+        assert('testProperty' in gaxOptions_);
```
stephenplusplus: Can you switch to `assert.strictEqual(gaxOptions_.testProperty, gaxOptions.testProperty);`?

laljikanjareeya: Done!!

stephenplusplus: Thank you!

PR #688, opened 2020-04-03
build: drop explicit dependency on grpc-js

JustinBeckwith:


system-test/mutate-rows.ts
```diff
@@ -22,16 +22,17 @@ const {
...
 import * as through from 'through2';
 import {EventEmitter} from 'events';
 import {ProjectIdCallback, GoogleAuth} from 'google-auth-library';
 import {PartialFailureError} from '@google-cloud/common/build/src/util';
 import {Entry} from '../src/table';
-import {CancellableStream} from 'google-gax';
+import {CancellableStream, GrpcClient} from 'google-gax';
 import {BigtableClient} from '../src/v2';
 
+const {grpc} = new GrpcClient();
```
alexander-fenster, line 34: This is smart :) I actually did not know that we can get it this way.

PR #694, opened 2020-04-05
refactor: use stream.pipeline instead of pumpify

JustinBeckwith:


stephenplusplus:
The more I'm experimenting with `stream.pipeline`, I've learned either 1) I don't get it, or 2) I don't think it fits here.

To help settle that, does anyone see if I'm doing something wrong here?

```js
const fs = require('fs')
const stream = require('stream')

function getStream() {
  const s1 = new stream.Readable()
  s1._read = () => s1.push('hi')

  const s2 = new stream.PassThrough()
  
  const s3 = new stream.Writable()
  s3._write = () => {}

  s1.on('error', () => console.log('s1 received error'))
  s2.on('error', () => console.log('s2 received error'))
  s3.on('error', () => console.log('s3 received error'))

  // Simulate error
  setTimeout(() => {
    s2.emit('error', new Error('Error from stream 2'))
  }, 2000)

  return stream.pipeline(s1, s2, s3, (err) => {
    console.log('callback received error')
  })
}

getStream()
  .on('error', err => {
    console.log('Stream listener received error')
  })
```

The docs make it seem like all of my `console.log`s should be firing, but only `s2 received error` and `Callback received error` do. But, I'm not sure we would want all the piped streams to emit errors. Example:

```js
const rowStream = pipeline(
  requestStream,
  chunkTransformer,
  // ...
```

We wouldn't want our internal "chunkTransformer" stream to receive the error from our API request stream. We would just want it destroyed and the user's stream to receive the error.

And maybe worth mentioning, `stream.pipeline`'s return value is the last stream it was passed. So in this case, `getStream()` returns `s3`. Duplexify would create a new stream and give that to our user. I'm not sure yet if this would present a problem throughout our uses of duplexify.

JustinBeckwith:
Yeah I've got more questions than answers here too 😨

PR #704, opened 2020-04-15
fix(deps): update dependency @google-cloud/common to v3 and handle stream ending

crwilcox:
co-authored by: @bcoe 

#678 fails system tests due to a change in pumpify/readable streams. Changes to which tick things are cleaned up on seems to have surfaced an unpiping issue.

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -17,6 +17,7 @@ import {promisifyAll} from '@google-cloud/promisify';
 import arrify = require('arrify');
 import {ServiceError} from 'google-gax';
 import {decorateStatus} from './decorateStatus';
+import {PassThrough} from 'stream';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const concat = require('concat-stream');
@@ -683,10 +684,10 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       rowsLimit = options.limit;
     }
 
-    const userStream = through.obj();
+    const userStream = new PassThrough({objectMode: true});
     const end = userStream.end.bind(userStream);
     userStream.end = () => {
-      rowStream.unpipe(userStream);
+      rowStream?.unpipe(userStream);
       if (activeRequestStream) {
         activeRequestStream.abort();
       }
--- a/src/table.ts
+++ b/src/table.ts
@@ -683,9 +683,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       rowsLimit = options.limit;
     }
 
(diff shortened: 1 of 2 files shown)


src/table.ts
```diff
@@ -683,16 +683,20 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       rowsLimit = options.limit;
     }
 
-    const userStream = through.obj();
+    const userStream = through.obj((rowData, enc, next) => {
+      next(null, rowData);
```
bcoe, line 687: whoops I left this in debugging, we should be able to just use ".obj".

src/table.ts
```diff
@@ -683,16 +683,20 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       rowsLimit = options.limit;
     }
 
-    const userStream = through.obj();
+    const userStream = through.obj((rowData, enc, next) => {
```
JustinBeckwith, line 686: I hate asking this question, but Imma do it anyway.  What would happen if you replaced this line with:

```js
const userStream = new stream.PassThrough({objectMode: true});
```

Theoretically ...  these are doing the exact same thing.

src/table.ts
```diff
@@ -683,16 +683,20 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+      next(null, rowData);
+    });
     const end = userStream.end.bind(userStream);
     userStream.end = () => {
+      rowStream.unpipe(userStream);
```
JustinBeckwith, line 691: Just to double check,  do we need to make sure `rowStream` is defined first?

src/table.ts
crwilcox: Sure. Ben and I hopped on a call and this came up a few times.

crwilcox: So, they are the same thing as far as I can tell. We have two other uses of through in this file. Though we may be able change those out for Readable

src/table.ts
bcoe: it might be good to add a check:

```
if (rowStream) rowStream.unpipe(userStream);
```

There is homegrown retry logic, such that `rowStream` isn't defined until a closure is called, and might be redefined if an error occurs, so the definition of `rowStream` can't easily happen above `userStream`.

crwilcox: I added a check. This is practically not an issue, but in theory it could happen.

crwilcox:
https://github.com/googleapis/nodejs-bigtable/issues/708 made to make sure we remember we may want to simplify some of this.

PR #711, opened 2020-04-17
fix: reduce through2 usage

JustinBeckwith:
This is the part where I admit to not understanding how the other stream fix y'all landed works, or if this would be subject to the same kind of problem.  I *do* know that the last 2 usages of `through2` in the codebase cause sample test failures if you swap them out with `PassThrough`.  Can I ask you both to poke holes in this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "docs-test": "linkinator docs",
     "fix": "gts fix",
     "prelint": "cd samples; npm link ../; npm install",
-    "lint": "gts fix",
+    "lint": "gts check",
     "prepare": "npm run compile",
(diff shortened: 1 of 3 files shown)


crwilcox:
The default `through.obj`seems equivalent to pass through. The other usages are nested streams to allow for retry. The other fix was due to one stream ending, but not unpiping causing write after end errors.
https://github.com/googleapis/nodejs-bigtable/blob/master/src/table.ts#L689

The instances in `table.ts` likely need to replace the usage with

```
stream.Transform(writableObjectMode: true, transform: (chunk, encoding, callback) => {
  // bespoke transform
});
```

PR #714, opened 2020-04-18
refactor: drop dependency on @google-cloud/common

JustinBeckwith:
So common is a somewhat large dependency.  It brings in a bunch of things we don't want or need here.  The only real thing I needed to lift was a simplified implementation of `PartialFailureError`, which was really easy to trim down:
https://github.com/googleapis/nodejs-common/blob/master/src/util.ts#L287

@alexander-fenster I suspect I may not even need that, in case there's something hiding in gax that I could just use.  

@crwilcox I suspect you'll like this one


alexander-fenster:
@JustinBeckwith The closest thing we have in gax is this:

```ts
export class GoogleError extends Error {
  code?: Status;
  note?: string;
}
```

but it's obviously a different thing. I'm OK with redefining this partial error class here.

PR #720, opened 2020-04-23
fix(types)!: improve types in index.ts

AVaksman:
BREAKING CHANGE: cluster.setMetadata(): only node count is updatable on an existing cluster; getInstancesCallback/Response: dropped nextQuery property as it is deprecated for this method, exposed failedLocations property; instance.createCluster(): removed unsupported params serveNodes and defaultStorageType

Improve type annotations, some clean up and minor refactoring.

### DIFF
--- a/src/cluster.ts
+++ b/src/cluster.ts
@@ -59,15 +59,20 @@ export type GetClustersCallback = (
   clusters?: Cluster[],
   apiResponse?: google.bigtable.admin.v2.IListClustersResponse
 ) => void;
+export interface SetClusterMetadataOptions {
+  nodes: number;
+}
 export type SetClusterMetadataCallback = GenericOperationCallback<
   Operation | null | undefined
 >;
-
-export interface CreateClusterOptions {
-  gaxOptions?: CallOptions;
-  location?: string;
+export interface BasicClusterConfig {
+  location: string;
   nodes: number;
   storage?: string;
+}
+
+export interface CreateClusterOptions extends BasicClusterConfig {
+  gaxOptions?: CallOptions;
 }
 export type GetClusterMetadataCallback = (
   err: ServiceError | null,
@@ -367,18 +372,15 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
   }
 
   setMetadata(
-    metadata: CreateClusterOptions
+    metadata: SetClusterMetadataOptions,
+    gaxOptions?: CallOptions
   ): Promise<SetClusterMetadataResponse>;
   setMetadata(
-    metadata: CreateClusterOptions,
-    gaxOptions: CallOptions
-  ): Promise<SetClusterMetadataResponse>;
-  setMetadata(
-    metadata: CreateClusterOptions,
+    metadata: SetClusterMetadataOptions,
     callback: SetClusterMetadataCallback
   ): void;
   setMetadata(
-    metadata: CreateClusterOptions,
+    metadata: SetClusterMetadataOptions,
     gaxOptions: CallOptions,
     callback: SetClusterMetadataCallback
   ): void;
@@ -399,7 +401,7 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
    * region_tag:bigtable_cluster_set_meta
    */
   setMetadata(
-    metadata: CreateClusterOptions,
+    metadata: SetClusterMetadataOptions,
     gaxOptionsOrCallback?: CallOptions | SetClusterMetadataCallback,
     cb?: SetClusterMetadataCallback
   ): void | Promise<SetClusterMetadataResponse> {
@@ -412,22 +414,8 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
 
     const reqOpts: ICluster = {
       name: this.name,
+      serveNodes: metadata.nodes,
     };
-
-    if (metadata.location) {
-      reqOpts.location = Cluster.getLocation_(
-        this.bigtable.projectId,
-        metadata.location
-      );
-    }
-
-    if (metadata.nodes) {
-      reqOpts.serveNodes = metadata.nodes;
-    }
-
-    if (metadata.storage) {
-      reqOpts.defaultStorageType = Cluster.getStorageType_(metadata.storage);
-    }
 
     this.bigtable.request<Operation>(
       {
--- a/src/cluster.ts
+++ b/src/cluster.ts
@@ -412,26 +412,16 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
         ? gaxOptionsOrCallback
         : ({} as CallOptions);
 
-    const reqOpts: ICluster = {
-      name: this.name,
-      serveNodes: metadata.nodes,
-    };
-
+    const reqOpts: ICluster = Object.assign(
+      {},
+      {
+        name: this.name,
+        serveNodes: metadata.nodes,
+      },
+      metadata
+    );
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    if ((metadata as any).location) {
-      reqOpts.location = Cluster.getLocation_(
-        this.bigtable.projectId,
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        (metadata as any).location
-      );
-    }
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    if ((metadata as any).storage) {
-      reqOpts.defaultStorageType = Cluster.getStorageType_(
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        (metadata as any).storage
-      );
-    }
+    delete (reqOpts as any).nodes;
 
     this.bigtable.request<Operation>(
       {
--- a/src/cluster.ts
+++ b/src/cluster.ts
@@ -417,6 +417,22 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
       serveNodes: metadata.nodes,
     };
 
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    if ((metadata as any).location) {
+      reqOpts.location = Cluster.getLocation_(
+        this.bigtable.projectId,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (metadata as any).location
+      );
+    }
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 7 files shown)


src/cluster.ts
```diff
@@ -59,16 +59,21 @@ export type GetClustersCallback = (
   clusters?: Cluster[],
   apiResponse?: google.bigtable.admin.v2.IListClustersResponse
 ) => void;
+export interface SetClusterMetadataOptions {
+  nodes: number;
```
AVaksman, line 63: Only `nodes` are updatable in cluster after it's creation

src/cluster.ts
```diff
@@ -412,23 +414,9 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
-    }
-
-    if (metadata.nodes) {
-      reqOpts.serveNodes = metadata.nodes;
-    }
-
-    if (metadata.storage) {
-      reqOpts.defaultStorageType = Cluster.getStorageType_(metadata.storage);
-    }
-
```
AVaksman, line 431: `storage` and `location` can not be updated on an existing cluster

src/index.ts
```diff
@@ -49,14 +48,14 @@ const {grpc} = new gax.GrpcClient();
 export interface GetInstancesCallback {
   (
     err: ServiceError | null,
-    result?: Instance[] | null,
-    nextQuery?: {} | null,
-    response?: google.bigtable.admin.v2.IListInstancesResponse | null
+    result?: Instance[],
+    failedLocations?: string[],
```
AVaksman, line 52: BREAKING CHANGE
exposing [`failedLocations`](https://github.com/googleapis/googleapis/blob/0d7c3565dd942d0cc41b16e895046486ae29e44e/google/bigtable/admin/v2/bigtable_instance_admin.proto#L305) in the callback/response

src/instance.ts
```diff
@@ -56,20 +58,15 @@ import {ServiceError} from 'google-gax';
 import {Bigtable} from '.';
 import {google} from '../protos/protos';
 
-export interface ClusterInfo {
-  id?: string;
-  location?: string;
-  serveNodes?: number;
-  nodes?: number;
-  storage?: string;
-  defaultStorageType?: number;
```
AVaksman, line 65: BREAKING CHANGE
Removing `serveNodes` and `defaultStorageType` as these are expected by GAPIC client, user facing parameters are `nodes` and `storage` respectively.

JustinBeckwith:
@AVaksman we prolly want to land this before shipping v3, right?  Can I trouble you to run through the open PRs and try to get these ready to go?

AVaksman:
BREAKING CHANGE: 
- cluster.setMetadata()
   - only node count is updatable on an existing cluster
   - removed `location` and `storage` from the params
- getInstancesCallback/Response
   - exposed `failedLocations` property that is also returned along with array of `Instnace`s
   - dropped `nextQuery` property as it is [deprecated](https://github.com/googleapis/googleapis/blob/2433bd50656264a2ef9f684bf646fb4d250d39ff/google/bigtable/admin/v2/bigtable_instance_admin.proto#L307-L308) for this method
- createInstance(id, options)
   - made `options` into a required param as it is not possible to create an instance without providing at least one `cluster` (part of `options` param)
- createCluster
   - removed duplicated, unsupported (by implementation), undocumented params `serveNodes` and `defaultStorageType` in favor of `nodes` and `location` respectfully

src/cluster.ts
stephenplusplus: I haven't tested this or anything, but the docs say that the UpdateCluster rpc accepts a full "Cluster" object: https://cloud.google.com/bigtable/docs/reference/admin/rpc/google.bigtable.admin.v2#google.bigtable.admin.v2.BigtableInstanceAdmin.UpdateCluster-table

Will it return an error if certain fields are attempted to be updated?

AVaksman: It won't return an error,
This change is done to eliminate a possibility of a confusion from the user side if attempted to change a filed that is not supported.

stephenplusplus: I see, thanks. I'm not sure we should use the whitelist approach of only allowing "serveNodes", when the API could evolve someday to accept something else, and we'd be blocking it. Does the upstream team know their API docs state multiple unsupported properties can be sent during the UpdateCluster call, and maybe they would want to make it return an error from their end, or some other solution?

We could also throw from our library if we receive unsupported input, but in general, we're best to simply match what the official docs state is supported. As soon as we go off that path, things get shaky.

AVaksman: > Does the upstream team know their API docs state multiple unsupported properties can be sent during the UpdateCluster call

While UpdateCluster rpc does accept a full "Cluster" object, it seems that upstream API docs do specify that among `Cluster` properties

- `name` is an _OutputOnly_ field
- `state` is an _OutputOnly_ field as well
- `location` is a _CreateOnly_ field
- `default_storage_type` is a _CreateOnly_ field as well
- leaving only `serveNodes ` field that can be updated

>  when the API could evolve someday to accept something else, and we'd be blocking it

If/when upstream API evolves to accept something else, that logic could be added as a new feature PR?

WDYT?

stephenplusplus: > If/when upstream API evolves to accept something else, that logic could be added as a new feature PR?

There's usually a big lag before we notice.

If the upstream API silently drops properties, I would think our best bet is to do the same. In terms of code, let's pass them all through. In terms of types, maybe that's where you could just list "serveNodes"? At least then, if the user knows better and is trying to set a property we haven't added to the types yet, they could have their TypeScript compiler ignore that (I think?).

AVaksman: Added the logic back, left the types with `nodes` only.

AVaksman: @stephenplusplus WDYT?

src/cluster.ts
```diff
@@ -399,34 +401,36 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
         : ({} as CallOptions);
 
     const reqOpts: ICluster = {
       name: this.name,
+      serveNodes: metadata.nodes,
     };
 
-    if (metadata.location) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    if ((metadata as any).location) {
```
stephenplusplus, line 421: If the API actually drops these parameters, I think we can skip all of the logic where we worry about formatting them.

src/cluster.ts
```diff
@@ -399,34 +401,36 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
       typeof gaxOptionsOrCallback === 'function' ? gaxOptionsOrCallback : cb!;
     const gaxOptions =
-      typeof gaxOptionsOrCallback === 'object' && gaxOptionsOrCallback
+      typeof gaxOptionsOrCallback === 'object'
         ? gaxOptionsOrCallback
         : ({} as CallOptions);
 
     const reqOpts: ICluster = {
       name: this.name,
+      serveNodes: metadata.nodes,
```
stephenplusplus, line 417: This is where I think we should pass all user-provided input straight through, as opposed to the whitelist approach where we only recognize certain parameters.

stephenplusplus: (mapping `metadata.nodes` to `serveNodes` is still good)

src/index.ts
stephenplusplus: Let's make sure we add docs for this new response argument.

stephenplusplus: Also, an example in the code block showing "if (failedLocations.length > 0) { // These locations contain instances which could not be retrieved. }"

src/cluster.ts
AVaksman: If applicable, user will need to pass the full
`location` as full path in the form of
`projects/<project>/locations/<zone_id>`

and a `defaultStorageType` (not `storage`) in upper case characters.

Done

test/cluster.ts
```diff
@@ -374,65 +374,57 @@ describe('Bigtable/Cluster', () => {
...
-        Cluster.getLocation_ = getLocation;
+        assert.strictEqual(config.reqOpts.serveNodes, options.nodes);
         done();
       };
 
       cluster.setMetadata(options, assert.ifError);
     });
 
-    it('should respect the nodes option', done => {
+    it('should accept and pass user provided unput through', done => {
```
stephenplusplus, line 391:
```suggestion
    it('should accept and pass user provided input through', done => {
```

src/index.ts
```diff
@@ -546,14 +544,9 @@ export class Bigtable {
    */
   createInstance(
     id: string,
-    optionsOrCallback?: InstanceOptions | CreateInstanceCallback,
```
stephenplusplus, line 549: Let's throw an error now that this is required (and add a test):

```ts
if (!options) {
  throw new Error('A configuration object is required to create an instance.');
}
```

src/index.ts
```diff
@@ -629,26 +624,21 @@ export class Bigtable {
...
-   * }, callback);
-   *
    * @example <caption>If the callback is omitted, we'll return a Promise.
    * </caption>
    * bigtable.getInstances().then(function(data) {
    *   const instances = data[0];
+   *
+   *   if (data[1]) {
+   *     // These locations contain instances which could not be retrieved.
+   *     const failedLocations = data[1]
```
stephenplusplus, line 640:
```suggestion
   *     const failedLocations = data[1];
```

src/index.ts
AVaksman: Done, thanks!

test/cluster.ts
AVaksman: Fixed, thanks.

src/index.ts
AVaksman: fixed, thanks!

src/index.ts
```diff
@@ -547,6 +547,11 @@ export class Bigtable {
     options: InstanceOptions,
     callback?: CreateInstanceCallback
   ): void | Promise<CreateInstanceResponse> {
+    if (!options.clusters) {
```
stephenplusplus, line 550: Since we require an object, let's throw if that's missing, too. Otherwise, this would be a TypeError ("Cannot read property 'clusters' of undefined")

AVaksman: Great point!
Done!

AVaksman:
@stephenplusplus before merging, does PR description look OK?

> BREAKING CHANGE: cluster.setMetadata(): only node count is updatable on an existing cluster; getInstancesCallback/Response: dropped nextQuery property as it is deprecated for this method, exposed failedLocations property; instance.createCluster(): removed unsupported params serveNodes and defaultStorageType

stephenplusplus:
It looks good to me!

PR #728, opened 2020-05-08
feat: Cloud functions example for Bigtable

billyjacobson:


billyjacobson:
question about npm run lint, on the CI tests it seems to be able to find the samples/functions directory, but when I run it locally it wasn't linting those files

samples/functions/test/index.test.js
```diff
@@ -0,0 +1,185 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const assert = require('assert');
+const execPromise = require('child-process-promise').exec;
```
bcoe, line 18: could se just swap this for `child_process.execSync`, one less dependency.

samples/functions/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@google-cloud/bigtable-samples",
```
bcoe, line 2: I would pull these dependencies up to `samples/package.json`, since the function framework is used in testing, but not part of the sample itself (no need to have a different `package.json`).

samples/functions/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "repository": "googleapis/nodejs-bigtable",
+  "engines": {
+    "node": ">=8.0.0"
+  },
+  "scripts": {
+    "test": "mocha test/*.test.js --timeout=20000",
+    "start": "functions-framework --target=get"
+  },
+  "dependencies": {
+    "@google-cloud/bigtable": "^2.3.2",
```
bcoe, line 15: Could `@google-cloud/functions-framework` be a development dependency? I'm wondering if we can get away with just showing the structure of the cloud function to the customer, without pulling in this dep (except for testing).

PR #761, opened 2020-07-13
feature: idiomatic table backup management implementation

omniadonum:
This is the hand-written implementation that introduces backup support for tables.

- A backup represents a table on a specific cluster
- Backups reside local to a cluster
- A backup is restored as a table to the cluster from where it originated
- proto methods have been mirrored to the relevant classes based on parent resource requirements:
  - `createBackup`, `getBackup`, `listBackups`, `updateBackup`, and `deleteBackup` are in `Cluster`
  - `restoreTable` is in `Instance`
- Friendlier methods that are more intuitive wrap the aforementioned proto proxies requiring less input criteria:
  - `table.backup()` to create a backup
  - Introduced a `Backup` class with `delete`, `get`, `restore`, and `update` methods
  - Dates are JS-friendly and params/responses convert to/from Timestamp Struct, Date, and PreciseDate
- Some methods return a `Backup` or `Backup[]` where applicable, but Longrunning Operations (LROs) do not
  - LROs do not transform their result into a `Backup`, this is left to the user
  - LROs return POJOs instead
  - Added `cluster.asBackup` casting helper to convert POJOs into `Backup`
- Metadata results were sparse for the response structs for backups, the "backup" _is_ the metadata in most cases
- Aside: wrote tests using async/await
  - promisify is refactored as pass through to allow for promises
  - Changed this in unrelated test suites and fixed signature tests where they were incorrect
- When testing, instances _cannot_ be deleted if they have backups defined
  - Remove all backups before removing the instances
  - A fully parallel backup reaper was introduced to help, it runs before the instance reaper

PR checklist:
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary)

Fixes ? 🦕

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,8 @@
     "api-documenter": "api-documenter yaml --input-folder=temp"
   },
   "dependencies": {
+    "@google-cloud/paginator": "^3.0.0",
+    "@google-cloud/precise-date": "^2.0.1",
     "@google-cloud/projectify": "^2.0.0",
     "@google-cloud/promisify": "^2.0.0",
     "arrify": "^2.0.0",
--- a/samples/backups.js
+++ b/samples/backups.js
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 // Imports the Google Cloud client library
-const {Bigtable, Backup} = require('@google-cloud/bigtable');
+const {Bigtable} = require('@google-cloud/bigtable');
 const uuid = require('uuid');
 const {inspect} = require('util');
 
@@ -51,18 +51,24 @@ async function runBackupOperations(
   // The backup cannot be interacted with (restore, update, delete) until it is ready.
   const [backupMeta] = await backupOperation.promise();
   console.log(
-    `Backup is now "${backupMeta.state}" (READY = 2, i.e. available).`
+    `Backup "${backupMeta.name}" is now "${backupMeta.state}" (READY = 2, i.e. available).`
   );
-  // [END bigtable_create_backup]
 
-  // [START bigtable_table_backup]
+  // The resulting backup can be cast to a helpful Backup object.
+  const backup = cluster.asBackup(backupMeta);
+  console.log(`Backup "${backup.backupId}" expires at ${backup.expireDate}.`);
+
   // An alternative way is to call backup right on a Table which will
   // automatically select the first cluster that has a READY cluster state.
   const [backupOperationFromTable] = await table.backup(`${backupID}-2`, {
     expireTime: new Date(Date.now() + 7 * 60 * 60 * 1000), // 7 hours from now
   });
-  const [backupFromTable] = await backupOperationFromTable.promise();
-  // [END bigtable_table_backup]
+  const [backupMetaFromTable] = await backupOperationFromTable.promise();
+  const backupFromTable = cluster.asBackup(backupMetaFromTable);
+  console.log(
+    `Backup "${backup.name}" was created from "${backup.sourceTable}".`
+  );
+  // [END bigtable_create_backup]
 
   // [START bigtable_list_backups]
   const [existingBackups] = await cluster.listBackups();
@@ -148,14 +154,21 @@ async function runBackupOperations(
     },
   });
   console.log('Deleted backup with ID %s, no response is returned.', backupID);
-  // [END bigtable_delete_backup]
 
-  // need to delete that other backup example
-  await new Backup(bigtable, backupFromTable).delete({
+  // Alternatively, if a backup object is available, use the `delete` helper.
+  const backupPartial = cluster.asBackup({
+    name: backupFromTable.name, // used from another example
+  });
+
+  // While the Backup instance only really has the name defined,
+  // there is enough information to perform a delete operation.
+  // But, this `delete` helper could even be used after a `listBackups` call.
+  await backupPartial.delete({
     gaxOptions: {
       timeout: 50 * 1000,
     },
   });
+  // [END bigtable_delete_backup]
 }
 
 require('yargs')
--- a/samples/backups.js
+++ b/samples/backups.js
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 // Imports the Google Cloud client library
-const {Bigtable} = require('@google-cloud/bigtable');
+const {Bigtable, Backup} = require('@google-cloud/bigtable');
 const uuid = require('uuid');
 const {inspect} = require('util');
 
@@ -151,8 +151,7 @@ async function runBackupOperations(
   // [END bigtable_delete_backup]
 
   // need to delete that other backup example
-  // await new Backup(backupFromTable).delete();
-  await cluster.deleteBackup(`${backupID}-2`, {
+  await new Backup(bigtable, backupFromTable).delete({
     gaxOptions: {
       timeout: 50 * 1000,
     },
--- a/samples/backups.js
+++ b/samples/backups.js
@@ -1,0 +1,114 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Imports the Google Cloud client library
+const {Bigtable} = require('@google-cloud/bigtable');
+const {inspect} = require('util');
+
+async function runBackupOperations(
+  instanceID,
+  clusterID,
+  tableID,
+  optionalBackupID
+) {
+  const bigtable = new Bigtable();
+  const instance = bigtable.instance(instanceID);
+  const cluster = instance.cluster(clusterID);
+
+  const backupID = optionalBackupID || generateBackupId(cluster, tableID);
+
+  // [START bigtable_create_backup]
+  const table = instance.table(tableID);
+  const [backupOperation] = await cluster.createBackup(table, backupID, {
+    expireTime: new Date(Date.now() + 24 * 60 * 60 * 1000), // in 1 day from now
+  });
+  console.log(
+    'Started a table backup operation:',
+    inspect(backupOperation.latestResponse.name, {depth: null, colors: true})
+  );
+
+  // The long running operation (LRO) "backupOperation" provides an EventEmitter interface
+  // or a promise to wait for completion.
+  // The backup cannot be interacted with (restore, update, delete) until it is ready.
+  await backupOperation.promise();
+  console.log('Backup is now "READY" (available).');
+  // [END bigtable_create_backup]
+
+  // [START bigtable_list_backups]
+  const [existingBackups] = await cluster.listBackups();
+  console.log(
+    'These backups exist:',
+    inspect(existingBackups, {depth: null, colors: true})
+  );
+  // [END bigtable_list_backups]
+
+  // [START bigtable_get_backup]
+  // Get information about a backup, like when it will expire or how big it is.
+  // NOTE: No need to poll with this RPC to get backup ready status,
+  //  instead use the EventEmitter or promise interface of the LRO.
+
+  // Convert GCPs Timestamp into a JavaScript Date, this will be used below...
+  function timestampToDate(timestamp) {
+    return new Date(
+      Number(timestamp.seconds) * 1000 + Math.floor(timestamp.nanos / 1000000)
+    );
+  }
+
+  const [backupInfo] = await cluster.getBackup(backupID);
+  console.log('The backup is %s bytes in size.', backupInfo.sizeBytes);
+
+  const expireDate = timestampToDate(backupInfo.expireTime);
+  console.log('The backup will auto delete at', expireDate.toISOString());
+
+  const endDate = timestampToDate(backupInfo.endTime);
+  console.log('The backup finished being created at', endDate.toISOString());
+  // [END bigtable_get_backup]
+
+  // [START bigtable_delete_backup]
+  await cluster.deleteBackup(backupID);
+  console.log('Deleted backup with ID %s, no response is returned.', backupID);
+  // [END bigtable_delete_backup]
+}
+
+function generateBackupId(cluster, table) {
+  return [cluster.id, table.id || table, Date.now()].join('-').slice(-50); // truncate to max backup id length of 50
+}
+
+require('yargs')
+  .demand(1)
+  .command(
+    'run',
+    'Create a backup of a table, restore it, and finally delete it.',
+    {},
+    argv =>
+      runBackupOperations(argv.instance, argv.cluster, argv.table, argv.backup)
+  )
+  .example(
+    'node $0 run --instance [instanceID] --cluster [clusterID] --table [tableID] [--backup [backupID]]',
+    'Create a backup, restore it, and finally delete it.'
+  )
+  .wrap(120)
+  .nargs('instance', 1)
+  .nargs('cluster', 1)
+  .nargs('table', 1)
+  .nargs('backup', 1)
+  .describe('instance', 'Cloud Bigtable Instance ID')
+  .describe('cluster', 'Cloud Bigtable Cluster ID')
+  .describe('table', 'Cloud Bigtable Table ID')
+  .describe('backup', 'A unique Cloud Bigtable Backup ID')
+  .demandOption(['instance', 'cluster', 'table'])
+  .recommendCommands()
+  .epilogue('For more information, see https://cloud.google.com/bigtable/docs')
+  .help()
+  .strict().argv;
--- a/samples/backups.js
+++ b/samples/backups.js
@@ -17,8 +17,9 @@ const {Bigtable} = require('@google-cloud/bigtable');
 const uuid = require('uuid');
 const {inspect} = require('util');
 
-function thirtyMinutesFromNow() {
-  return new Date(Date.now() + 24 * 60 * 60 * 1000);
+function xHoursFromNow(x) {
+  const pad = 60 * 1000;
+  return new Date(Date.now() + x * 60 * 60 * 1000 + pad);
 }
 
 function generateBackupId() {
@@ -39,12 +40,10 @@ async function runBackupOperations(
 
   // [START bigtable_create_backup]
   const table = instance.table(tableID);
-  const expireTime = thirtyMinutesFromNow(); // accepts either a Date or an ITimestamp
-  const [backupOperation] = await cluster.createBackup(
-    table,
-    backupID,
-    expireTime
-  );
+  const expireTime = xHoursFromNow(6); // accepts either a Date or an ITimestamp
+  const [backupOperation] = await cluster.createBackup(table, backupID, {
+    expireTime,
+  });
   console.log(
     'Started a table backup operation:',
     inspect(backupOperation.latestResponse.name, {depth: null, colors: true})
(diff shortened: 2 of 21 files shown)


zamnuts:
System and samples tests are both currently failing due to:
```
Error: 12 UNIMPLEMENTED: Method not found.
```
All necessary methods are present (in master) within `v2.BigtableTableAdminClient`, and these work locally against my test cluster. I expect this is an issue with the RPC version available to kokoro?

AVaksman:
It seems the `Backup` class structure here is a diversion from the pattern followed by other classes (`Instance`, `Cluster`, `Table`, `AppProfile`) in the library, where the instance of a class represents an object that can interact with a Cloud Bigtable object. 

Factory reference usually would take a pathway to service (like `instance` object, `instance.bigtable.request`) and an `id`, from here an object can interact with  Cloud Bigtable.
```ts
const backup = instance.backup('my-backup', 'my-cluster');
backup.getMetadata();
backup.update(...);
```

src/cluster.ts
```diff
@@ -80,6 +90,103 @@ export type GetClusterMetadataCallback = (
...
+   * when the backup will automatically be deleted. This must be at a
+   * minimum 6 hours from the time of the backup request and a maximum of 30
+   * days.
+   */
+  expireTime?: BackupTimestamp;
+}
+
+export interface CreateBackupOptions {
+  gaxOptions?: CallOptions;
+}
```
AVaksman, line 106: Perhaps we should package all but `id` params in one object
```ts
export interface CreateBackupOptions {
  tableId: string; // tableId might be enough since we can build full table path
  expireTime: BackupTimestamp;
  gaxOptions?: CallOptions;
}
```

Similar to `creteInstance`, `createCluster`, `createTable` functions.

src/cluster.ts
```diff
@@ -80,6 +90,103 @@ export type GetClusterMetadataCallback = (
...
 
+export type BackupTimestamp = google.protobuf.ITimestamp | PreciseDate | Date;
+export interface ModifiableBackupFields {
+  /**
+   * The ITimestamp (Date or PreciseDate will be converted) representing
+   * when the backup will automatically be deleted. This must be at a
+   * minimum 6 hours from the time of the backup request and a maximum of 30
+   * days.
+   */
+  expireTime?: BackupTimestamp;
```
AVaksman, line 101: `expireTime` could be required here since it is required for creation of a backup, and is the only field that can be updated on an existing backup.

src/instance.ts
```diff
@@ -136,6 +136,23 @@ export type SetInstanceMetadataCallback = (
...
+export interface RestoreTableOptions {
+  gaxOptions?: CallOptions;
+}
+export type RestoreTableCallback = (
+  err: ServiceError | null,
+  apiResponse?: LROperation<
+    google.bigtable.admin.v2.ITable,
+    google.bigtable.admin.v2.IRestoreTableMetadata
+  >
+) => void;
```
AVaksman, line 148: Long running operation functions that return a resource object usually have the return signatures as in this example

https://github.com/googleapis/nodejs-bigtable/blob/5e47bbc3d1304fb06cd2e6df5ff17c2a197a1352/src/instance.ts#L102-L111

src/instance.ts
```diff
@@ -905,6 +922,100 @@ Please use the format 'my-instance' or '${bigtable.projectName}/instances/my-ins
...
+        google.bigtable.admin.v2.IRestoreTableMetadata
+      >
+    >(
+      {
+        client: 'BigtableTableAdminClient',
+        method: 'restoreTable',
+        reqOpts,
+        gaxOpts: options.gaxOptions,
+      },
+      callback // TODO cast as Table
```
AVaksman, line 1015: should splice in a table Object in the callback.

samples/backups.js
```diff
@@ -0,0 +1,200 @@
...
+      // if DEADLINE_EXCEEDED is witnessed, then deleteBackup exceeds the default timeout
+      // this overrides it
+      timeout: 50 * 1000,
+    },
+  });
+  console.log('Deleted backup with ID %s, no response is returned.', backupID);
+
+  // Alternatively, if a backup object is available, use the `delete` helper.
+  const backupPartial = cluster.asBackup({
+    name: backupFromTable.name, // used from another example
```
crwilcox, line 160: I don't understand this comment?

samples/document-snippets/cluster.js
```diff
@@ -129,6 +129,86 @@ const snippets = {
...
+    // [START bigtable_cluster_delete_backup]
+    const {Bigtable} = require('@google-cloud/bigtable');
+    const bigtable = new Bigtable();
+    const instance = bigtable.instance(instanceId);
+    const cluster = instance.cluster(clusterId);
+
+    cluster
+      .deleteBackup(backupId, {
+        gaxOptions: {
+          timeout: 50 * 1000, // this op takes a good 30-40 seconds
```
crwilcox, line 201: nit: time cannot itself be good or bad ;) also, probably shouldn't abbreviate op

src/backup.ts
```diff
@@ -0,0 +1,302 @@
...
+import * as Long from 'long';
+
+export class Backup implements google.bigtable.admin.v2.IBackup {
+  name?: string | null;
+  sourceTable?: string | null;
+  expireTime?: google.protobuf.ITimestamp | null;
+  startTime?: google.protobuf.ITimestamp | null;
+  endTime?: google.protobuf.ITimestamp | null;
+  sizeBytes?: number | Long | string | null;
+  state?:
```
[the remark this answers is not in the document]
crwilcox: @AVaksman I don't quite follow this comment. Could you explain a bit? Or maybe this comment is showing in the wrong place?

src/cluster.ts
```diff
@@ -80,6 +90,103 @@ export type GetClusterMetadataCallback = (
   apiResponse?: IOperation | null
 ) => void;
 
+export type BackupTimestamp = google.protobuf.ITimestamp | PreciseDate | Date;
+export interface ModifiableBackupFields {
+  /**
+   * The ITimestamp (Date or PreciseDate will be converted) representing
+   * when the backup will automatically be deleted. This must be at a
+   * minimum 6 hours from the time of the backup request and a maximum of 30
```
crwilcox, line 98: this min/max is mentioned but I didn't see anything validating this.

src/cluster.ts
```diff
@@ -435,14 +546,451 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
+        if (resp) {
+          backup = this.asBackup(resp);
+        }
+
+        callback(err, backup);
+      }
+    );
+  }
+
+  listBackups(options?: ListBackupsOptions): Promise<ListBackupsResponse>;
```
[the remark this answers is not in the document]
crwilcox: are the other underlying rpcs similarly listTable and not getTables? Either way this may make sense for interface consistency.

crwilcox (changes requested):

src/cluster.ts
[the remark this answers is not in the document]
AVaksman: Yes, API has `getInstances`, `getClusters`, `getTables`, `getAppProfiles`, ... that wrap underlying list<Resource> RPCs.

src/backup.ts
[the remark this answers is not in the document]
AVaksman: What I meant to say it usually all proto properties of a resource are located in the `metadata` property of the resource object.
Instance.metadata has `type`, `labels`, `displayName`, `state`
Cluster.metadata has `defaultStorageType`, `serveNodes`, `location`, `state`
Table.metadata has `clusterStates`, `columnFamilies`, `granularity`
Class objects usually have properties `id`, `name`, `metadata`, `bigtable`, `instance`(if it is not an instance of class `Instance`).

src/instance.ts
```diff
@@ -136,6 +136,23 @@ export type SetInstanceMetadataCallback = (
 ) => void;
 export type SetInstanceMetadataResponse = [google.protobuf.Empty];
 
+export interface RestoreTableOptions {
```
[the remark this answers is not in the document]
zamnuts: using only `gaxOptions` without the interface wrapper does not conform to the other options patterns. having mostly `foo(..., {gaxOptions})`, and then this one would be `foo(..., gaxOptions)`. so i'm not going to implement this suggestion.

src/backup.ts
```diff
@@ -0,0 +1,302 @@
...
+   * Deletes this pending or completed Cloud Bigtable backup.
+   *
+   * @param {DeleteBackupOptions} [options]
+   * @return {Promise<DeleteBackupResponse>}
+   * @see {Cluster#deleteBackup}
+   *
+   * @example <caption>include:samples/document-snippets/cluster.js</caption>
+   * region_tag:bigtable_cluster_delete_backup
+   */
+  delete(options?: DeleteBackupOptions): Promise<DeleteBackupResponse> {
```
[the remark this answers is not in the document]
zamnuts: to be honest, i was hoping to get away with just the promise interface, hoping we could start just avoiding callbacks altogether. but i'll add support for both.

src/backup.ts
```diff
@@ -0,0 +1,302 @@
...
+      return Promise.reject(
+        new TypeError('A backup name is required to get a backup.')
+      );
+    }
+
+    return this._bigtable
+      .instance(this.instanceId)
+      .cluster(this.clusterId)
+      .getBackup(this.backupId, options);
+  }
```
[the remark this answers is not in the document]
zamnuts: Maybe i don't understand the API's definition of "metadata," but the necessary/valuable information in the backup object (e.g. `name`, `expireTime`, `sourceTable`) has sibling properties that are more "meta" (e.g. `state`, `sizeBytes`). When you say "object," does that mean the class instance (e.g. `Table`) of the actual object (`{name: '.../tables/my-table'}`)? Otherwise, given the placement of the properties, this becomes overkill to separate the two, so for now I won't implement `getMetdata`, and _will_ implement `get`.

src/cluster.ts
```diff
@@ -435,14 +546,451 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
+ */
+Cluster.prototype.listBackupsStream = paginator.streamify<Backup>(
+  'listBackups'
+);
+
+/*! Developer Documentation
+ *
+ * These methods can be auto-paginated.
+ */
+paginator.extend(Cluster, ['listBackups']);
```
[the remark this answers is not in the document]
zamnuts: I went through pagination thoroughly and recall finding that gapic was incompatible with what was expected from the backups RPC, at least per the generated types and protobufs. The implementation as it stands using the paginator library works as expected.

src/cluster.ts
zamnuts: from what i found, this min/max was _not_ documented. the grpc response will respond with an error outside of these bounds. i opted to _not_ perform client-side validation in case this changes in the future.

package.json
```diff
@@ -47,6 +47,8 @@
     "api-documenter": "api-documenter yaml --input-folder=temp"
   },
   "dependencies": {
+    "@google-cloud/paginator": "^3.0.0",
```
JustinBeckwith, line 50: `@google-cloud/paginator` was explicitly removed here:
https://github.com/googleapis/nodejs-bigtable/commit/8a2a74e00c1241c89bf344db0b82440577270223

@AVaksman want to work with @zamnuts on what we're supposed to be doing here now?

samples/backups.js
```diff
@@ -0,0 +1,200 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Imports the Google Cloud client library
+const {Bigtable} = require('@google-cloud/bigtable');
+const uuid = require('uuid');
+const {inspect} = require('util');
+
+function generateBackupId() {
```
JustinBeckwith, line 20: These almost always belong in the samples test, and are passed in as a parameter to the main function (will keep reading though)

samples/backups.js
```diff
@@ -0,0 +1,200 @@
...
+  return `gcloud-tests-${uuid.v4()}`.substr(0, 48);
+}
+
+async function runBackupOperations(
```
JustinBeckwith, line 24: It's important that we have an outer `main` function, with an inner async function due to the way these get embedded in cloudsite.  Please follow the pattern here:
https://github.com/googleapis/nodejs-bigtable/blob/master/samples/quickstart.js

samples/backups.js
```diff
@@ -0,0 +1,200 @@
...
+  const table = instance.table(tableID);
+  const expireTime = new Date(Date.now() + 7 * 60 * 60 * 1000); // 7 hours from now
+
+  // expireTime can be either a Date, PreciseDate, or an ITimestamp
+  const [backupOperation] = await cluster.createBackup(table, backupID, {
+    expireTime,
+  });
+  console.log(
+    'Started a table backup operation:',
+    inspect(backupOperation.latestResponse.name, {depth: null, colors: true})
```
JustinBeckwith, line 46: Typically we don't drop down to `util.inspect`, but just drop individual properties in the logs.  Since this is a `name`, is it just a `string` type?

samples/backups.js
```diff
@@ -0,0 +1,200 @@
...
+  // But, this `delete` helper could even be used after a `listBackups` call.
+  await backupPartial.delete({
+    gaxOptions: {
+      timeout: 50 * 1000,
+    },
+  });
+  // [END bigtable_delete_backup]
+}
+
+require('yargs')
```
JustinBeckwith, line 174: In our modern samples we do not use `yargs`, and instead use the model found in https://github.com/googleapis/nodejs-bigtable/blob/master/samples/quickstart.js

samples/document-snippets/cluster.js
```diff
@@ -129,6 +129,86 @@ const snippets = {
       });
     // [END bigtable_cluster_set_meta]
   },
+
+  createBackup: (instanceId, clusterId, tableId, backupId) => {
```
JustinBeckwith, line 133: These snippets are tech debt :/  I believe we don't have tests for many of them.  For any new sample code, please create a new file, and follow https://github.com/googleapis/nodejs-bigtable/blob/master/samples/quickstart.js as a guide.  One per file!

samples/test/backups.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+const instanceId = `gcloud-tests-${uuid.v4()}`.substr(0, 30); // Bigtable naming rules
+const clusterId = `gcloud-tests-${uuid.v4()}`.substr(0, 30); // Bigtable naming rules
+const tableId = `gcloud-tests-${uuid.v4()}`.substr(0, 30); // Bigtable naming rules
+const backupId = `gcloud-tests-${uuid.v4()}`.substr(0, 30); // Bigtable naming rules
+const instance = bigtable.instance(instanceId);
+
+describe('backups', () => {
+  before(async () => {
+    await instance.create({
```
JustinBeckwith, line 35: I seem to remember there being a fairly low cap on the number of clusters you can have at a time.  I created `samples/test/util.js` to help do a check-in/out re-use thing to keep the number of clusters we create down if possible.

samples/test/cluster.js
```diff
@@ -75,6 +77,22 @@ describe.skip('Cluster Snippets', () => {
     clusterSnippets.setMeta(INSTANCE_ID, CLUSTER_ID);
   });
 
+  it('should create a backup of a table', () => {
+    clusterSnippets.createBackup(INSTANCE_ID, CLUSTER_ID, TABLE_ID, BACKUP_ID);
```
JustinBeckwith, line 81: Are these all async functions?  Do we need any assertions to make sure they ran successfully?  Is there a chance these tests are all skipped in the describe block?

JustinBeckwith (changes requested):

package.json
zamnuts: > `@google-cloud/paginator` was explicitly removed here:
> https://github.com/googleapis/nodejs-bigtable/commit/8a2a74e00c1241c89bf344db0b82440577270223
> 
> @AVaksman want to work with @zamnuts on what we're supposed to be doing here now?

I noticed that commit. I've successfully refactored the stream portion, but am still having trouble understanding the pagination.

samples/document-snippets/cluster.js
zamnuts: Thanks for reviewing the samples. There were competing patterns, and I wasn't sure what to implement. I'll revise based on your feedback.

package.json
AVaksman: Sure,

- add gapic method name (`listBackupsStream: true`) to [`gapicStreamingMethods`](https://github.com/googleapis/nodejs-bigtable/blob/2917d177fe2a69bb7cd0f652bd3318aee359dd4d/src/index.ts#L753)
- pipe, rather `pumpify` received objects through a `Transform` stream that will convert proto objects into this library `Backup` objects. Similar reference below.

https://github.com/googleapis/nodejs-bigtable/blob/2917d177fe2a69bb7cd0f652bd3318aee359dd4d/src/instance.ts#L727-L744

src/backup.ts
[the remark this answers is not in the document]
AVaksman: You should keep the overloads as is
```ts
delete(gaxOptions?: CallOptions): Promise<DeleteBackupResponse>;
delete(callback: DeleteBackupCallback): void;
delete(gaxOptions: CallOptions, callback: DeleteBackupCallback): void;
```
and use `async` for the function implementation
```ts
async delete(
    optionsOrCallback?: CallOptions | DeleteBackupCallback
  ): void | Promise<DeleteIBackupResponse> {
```
Provide callback support via `callbackifyAll`

AVaksman (changes requested):

src/instance.ts
```diff
@@ -139,6 +139,23 @@ export type SetInstanceMetadataCallback = (
...
+  gaxOptions?: CallOptions;
+}
+export type RestoreTableCallback = (
```
kolea2, line 145: should these types be on table rather than instance?

kolea2:
@zamnuts can you take a look at the build failures? Thanks!

zamnuts:
> @zamnuts can you take a look at the build failures? Thanks!

Build failures after due to automated tests not being update to reflect the refactored implementation.

src/table.ts
```diff
@@ -482,6 +487,76 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+   * @param {BackupTimestamp} fields.expireTime When the backup will be
+   *   automatically deleted.
+   * @param {CallOptions | CreateBackupCallback} [gaxOptionsOrCallback]
+   * @param {CreateBackupCallback} [cb]
+   * @return {void | Promise<CreateBackupResponse>}
+   *
+   * @example <caption>include:samples/document-snippets/table.js</caption>
+   * region_tag:bigtable_create_table
+   */
+  backup(
```
AVaksman, line 524: Not sure about this convenience function, as it internally  sends additional RPC.

AVaksman (changes requested):

src/cluster.ts
```diff
@@ -435,8 +452,122 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
+    delete (reqOpts as GetBackupsOptions).gaxOptions;
+
+    this.bigtable.request<google.bigtable.admin.v2.IBackup[]>(
+      {
+        client: 'BigtableTableAdminClient',
+        method: 'listBackups',
+        reqOpts,
+        gaxOpts: options.gaxOptions,
+      },
+      (err, resp) => {
```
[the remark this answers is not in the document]
stephenplusplus: @AVaksman looking at `instance.getAppProfiles()` and `instance.getClusters()`, they follow the pattern @zamnuts used here. Do we need to fix those functions as well?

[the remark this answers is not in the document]
AVaksman: Yes for `getAppProfiles`, the changes are already included in existing PR #731
`instance.getAppProfiles()` - [return signature](https://github.com/AVaksman/nodejs-bigtable/blob/67d17e30c7fbc53377ccb5453b6f6afe3375d9fe/src/app-profile.ts#L105-L114), method [implementation](https://github.com/AVaksman/nodejs-bigtable/blob/67d17e30c7fbc53377ccb5453b6f6afe3375d9fe/src/instance.ts#L744-L759)
`instance.getClusters()` is different, it does not support pagination anymore. `pageToken` in the request is [deprecated](https://cloud.google.com/bigtable/docs/reference/admin/rpc/google.bigtable.admin.v2#google.bigtable.admin.v2.ListClustersRequest) and `next_page_token` is [deprecated](https://cloud.google.com/bigtable/docs/reference/admin/rpc/google.bigtable.admin.v2#google.bigtable.admin.v2.ListClustersResponse) from the response.

Issue #708, opened 2020-04-15
Refactor streaming of table rows and remove through2 dependency


crwilcox:
The code for table.ts is a bit complicated and can likely be streamlined. In order to retry, we nest two streams.

It is likely worth removing the two instances of through and subclassing stream. This should help to shutdown on termination and not attempt writes past that point.

This also will remove dependencies on `through2` and `readable-stream`

crwilcox:
#770 removed the remaining usage. 🎉

PR #770, opened 2020-07-20
feat: remove through2 dependency

laljikanjareeya:
Towards #708

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -809,17 +809,19 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
 
       requestStream!.on('request', () => numRequestsMade++);
 
-      const transform = new Transform({objectMode: true});
-      transform._transform = (rowData, _, next) => {
-        if (chunkTransformer._destroyed || !userStream.writable) {
-          return next();
-        }
-        numRequestsMade = 0;
-        rowsRead++;
-        const row = this.row(rowData.key);
-        row.data = rowData.data;
-        next(null, row);
-      };
+      const transform = new Transform({
+        transform: (rowData, _, next) => {
+          if (chunkTransformer._destroyed || !userStream.writable) {
+            return next();
+          }
+          numRequestsMade = 0;
+          rowsRead++;
+          const row = this.row(rowData.key);
+          row.data = rowData.data;
+          next(null, row);
+        },
+        objectMode: true,
+      });
 
       rowStream = pumpify.obj([requestStream, chunkTransformer, transform]);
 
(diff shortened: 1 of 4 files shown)


src/table.ts
```diff
@@ -810,24 +809,21 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
-      rowStream = pumpify.obj([
-        requestStream,
-        chunkTransformer,
-        through.obj((rowData, enc, next) => {
-          if (
-            chunkTransformer._destroyed ||
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            (userStream as any)._writableState.ended
-          ) {
+      const transform = new Transform({
```
stephenplusplus, line 812: Can we call this "'toRowStream" or something similar to indicate what it's doing?

src/table.ts
```diff
@@ -1561,19 +1557,24 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       appProfileId: this.bigtable.appProfileId,
     };
 
+    const transform = new Transform({
```
stephenplusplus, line 1560: Is there a more descriptive name we can use for this stream as well?

system-test/read-rows-acceptance-tests.ts
```diff
@@ -69,31 +71,29 @@ describe('Read Row Acceptance tests', () => {
...
+            );
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            let readRowsResponse: any = {chunks: [cellChunk]};
+            readRowsResponse = ReadRowsResponse.create(readRowsResponse);
+            readRowsResponse = ReadRowsResponse.toObject(readRowsResponse, {
+              defaults: true,
+              longs: String,
+              oneofs: true,
+            });
+            stream.emit('data', readRowsResponse);
```
stephenplusplus, line 91: `stream.push()` won't work anymore?

stephenplusplus (changes requested):

src/table.ts
laljikanjareeya: Done!

system-test/read-rows-acceptance-tests.ts
laljikanjareeya: yes that works fine.

laljikanjareeya:
@stephenplusplus Thanks for the review.

Two tests were failing:
1. invalid - duplicate row key
2. invalid - new row missing row key

I have also cross verify it with `through2` version `4.0.2` these tests are also failing as they have removed `destroy()`.
Based on that I just change the test cases.

stephenplusplus (approved):

PR #781, opened 2020-08-03
fix: respect pageSize and pageToken from gaxOptions

AVaksman:

#### Changes
- Expose `pageSize` and `pageToken` in `getTablesOptions`
- Respect `pageSize` and `pageToken` passed from `gaxOptions`
- Values passed in `options` take precedence vs values from `gaxOptions`

### DIFF
--- a/src/instance.ts
+++ b/src/instance.ts
@@ -16,6 +16,7 @@ import {promisifyAll} from '@google-cloud/promisify';
 import {Transform} from 'stream';
 import arrify = require('arrify');
 import * as is from 'is';
+import * as extend from 'extend';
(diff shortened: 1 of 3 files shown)


bcoe (approved):
I would lean towards calling this a `fix:`. 

Even though this is technically a breaking change, I think given the prior behavior was obviously a bug, I'm comfortable landing as patch.

Would be curious if any other folks have thoughts though.

AVaksman:
It shouldn't break an existing code, I believe.

bcoe:
@AVaksman thank you 👌

PR #794, opened 2020-09-02
Introduce Backups

stephenplusplus:
#### Access or create Backup objects through a Cluster reference
```js
cluster.backup(id) // references a pre-existing "Backup" object (no remote calls)
cluster.createBackup({ id: 'backup-id', table: 'table-id' }) // creates and returns a "Backup" object
cluster.getBackups() // returns "Backup" objects
cluster.getBackupsStream() // returns "Backup" objects as a readable stream
```

#### Access Backup objects through an Instance reference
```js
instance.createTableFromBackup({ table: 'table-id', backup: 'backup-id' }, gaxOptions)
instance.getBackups() // returns "Backup" objects
instance.getBackupsStream() // returns "Backup" objects as a readable stream
```

#### Create a Backup through a Table reference
```js
table.createBackup(id, metadata) // creates and returns a "Backup" object
```

#### Once you have a reference to a Backup object:
*All methods below besides "restore()" are common CRUD methods defined throughout our API*
```js
backup.create({ table: 'table-id', metadata: {...} })
backup.delete()
backup.get()
backup.getMetadata()
backup.restore('table-id')
backup.setMetadata({...})
```

#### References

- [Original implementation](https://github.com/googleapis/nodejs-bigtable/pull/761)
- [Proto file](https://github.com/googleapis/googleapis/blob/master/google/bigtable/admin/v2/bigtable_table_admin.proto)

## To Dos

- [x] Source code
- [x] TypeScript annotations
- [x] JSDoc parameters
- [x] Unit tests
  - [x] Backup
  - [x] Cluster
  - [x] Index
  - [x] Instance
  - [x] Table
- [x] System tests

This PR splits the work from #761 into three parts-- source code, tests, and then documentation and samples to come in a follow-up PR. This should make reviewing easier, as well as lets us lock in the source code changes before attempting to adapt the previously written tests and samples.

Currently, this PR has only the source code. Most of it was carried from #761, but there were some structural changes to follow our conventions:

```diff
- Backup(id, cluster, fields)
+ Backup(cluster, id)

get backup.expireDate()
get backup.startDate()
get backup.endDate()

- backup.create(table, metadata)
+ backup.create({table, metadata})
backup.delete()
backup.get()
+ backup.getMetadata()
backup.restore(tableId, gaxOptions)
- backup.update(metadata, gaxOptions)
+ backup.setMetadata(metadata, gaxOptions)

+ cluster.backup(id)
+ cluster.createBackup({id, table}, gaxOptions)
cluster.getBackups(options)
cluster.getBackupsStream(options)

- instance.backup(id, cluster, fields)
- instance.createTableFromBackup(tableId, backup, gaxOptions)
+ instance.createTableFromBackup({table, backup}, gaxOptions)
+ instance.getBackups(options, callback)
+ instance.getBackupsStream(options, callback)

- table.backup(id, metadata, gaxOptions)
+ table.createBackup(id, metadata, gaxOptions)
```

@AVaksman could you please review this when you get a chance? I will likely need special attention to the TypeScript, making sure I'm properly compliant with our conventions and use cases I may have missed. Thank you!


src/cluster.ts
```diff
@@ -437,6 +655,12 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
   }
 }
 
+/*! Developer Documentation
+ *
+ * These methods can be auto-paginated.
+ */
+paginator.extend(Cluster, ['getBackups']);
```
AVaksman, line 662: I believe this is no longer needed

src/instance.ts
```diff
@@ -1084,6 +1092,63 @@ Please use the format 'my-instance' or '${bigtable.projectName}/instances/my-ins
...
+   * to track the progress of the operation, and to cancel it.
+   *
+   * @param {CreateTableFromBackupConfig} config Configuration object.
+   * @param {string} config.tableId The id of the table to create and restore
+   *     to.
+   * @param {Backup | string} config.backup The name of the backup from which to
+   *     restore of the form
+   *     `projects/<project>/instances/<instance>/clusters/<cluster>/backups/<backup>`,
+   *     or a Backup instance.
+   * @param {CallOptions | RestoreTableCallback} [config.gaxOptions] Request
```
AVaksman, line 1117:
```suggestion
   * @param {CallOptions} [config.gaxOptions] Request
```

src/instance.ts
```diff
@@ -1084,6 +1092,63 @@ Please use the format 'my-instance' or '${bigtable.projectName}/instances/my-ins
...
+   */
+  createTableFromBackup(
+    config: CreateTableFromBackupConfig,
+    callback?: RestoreTableCallback
+  ): void | Promise<RestoreTableResponse> {
+    if (!config.table) {
+      throw new Error('An table id is required to restore from a backup.');
+    }
+
+    const tableAdminClient = this.bigtable.api[
```
AVaksman, line 1131: `this.bigtable.api` may be still empty

src/instance.ts
```diff
@@ -1084,6 +1092,63 @@ Please use the format 'my-instance' or '${bigtable.projectName}/instances/my-ins
...
+      'BigtableTableAdminClient'
+    ] as BigtableTableAdminClient;
+
+    let backup: Backup;
+
+    if (config.backup instanceof Backup) {
+      backup = config.backup;
+    } else {
+      const clusterId = tableAdminClient
```
AVaksman, line 1140: `tableAdminClient` may be `undefined`

src/cluster.ts
```diff
@@ -321,6 +429,116 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
+   * //-
+   * // If you anticipate many results, you can end a stream early to prevent
+   * // unnecessary processing and API requests.
+   * //-
+   * cluster.getBackupsStream()
+   *   .on('data', function(backup) {
+   *     this.end();
+   *   });
+   */
+  getBackupsStream(options: GetBackupsOptions): NodeJS.ReadableStream {
```
AVaksman, line 517: For this method to work `listBackupsStream: true` needs to be added to [`gapicStreamingMethods`](https://github.com/googleapis/nodejs-bigtable/blob/2a199aae5646fe4e19836e13a61fcb83b527e4c5/src/index.ts#L753)

AVaksman (changes requested):

src/backup.ts
```diff
@@ -0,0 +1,505 @@
...
+        google.bigtable.admin.v2.ITable,
+        google.bigtable.admin.v2.IRestoreTableMetadata
+      >
+    >(
+      {
+        client: 'BigtableTableAdminClient',
+        method: 'restoreTable',
+        reqOpts: {
+          parent: this.cluster.name,
+          tableId: table.name,
```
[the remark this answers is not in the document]
stephenplusplus: Woops, thanks!

src/backup.ts
```diff
@@ -0,0 +1,505 @@
...
+    gaxOptionsOrCallback?: CallOptions | GetMetadataCallback,
+    cb?: GetMetadataCallback
+  ): void | Promise<GetMetadataResponse> {
+    const gaxOpts =
+      typeof gaxOptionsOrCallback === 'object' ? gaxOptionsOrCallback : {};
+    const callback =
+      typeof gaxOptionsOrCallback === 'function' ? gaxOptionsOrCallback : cb!;
+
+    // @TODO `this.metadata` isn't a `Backup`. Figure out why it has to be cast.
+    this.get(gaxOpts, err => callback(err, this.metadata as Backup));
```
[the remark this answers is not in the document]
stephenplusplus: I didn't see the suggestion about moving the function body from `getMetadata()` to `get()`. I'm not opposed to it, but why is one better than the other? (I've since made the other corrections to these functions, and they all seem to be co-existing happily)

stephenplusplus:
@AVaksman thank you very much for the thorough review! I've made the changes (thank you for all of the suggestions) and left just one thing: https://github.com/googleapis/nodejs-bigtable/pull/794#discussion_r482692426.

src/table.ts
```diff
@@ -512,6 +521,74 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+  createBackup(
+    id: string,
+    config: CreateBackupConfig,
+    callback?: CreateBackupCallback
+  ): void | Promise<CreateBackupResponse> {
+    if (!id) {
+      throw new TypeError('An id is required to create a backup.');
+    }
+
+    this.getReplicationStates(config.gaxOptions)
```
JustinBeckwith, line 569: The mixing of promises and callbacks is hurting my head a little here.  Is it like this because `getReplicationStates` doesn't support callbacks?

src/cluster.ts
```diff
@@ -321,6 +429,119 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
+
+    this.bigtable.request<google.bigtable.admin.v2.IBackup[]>(
+      {
+        client: 'BigtableTableAdminClient',
+        method: 'listBackups',
+        reqOpts,
+        gaxOpts: options.gaxOptions,
+      },
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (err, ...resp: any[]) => {
```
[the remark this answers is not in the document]
stephenplusplus: @AVaksman could you advise here? This was taken from your suggestion on the other PR. I'm fighting a lot of TypeScript errors when I tamper with this.

stephenplusplus:
Looks like you can't delete old @google-cla posts-- it'll just post it again 🤣

src/cluster.ts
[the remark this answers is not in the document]
AVaksman: I took a stab at it as a part of an existing PR #731 where I am differentiating `bigtable.request` signature between `NormalCallback`, `LROCallback` and `PaginatedCallback`
```ts
export type RequestCallback<T = void, R = void> = T extends void
  ? LROCallback
  : R extends void
  ? NormalCallback<T>
  : PagedCallback<T, R>;

export interface NormalCallback<TResponse> {
  (err: ServiceError | null, res?: TResponse): void;
}

export interface PagedCallback<Item, Response> {
  (
    err: ServiceError | null,
    results?: Item[] | null,
    nextQuery?: {} | null,
    response?: Response | null
  ): void;
}

export interface LROCallback {
  (
    err: ServiceError | null,
    operation?: Operation,
    apiResponse?: IOperation
  ): void;
}
```

Then here the something like this would work
```ts
this.bigtable.request<
      google.bigtable.admin.v2.IBackup,
      google.bigtable.admin.v2.IListBackupsResponse
    >(
      {
        client: 'BigtableInstanceAdminClient',
        method: 'listAppProfiles',
        reqOpts,
        gaxOpts,
      },
      (err, IBackups, nextPageRequest, apiResponse) => {
        if (err) {
          callback(err);
          return;
        }
        const backups = IBackups!.map(backup => {
          const backupInstance = this.backup(backup.name!.split('/').pop()!);
          backupInstance.metadata = backup;
          return backupInstance;
        });
        const nextQuery = nextPageRequest!
          ? extend({}, options, nextPageRequest!)
          : null;
        callback(null, appProfiles, nextQuery, apiResponse!);
      }
    );
```
Is this an overkill?
`bigtable.request` signature changes could be done in a separate PR in a non breaking way.
WDYT?

AVaksman:
I've seen commenting "@googlebot I consent." have helped @stephenplusplus

src/cluster.ts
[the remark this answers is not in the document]
AVaksman: If agree, I can do the `bigtable.request` PR?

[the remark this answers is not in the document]
stephenplusplus: Curious for @JustinBeckwith's thoughts. Whatever makes sense to you guys :)

For now, should I leave the code as-is? Also, please LMK if everything else looks correct so far and I'll get the tests going.

src/backup.ts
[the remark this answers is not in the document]
AVaksman: Perhaps didn't click "Add review comment"
My suggestion was to move the function body from `get()` to `getMetadata()` as the snippet above.
Then in `get()` do 
```ts
    this.getMetadata(
      gaxOptions,
      (err, metadata) => {
        callback(err, err ? null : this, metadata);
      }
    );
```
Following an existing example in [Cluster](https://github.com/googleapis/nodejs-bigtable/blob/cb77d569551bab8dd2897c31ea5a47aa0ae99706/src/cluster.ts#L305-L322), [Instance](https://github.com/googleapis/nodejs-bigtable/blob/cb77d569551bab8dd2897c31ea5a47aa0ae99706/src/instance.ts#L620-L635)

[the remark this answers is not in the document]
stephenplusplus: Gotcha, thanks!

src/backup.ts
```diff
@@ -0,0 +1,516 @@
...
+    metadata: ModifiableBackupFields,
+    gaxOptionsOrCallback?: CallOptions | SetMetadataCallback,
+    cb?: SetMetadataCallback
+  ): void | Promise<SetMetadataResponse> {
+    const gaxOpts =
+      typeof gaxOptionsOrCallback === 'object' ? gaxOptionsOrCallback : {};
+    const callback =
+      typeof gaxOptionsOrCallback === 'function' ? gaxOptionsOrCallback : cb!;
+
+    const {expireTime, ...restMetadata} = metadata;
```
AVaksman, line 458: As of right now the only updatable field is `expireTime`.
In general, do we want to require that "at least one field value should be provided" and throw if none provided?

src/cluster.ts
```diff
@@ -201,6 +230,85 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
+   * @param {object} config Configuration object.
+   * @param {BackupTimestamp} config.expireTime When the backup will be
+   *     automatically deleted.
+   * @param {string|Table} config.table Table to create the backup from.
+   * @param {CallOptions} [config.gaxOptions] Request configuration options,
+   *     outlined here:
+   *     https://googleapis.github.io/gax-nodejs/CallSettings.html.
+   * @param {CreateBackupCallback} [callback] The callback function.
+   * @param {?error} callback.err An error returned while making this request.
+   * @param {Cluster} callback.backup The newly created Backup.
```
AVaksman, line 255:
```suggestion
   * @param {Backup} callback.backup The newly created Backup.
```

src/cluster.ts
```diff
@@ -171,6 +190,16 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
     return storageTypes[type] || storageTypes.unspecified;
   }
 
+  /**
+   * Get a reference to a Bigtable Cluster.
+   *
+   * @param {string} id The backup name or id.
+   * @returns {Backup}
+   */
+  backup(id: string): Backup {
```
AVaksman, line 199: `'backup'` should be added to promisifyAll exclude.

src/instance.ts
```diff
@@ -1084,6 +1091,58 @@ Please use the format 'my-instance' or '${bigtable.projectName}/instances/my-ins
...
+    let backup: Backup;
+
+    if (config.backup instanceof Backup) {
+      backup = config.backup;
+    } else {
+      try {
+        const clusterId = config.backup.match(/clusters\/([^/]+)/)![1];
+        backup = this.cluster(clusterId).backup(config.backup);
+      } catch (e) {
+        throw new Error('A complete backup name (path) is required.');
```
AVaksman, line 1139: Or a backup object reference

src/instance.ts
```diff
@@ -140,6 +141,12 @@ export type SetInstanceMetadataCallback = (
 ) => void;
 export type SetInstanceMetadataResponse = [google.protobuf.Empty];
 
+export interface CreateTableFromBackupConfig {
+  table: string;
```
AVaksman, line 145: Should we split `tableId` param from the rest of the config values to have similar feel as `Instance.createTable()`?

src/instance.ts
```diff
@@ -1084,6 +1091,58 @@ Please use the format 'my-instance' or '${bigtable.projectName}/instances/my-ins
     ]);
   }
 
+  createTableFromBackup(
+    config: CreateTableFromBackupConfig
```
AVaksman, line 1095: Question about passing `tableId` separately from the rest of the config to have a similar feel as `Instance.createTable()`

src/table.ts
```diff
@@ -512,6 +521,80 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+    config: CreateBackupConfig,
+    callback: CreateBackupCallback
+  ): void;
+  /**
+   * Backup a table with cluster auto selection.
+   *
+   * Backups of tables originate from a specific cluster. This is a helper
+   * around `Cluster.createBackup` that automatically selects the first ready
+   * cluster from which a backup can be performed.
+   *
```
AVaksman, line 544: Should probably warn that this method uses another RPC?

AVaksman (changes requested):

src/backup.ts
```diff
@@ -0,0 +1,516 @@
...
+export type GetMetadataCallback = (
+  err?: ServiceError | null,
+  metadata?: IBackup | null
+) => void;
+export type GetMetadataResponse = [IBackup];
+
+export type SetMetadataCallback = (
+  err: ServiceError | null,
+  metadata: IBackup,
+  resp: IBackup
```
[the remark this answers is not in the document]
stephenplusplus: I tried to copy what ServiceObject#setMetadata does: https://github.com/googleapis/nodejs-common/blob/981a17650007d5c2a8b30051909baf2cc795d05d/src/service-object.ts#L510:

```js
callback!(err, this.metadata, res);
```

src/instance.ts
stephenplusplus: That is actually how it used to be, and I changed it to the config object. I didn't think about matching "createTable(tableId)", although that makes sense. However, I still think it's best to do a configuration object, as my reasoning was it's not intuitive from the method name which argument should come first, "table" or "backup". Each would be logical and could trip users up:

```js
await createTableFromBackup('my-backup', 'my-table') // :(
await createTableFromBackup('my-table', 'my-backup') // :(
await createTableFromBackup({table: 'my-table', backup: 'my-backup'}) // :)
```

LMK if you still think we should switch!

src/cluster.ts
```diff
@@ -321,6 +429,119 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
+  ): void | Promise<GetBackupsResponse> {
+    const options =
+      typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+    const callback =
+      typeof optionsOrCallback === 'function' ? optionsOrCallback : cb!;
+
+    const reqOpts: google.bigtable.admin.v2.IListBackupsRequest = {
+      ...options,
+      parent: this.name,
+    };
```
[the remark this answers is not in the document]
stephenplusplus: Done! Although, please take a look as I had to make some changes to the way you had it written.

src/backup.ts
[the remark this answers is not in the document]
AVaksman: Actually all other setMetadata functions return LROs.
So no preference here.

src/instance.ts
AVaksman: Let's have the config object 👍

src/backup.ts
stephenplusplus: I don't feel too strongly, although I don't think we have a consistent precedent for that. I lean towards letting the API fail if it has an issue, when we don't have any code that requires it be defined (e.g. if we did "expireTime.toStruct()" without making sure if it works first).

stephenplusplus:
@AVaksman I believe I have completed the unit tests, although I'm always left wondering if there's one too many "any"s, or if I flubbed some other TypeScript best practice. If you could take a look, I would appreciate any help! (The 'test/backup.ts' file is especially any-heavy)

test/backup.ts
```diff
@@ -0,0 +1,550 @@
...
+      backup.restore(tableId, assert.ifError);
+    });
+
+    it('should accept gaxOptions', done => {
+      const tableId = 'table-id';
+      const gaxOptions = {};
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (backup as any).bigtable.request = (config: any) => {
+        assert.deepStrictEqual(config.gaxOpts, gaxOptions);
```
AVaksman, line 413:
```suggestion
        assert.strictEqual(config.gaxOpts, gaxOptions);
```

test/backup.ts
```diff
@@ -0,0 +1,550 @@
...
+  },
+});
+
+describe.only('Bigtable/Backup', () => {
+  const BACKUP_ID = 'my-backup';
+  let CLUSTER: clusterTypes.Cluster;
+  let BACKUP_NAME: string;
+
+  let Backup: typeof backupTypes.Backup;
+  let backup: backupTypes.Backup;
```
AVaksman, line 49: It seems if declare `backup` type `any` here, most subsequent casts to any (`(backup as any).someMethod = ...`) could be avoided?

AVaksman (changes requested):

AVaksman:
Are we interested in providing `instance#getBackups()` function to list all backups belonging to this instance across all clusters?

By providing `reqOpts.parent` in the form of `'projects/{project}/instances/{instance}/clusters/-'` 

WDYT?

test/backup.ts
stephenplusplus: Nice, thanks! (Woops)

test/table.ts
```diff
@@ -234,6 +234,135 @@ describe('Bigtable/Table', () => {
...
+      });
+
+      table.instance.cluster = (id: string) => {
+        assert.strictEqual(id, readyClusterId);
+        setImmediate(done);
+        return {createBackup: () => {}};
+      };
+
+      table.getReplicationStates = (gaxOptions: {}, callback: Function) => {
+        callback(null, REPLICATION_STATES);
```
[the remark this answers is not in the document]
stephenplusplus: Nice catch!

stephenplusplus:
> Are we interested in providing `instance#getBackups()` function to list all backups belonging to this instance across all clusters?

Great idea, I missed that was possible. I'll add that in now.

Thank you for the review. I've addressed the other comments and added system tests.

stephenplusplus:
`Instance#getBackups()` and `Instance#getBackupsStream()` added.

system-test/bigtable.ts
```diff
@@ -1111,8 +1132,148 @@ describe('Bigtable', () => {
...
+      await BACKUP.getMetadata();
+
+      assert.strictEqual(
+        BACKUP.metadata!.name,
+        replaceProjectId(bigtable, backupNameFromCluster)
+      );
+
+      const expectedTime = expireTime.toStruct();
+      assert.deepStrictEqual(
+        BACKUP.metadata?.expireTime?.seconds?.toString(),
```
AVaksman, line 1179: Do you want to use `BACKUP#expireDate()` here and in other tests?

system-test/bigtable.ts
```diff
@@ -35,20 +36,39 @@ describe('Bigtable', () => {
...
-        const oneHourAgo = new Date(Date.now() - 3600000);
-        return !timeCreated || timeCreated <= oneHourAgo;
-      });
+    const testInstances = instances.filter(i => i.id.match(PREFIX));
+    // .filter(i => {
+    //   const timeCreated = (i.metadata!.labels!.time_created as {}) as Date;
+    //   // Only delete stale resources.
+    //   const oneHourAgo = new Date(Date.now() - 3600000);
+    //   return !timeCreated || timeCreated <= oneHourAgo;
+    // });
```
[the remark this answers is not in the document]
stephenplusplus: Yes, oops. Didn’t mean to leave this in.

stephenplusplus:
@AVaksman thanks again. Good call using the new Instance#getBackups() method from the tests. Caught an issue where I wasn't replacing the "-" cluster ID with the actual one returned from the API.

stephenplusplus:
@AVaksman if all source and test code looks good, we can merge this in, then I'll be taking on the samples in a follow-up PR.

system-test/bigtable.ts
```diff
@@ -1111,8 +1130,152 @@ describe('Bigtable', () => {
...
+  describe('backups', () => {
+    const CLUSTER = INSTANCE.cluster(CLUSTER_ID);
+    let BACKUP: Backup;
+
+    // For these tests, two backups are needed. The backups are labeled for what
+    // they are intended to originate/interact from/with, but this is just for
+    // testing and the naming convention used here does not actually influence
+    // the real functionality - it is just a way to keep things organized!
+    const backupIdFromCluster = generateId('backup');
+    const backupNameFromCluster = `${CLUSTER.name}/backups/${backupIdFromCluster}`;
```
AVaksman, line 1143: Does it pay to use `replaceProjectId` or [`replaceProjectIdToken`](https://github.com/stephenplusplus/nodejs-bigtable/blob/spp--backups/src/index.ts#L15) here and a few lines below for backupNameFromTable?

AVaksman:
> if all source and test code looks good, we can merge this in, then I'll be taking on the samples in a follow-up PR.

Left a couple of suggestions. Can you tale a look? WDYT?

src/instance.ts
```diff
@@ -827,12 +827,30 @@ Please use the format 'my-instance' or '${bigtable.projectName}/instances/my-ins
...
+        return;
+      }
+
+      const backupsFromCorrectCluster = backups?.map(backup => {
+        const matchedClusterId = backup.metadata?.name?.match(
+          /clusters\/([^/]+)/
+        )![1];
+        const backupInstance = this.cluster(matchedClusterId!).backup(
+          backup.id
+        );
```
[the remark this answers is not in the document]
AVaksman: Or just 
```ts
this.instance.cluster(backup.name?.match(/clusters\/([^/]+)/)![1]!)
              .backup(backup.name!.split('/').pop()!)
```
without condition

stephenplusplus:
@AVaksman thank you again. Suggestions looked good to me 👍🏻

src/cluster.ts
```diff
@@ -504,6 +504,7 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
         if (resp[0]) {
           backups = resp[0].map((backup: IBackup) => {
             const backupInstance = this.backup(backup.name!.split('/').pop()!);
+            console.log('instance name when created', backupInstance.name);
```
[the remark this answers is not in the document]
stephenplusplus: 🤦🏻

AVaksman (approved):

AVaksman:
I think it is good to go!

src/instance.ts
```diff
@@ -29,6 +29,11 @@ import {
   GetAppProfilesCallback,
   GetAppProfilesResponse,
 } from './app-profile';
+import {
```
kolea2, line 32: how come this is not on `table.ts`? In other clients, all backup operations are on the table admin client. ex: https://github.com/liubonan/java-bigtable/blob/master/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClient.java#L805. Let me know if my understanding is incorrect :)

AVaksman:
Might need a couple of unit tests to keep codecov happy

src/instance.ts
stephenplusplus: Hello! We ended up with these methods defined throughout various classes:

#### Access or create Backup objects through a Cluster reference
```js
cluster.backup(id) // references a pre-existing "Backup" object (no remote calls)
cluster.createBackup({ id: 'backup-id', table: 'table-id' }) // creates and returns a "Backup" object
cluster.getBackups() // returns "Backup" objects
cluster.getBackupsStream() // returns "Backup" objects as a readable stream
```

#### Access Backup objects through an Instance reference
```js
instance.createTableFromBackup({ table: 'table-id', backup: 'backup-id' }, gaxOptions)
instance.getBackups() // returns "Backup" objects
instance.getBackupsStream() // returns "Backup" objects as a readable stream
```

#### Create a Backup through a Table reference
```js
table.createBackup(id, metadata) // creates and returns a "Backup" object
```

#### Once you have a reference to a Backup object:
*All methods below besides "restore()" are common CRUD methods defined throughout our API*
```js
backup.create({ table: 'table-id', metadata: {...} })
backup.delete()
backup.get()
backup.getMetadata()
backup.restore('table-id')
backup.setMetadata({...})
```

src/cluster.ts
```diff
@@ -201,6 +239,91 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
+    const reqOpts = {
+      parent: this.name,
+      backupId: id,
+      backup: {
+        sourceTable: typeof table === 'string' ? table : table.name,
+        ...config,
+      },
+    };
+
+    if (reqOpts.backup.expireTime instanceof Date) {
```
JustinBeckwith, line 298: There used to be all kinds of nuance on when it was ok to use `instanceof Date`, and you had to do weird tricks to detect date objects.  It looks like `is` now does exactly what you're doing here though:
https://github.com/sindresorhus/is/blob/master/source/index.ts

So I'm good with it :)

src/cluster.ts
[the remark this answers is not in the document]
JustinBeckwith: I am ok with leaving this as is for now, and coming back around to improve the under types here later.  I'm ok with it because I don't think they affect the public types.  Not a big deal.

src/cluster.ts
```diff
@@ -321,6 +444,150 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
+  getBackups(
+    optionsOrCallback?: GetBackupsOptions | GetBackupsCallback,
+    cb?: GetBackupsCallback
+  ): void | Promise<GetBackupsResponse> {
+    let options =
+      typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+    const callback =
+      typeof optionsOrCallback === 'function' ? optionsOrCallback : cb!;
+
+    options = extend(true, {}, options);
```
JustinBeckwith, line 470: I spent a year trying to remove `extend` in favor of `Object.assign`, and I just want to say out loud that @stephenplusplus was right all along and I was wrong.

JustinBeckwith (approved):

src/cluster.ts
stephenplusplus: Hahaha, stdlib got so close on this one!

src/backup.ts
```diff
@@ -0,0 +1,542 @@
...
+   * number, or a boolean. The comparison operator must be <, >, <=, >=, !=, =,
+   * or :. Colon ‘:’ represents a HAS operator which is roughly synonymous with
+   * equality. Filter rules are case insensitive.
+   */
+  filter?: string;
+
+  /**
+   * An expression for specifying the sort order of the results of the request.
+   * The string value should specify one or more fields in
+   * {@link google.bigtable.admin.v2.Backup|Backup}. The full syntax is
```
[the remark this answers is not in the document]
stephenplusplus: Hmm, well I don't actually know where that gets rendered. That's part of `interface GetBackupsOptions`, and I can't find where it went. It's not linked to where we use it from the `getBackups()` methods :\. All of the other links from descriptions seem to work, though!

kolea2 (approved):

PR #795, opened 2020-09-03
feat: getAppProfilesStream emits (if there are any failed locations present) decorated error with failedLocations info

AVaksman:

### DIFF
--- a/src/instance.ts
+++ b/src/instance.ts
@@ -779,7 +779,7 @@ Please use the format 'my-instance' or '${bigtable.projectName}/instances/my-ins
       gaxOpts,
     });
     stream.on('response', apiResp => {
-      if (apiResp.failedLocations.length > 0) {
+      if (arrify(apiResp.failedLocations).length > 0) {
         failedLocations = failedLocations.concat(apiResp.failedLocations);
       }
     });
--- a/src/instance.ts
+++ b/src/instance.ts
@@ -758,14 +758,38 @@ Please use the format 'my-instance' or '${bigtable.projectName}/instances/my-ins
       appProfile.metadata = chunk;
       callback(null, appProfile);
     };
+    let failedLocations: string[] = [];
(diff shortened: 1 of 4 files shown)


src/instance.ts
```diff
@@ -758,14 +758,38 @@ Please use the format 'my-instance' or '${bigtable.projectName}/instances/my-ins
...
+      }
+    };
+    const stream = this.bigtable.request({
+      client: 'BigtableInstanceAdminClient',
+      method: 'listAppProfilesStream',
+      reqOpts,
+      gaxOpts,
+    });
+    stream.on('response', apiResp => {
+      if (apiResp.failedLocations.length > 0) {
```
stephenplusplus, line 782: Will `failedLocations` always be defined as an array on the API response?

AVaksman: As per the documentation [here](https://cloud.google.com/bigtable/docs/reference/admin/rpc/google.bigtable.admin.v2#listappprofilesresponse) `failed_locations` property is marked as an array.
Should I `arrify` it just in case?

stephenplusplus: Hmm, wouldn't hurt. Does it make sense to check a real response real quick to see what happens when there aren't any failures?

AVaksman: Is this what you are asking for?
image

stephenplusplus: I was thinking of actually making the request with the code and seeing what comes back. But if you’re confident from the types, don’t worry about it.

AVaksman: The image is from a breakpoint of an actual [system-test](https://github.com/googleapis/nodejs-bigtable/blob/master/system-test/bigtable.ts#L166), this is how a response without `faledLocations` looked like.

stephenplusplus: Oh, sorry. Fancy! Thanks :)

stephenplusplus (approved):

PR #799, opened 2020-09-14
feat: add backup level IAM policy controls

AVaksman:
The new features allow user to check and tune [IAM policy](https://cloud.google.com/iam/docs/reference/rest/v1/Policy) for the backup level:

- `getIamPolicy` - allows a user obtain the current resource IAM policy.
- `setIamPolicy` - allows a user to set resource level IAM policy.
- `testIamPermissions` - allows a user to pass a list of [`permissions`](https://cloud.google.com/bigtable/docs/access-control#permissions) and get back a sub-list of granted permissions.


system-test/bigtable.ts
```diff
@@ -1259,6 +1259,37 @@ describe('Bigtable', () => {
       assert.strictEqual(metadata.name, backupNameFromCluster);
       assert.deepStrictEqual(backup.expireDate, updateExpireTime);
     });
+
+    it.only('should get an Iam Policy for the backup', async () => {
```
stephenplusplus, line 1263: Yay, I'm not the `.only` one!

system-test/bigtable.ts
```diff
@@ -1259,6 +1259,37 @@ describe('Bigtable', () => {
...
+      const policyProperties = ['version', 'bindings', 'etag'];
+      const [policy] = await BACKUP.getIamPolicy();
+      console.log(`policy => \n${JSON.stringify(policy)}`);
```
stephenplusplus, line 1266: I don't think we should leave any logging in the tests.

system-test/bigtable.ts
```diff
@@ -1259,6 +1259,37 @@ describe('Bigtable', () => {
...
+
+      policyProperties.forEach(property => {
+        assert.strictEqual(Object.keys(policy).includes(property), true);
```
stephenplusplus, line 1269: I think we'd want to test that `policyProperties.length` is always > 0, or confirm the inner asserts from some of these tests will run.

system-test/bigtable.ts
```diff
@@ -1259,6 +1259,37 @@ describe('Bigtable', () => {
...
+
+    it.only('should set Iam Policy on the backup', async () => {
+      const backup = CLUSTER.backup(backupIdFromCluster);
+
+      const [policy] = await backup.getIamPolicy();
+      const [updatedPolicy] = await backup.setIamPolicy(policy);
+      console.log(`policy => \n${JSON.stringify(policy)}`);
+      console.log(`updatedPolicy => \n${JSON.stringify(updatedPolicy)}`);
+
+      assert.notStrictEqual(updatedPolicy, null);
```
stephenplusplus, line 1291: Can we instead confirm the type that it is?

test/backup.ts
```diff
@@ -328,6 +339,49 @@ describe('Bigtable/Backup', () => {
     });
   });
 
+  describe('getIamPolicy', () => {
+    it('should provide the proper request options', done => {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      backup.bigtable.request = (config: any) => {
```
stephenplusplus, line 345: I don't think we should do this extensive of testing. The method itself does not call this method or set the strings and values being tested here. Same for the error handling and callback executing, etc as those are behaviors of the Table class. Instead, we should stub the Table.prototype methods, and be sure they are called correctly.

stephenplusplus (changes requested):

system-test/bigtable.ts
AVaksman: How about 
```ts
      policyProperties.forEach(property => {
        assert(property in policy);
      });
```

stephenplusplus: Sorry, I misread the code. I thought `policyProperties` was part of the response from the API, not a locally-defined array. Either way looks fine to me!

system-test/bigtable.ts
AVaksman: Not really, other than the type is an `object`, since `Policy` is not a `class`.
How about
```ts
      Object.keys(policy).forEach(key => assert(key in updatedPolicy));
```

test/backup.ts
AVaksman: Done!
Thanks

system-test/bigtable.ts
AVaksman: Removed, thanks.

system-test/bigtable.ts
AVaksman: 🤣 
Removed!

system-test/bigtable.ts
stephenplusplus: I think that's fine, it seems better than just not null.

test/backup.ts
```diff
@@ -585,140 +568,75 @@ describe('Bigtable/Backup', () => {
...
+          assert.deepStrictEqual(gaxOpts, {});
+          callback(); // done()
+        });
+      backup.setIamPolicy(policy, done);
     });
 
     it('should accept gaxOptions', done => {
       const gaxOptions = {};
 
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
stephenplusplus, line 589: I think this line can go.

test/backup.ts
stephenplusplus: Nice, thank you!

test/backup.ts
AVaksman: Yep, missed that one.
Done

stephenplusplus (approved):

kolea2 (changes requested):
will follow up on this feature

src/backup.ts
```diff
@@ -16,7 +16,19 @@ import {PreciseDate} from '@google-cloud/precise-date';
...
+  GetIamPolicyCallback,
+  GetIamPolicyOptions,
+  GetIamPolicyResponse,
+  Policy,
+  SetIamPolicyCallback,
+  SetIamPolicyResponse,
+  TestIamPermissionsCallback,
+  TestIamPermissionsResponse,
+} from './';
+import {Table} from '../src/table';
```
crwilcox, line 31: Why are we importing Table from a different location now? I didn't notice any refactoring to move that around, but I may have missed it.

crwilcox (changes requested):

src/backup.ts
AVaksman: `Table.prototype.<method>` is used for `getIamPolicy`, `setIamPolicy` and `testIamPermissions` methods to be bound with `backup` object and at run time `Table` is undefined if not imported directly from `'./table'`

test/backup.ts
```diff
@@ -529,4 +565,76 @@ describe('Bigtable/Backup', () => {
...
+        .callsFake((_permissions, gaxOpts, callback) => {
+          assert.strictEqual(_permissions, permissions);
+          assert.deepStrictEqual(gaxOpts, {});
+          callback(); // done()
+        });
+      backup.testIamPermissions(permissions, done);
+    });
+
+    it('should accept permissions as array', done => {
+      const permissions = ['bigtable.tables.get', 'bigtable.tables.list'];
```
kolea2, line 617: I think these should be similar to: https://github.com/googleapis/java-bigtable/pull/450/files#diff-84d1a07ba982b6e9cac6f4e6ae0bb75a4531804decf3de85ece5e475c7720801R356-R359

kolea2: (and elsewhere in these tests)

test/backup.ts
```diff
@@ -529,4 +565,76 @@ describe('Bigtable/Backup', () => {
       );
     });
   });
+
+  describe('setIamPolicy', () => {
+    afterEach(() => {
+      sinon.restore();
+    });
+    const policy = {};
+    it('should correctly call Table#setIamPolicy()', done => {
```
kolea2, line 574: `Backup#setIamPolicy()`

kolea2 (changes requested):

test/backup.ts
AVaksman: Done
Thanks

kolea2 (approved):
lgtm, @crwilcox can you take another look?

crwilcox (approved):

PR #830, opened 2020-12-14
feat: implement restore table to a different instance feature

AVaksman:

**#ToDo**

- [x] Add System test
- [ ] Add Samples for
   - [ ] `Backup#restoreTo()`
   - [ ] `Instance#createTableFromBackup()`

### This PR proposes the following design

- Introduce an additional function `Backup#restoreTo(config, callback?)` which accepts a config parameter of the type
   ```ts
   export interface RestoreTableConfig {
     tableId: string;
     instance?: Instance | string;
     gaxOptions?: CallOptions;
   }
   ```
- `Backup#restore()` delegates to `Backup#restoreTo()` with `this.cluster.instance` value for `config.instance`.
- `Instance#createTableFromBackup()` now respects both 
   - `Backup` object created on a different instance.
   - As well as a `string` of full backup name (path) that contains a different (form `this.id`) instanceId.

### Other design options 
**!Breaking:**
- Replace `Backup#restore()` with the proposed `Backup#restoreTo()` implementation 
- Include the changes with current pending **4.0.0** release 

**Non Breaking:**
- Add an optional `instance` parameter to `Backup#restore()`
   - This design might be a bit messier than desired
      ```ts
      restore(
        tableId: string,
        instanceOrGaxOptionsOrCallback?:
          | Instance
          | string
          | CallOptions
          | RestoreTableCallback,
        gaxOptionsOrCallback?: CallOptions | RestoreTableCallback,
        cb?: RestoreTableCallback
      ): void | Promise<RestoreTableResponse> {...}
      ```
    - This design will diverge from the established pattern
       ```ts
        function(OneRequiredParamIfAny, optionsOrCallback?, callback?) {...}
       ```

### DIFF
--- a/test/backup.ts
+++ b/test/backup.ts
@@ -51,6 +51,16 @@ class FakeTable extends Table {
   }
 }
 
+class FakeInstance extends instanceTypes.Instance {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  calledWith_: Array<{}>;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  constructor(...args: [Bigtable, string]) {
+    super(...args);
+    this.calledWith_ = args;
+  }
+}
+
 describe('Bigtable/Backup', () => {
   const BACKUP_ID = 'my-backup';
   let CLUSTER: clusterTypes.Cluster;
@@ -64,6 +74,7 @@ describe('Bigtable/Backup', () => {
     Backup = proxyquire('../src/backup.js', {
       '@google-cloud/promisify': fakePromisify,
       './table.js': {Table: FakeTable},
+      './instance.js': {Instance: FakeInstance},
       pumpify,
     }).Backup;
   });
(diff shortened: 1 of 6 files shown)


stephenplusplus:
@AVaksman is there any additional context for the reasoning of this addition?

kolea2:
> @AVaksman is there any additional context for the reasoning of this addition?

Hey @stephenplusplus! We will be working on this in all languages, here's the Java draft for reference: https://github.com/googleapis/java-bigtable/pull/515

stephenplusplus:
@kolea2 thank you!

test/backup.ts
```diff
@@ -474,8 +577,8 @@ describe('Bigtable/Backup', () => {
         callback(null, ...args);
       };
 
-      backup.restore(
-        tableId,
+      backup.restoreTo(
```
kolea2, line 580: I think we may want to leave some of these tests as is, so we don't lose coverage on `.restore()` functionality?

AVaksman: To have gRPC call implementation only in one place `Backup#restore()` (as more restrictive wrapper since can only restore to the same instance) [delegates](https://github.com/googleapis/nodejs-bigtable/blob/53d9ca586c1a86aa6f86c571dc918ea49dcb85a2/src/backup.ts#L507-L514) to `Backup#restoreTo()` and passing `this.cluster.instance` value as the [`config.instance`](https://github.com/googleapis/nodejs-bigtable/blob/53d9ca586c1a86aa6f86c571dc918ea49dcb85a2/src/backup.ts#L529) param.
The current unit tests for `Backup#restore()` are to verify that it properly delegates to `Backup#restoreTo()` with and without optional `gaxOptions` param.

kolea2: got it, just want to confirm we still have at least one `restore` test in place for coverage?

kolea2 (approved):

kolea2:
Looking good, merging into `backup-different-instance` as a WIP

PR #834, opened 2021-01-06
fix: Renaming region tags to not conflict with documentation snippets

billyjacobson:


billyjacobson:
Docs snippets seems to be failing just because I moved the api reference docs, but should be fine once code is submitted

bcoe (commented):
Does this mean we'll have essentially the same samples in two places? wondering if we could just retire the other samples and have their home be in this repository (this is what we've done for some other libraries).

billyjacobson:
> Does this mean we'll have essentially the same samples in two places? wondering if we could just retire the other samples and have their home be in this repository (this is what we've done for some other libraries).

Potentially yes, but currently these samples were written with the intent for just the API reference docs, not for the cloud site documentation. Right now there are some duplicated samples, and ideally these samples are the same in both places, but I'm trying to separate them for easier indexing by the samples tracker. 

I'm not really sure what other products do for this, but I know I have the same issue in python

bcoe (approved):

PR #847, opened 2021-02-25
fix: rollback to v3 release line

crwilcox:
This reverts commit d37fe9ecdadde8d335fc1f83ab5ac3c3f5813e0d

Release-As: v3.2.3

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -47,11 +47,13 @@ export interface GetInstancesCallback {
   (
     err: ServiceError | null,
     result?: Instance[],
+    failedLocations?: string[],
     response?: google.bigtable.admin.v2.IListInstancesResponse
   ): void;
 }
 export type GetInstancesResponse = [
   Instance[],
+  string[],
   google.bigtable.admin.v2.IListInstancesResponse
 ];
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -607,19 +609,15 @@ export class Bigtable {
...
+   * @property {string[]} 1 locations from which Instance information could not be retrieved
+   * @property {object} 2 The full API response.
    */
   /**
    * @callback GetInstancesCallback
    * @param {?Error} err Request error, if any.
    * @param {Instance[]} instances Array of {@link Instance} instances.
+   * @param {string[]} locations from which Instance information could not be retrieved
    * @param {object} apiResponse The full API response.
-   *     Note: 'failedLocations' property may contain locations from which
```
crwilcox, line 620: @bcoe wdyt about keeping these comments?

bcoe: Do we know whether `failedlocations` is actually populated? I'm curious about the motivation for its removal.

If it's populated, I see no reason not to have the comments?

bcoe (approved):

PR #853, opened 2021-03-17
fix: remove common protos

alexander-fenster:
Common protos do not belong here, they are a part of google-gax.


JustinBeckwith (approved):
I'm kind of curious .... how did these end up here?

JustinBeckwith:
ok woah this was a pre-gax thing

PR #855, opened 2021-04-06
feat: implement support for clusters with CMEK encryption

stephenplusplus:
edit (kolea2): removing doc

This adds support for CMEK configuration of a Cluster when it is created.

#### Creating an instance

```js
await bigtable.createInstance('my-instance', {
  clusters: [
    {
      id: 'my-cluster',
      encryption: {
        kmsKeyName: 'kms-key-name',
      },
      // ... other options
    },
  ],
});
```

#### Creating a cluster

```js
await instance.createCluster('my-cluster', {
  encryption: {
    kmsKeyName: 'kms-key-name',
  },
  // ... other options
});
```

#### The `key` alias

Instead of providing the `encryption` option, the user has a choice to use the alias `key`:

```diff
await bigtable.createInstance('my-instance', {
  clusters: [
    {
      id: 'my-cluster',
-     encryption: {
-       kmsKeyName: 'kms-key-name',
-     },
+     key: 'kms-key-name',
      // ... other options
    },
  ],
});

await instance.createCluster('my-cluster', {
- encryption: {
-   kmsKeyName: 'kms-key-name',
- },
+ key: 'kms-key-name',
  // ... other options
});
```

If both options are set, an error is thrown.

### DIFF
--- a/system-test/bigtable.ts
+++ b/system-test/bigtable.ts
@@ -175,12 +175,15 @@ describe('Bigtable', () => {
     const CMEK_INSTANCE = bigtable.instance(generateId('instance'));
     const CMEK_CLUSTER = CMEK_INSTANCE.cluster(generateId('cluster'));
 
+    const cryptoKeyId = generateId('key');
+    const keyRingId = generateId('key-ring');
+    let keyRingsBaseUrl: string;
+    let cryptoKeyVersionName: string;
+
     before(async () => {
       const projectId = await bigtable.auth.getProjectId();
-      const cryptoKeyId = generateId('key');
-      const keyRingId = generateId('key');
-      const keyRingsBaseUrl = `https://cloudkms.googleapis.com/v1/projects/${projectId}/locations/us-central1/keyRings`;
       kmsKeyName = `projects/${projectId}/locations/us-central1/keyRings/${keyRingId}/cryptoKeys/${cryptoKeyId}`;
+      keyRingsBaseUrl = `https://cloudkms.googleapis.com/v1/projects/${projectId}/locations/us-central1/keyRings`;
 
       await bigtable.auth.request({
         method: 'POST',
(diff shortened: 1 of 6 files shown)


kolea2 (changes requested):
Thanks for working on this! For this to be feature complete, I think we will need to add API to both backups and table as well, specifically to get encryption info on them - see the java impl for more details https://github.com/googleapis/java-bigtable/pull/656/files. Let me know if I'm missing anything :)

stephenplusplus:
Thanks for the review @kolea2! Generally for anything that is accessible through the response from "getMetadata()", we won't create an additional method. However, we can of course make an exception if that's a requirement.

How it works now:

```js
const [metadata] = await backup.getMetadata();
console.log(metadata.encryptionInfo);
```

craiglabenz (approved):
Looks good from a Node perspective.

kolea2:
> Thanks for the review @kolea2! Generally for anything that is accessible through the response from "getMetadata()", we won't create an additional method. However, we can of course make an exception if that's a requirement.
> 
> How it works now:
> 
> ```js
> const [metadata] = await backup.getMetadata();
> console.log(metadata.encryptionInfo);
> ```

Thank you! Makes sense to me, thanks for the explanation. Will take a look at the rest.

system-test/bigtable.ts
```diff
@@ -176,6 +176,85 @@ describe('Bigtable', () => {
...
+    const CMEK_CLUSTER = CMEK_INSTANCE.cluster(generateId('cluster'));
+
+    before(async () => {
+      const projectId = await bigtable.auth.getProjectId();
+      const cryptoKeyId = generateId('key');
+      const keyRingId = generateId('key');
+      const keyRingsBaseUrl = `https://cloudkms.googleapis.com/v1/projects/${projectId}/locations/us-central1/keyRings`;
+      kmsKeyName = `projects/${projectId}/locations/us-central1/keyRings/${keyRingId}/cryptoKeys/${cryptoKeyId}`;
+
+      await bigtable.auth.request({
```
kolea2, line 192: is this creating a key per test run? If so, I think these will need to be cleaned up as well?

kolea2 (changes requested):

system-test/bigtable.ts
stephenplusplus: Good catch, it was. I added a commit that uses the [`destroy`](https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys.cryptoKeyVersions/destroy) method to schedule the destroy of the key. Please let me know if there's a better or more thorough way to remove these crypto key resources!

kolea2 (approved):

PR #882, opened 2021-05-26
chore: migrate to owl bot

SurferJeffAtGoogle:


SurferJeffAtGoogle:
I believe this code is good to merge.   @tmatsuo   Please review.

owlbot.py
```diff
@@ -0,0 +1,63 @@
...
+        library = staging / version
+        _tracked_paths.add(library)
+        s.copy([library], excludes=['package.json', 'README.md', 'src/index.ts', 'src/v2/index.ts', 'tsconfig.json', 'tslint.json'])
+
+    # Copy the admin library.
+    # Not override system-test for admin/v2, just keep the v2 version.
+    for version in ['v2']:
+        library = staging / 'admin' / version
+        _tracked_paths.add(library)
+        s.copy([library], excludes=['package.json', 'README.md', 'src/index.ts', 'src/v2/index.ts', 'tsconfig.json', 'tslint.json', 'system-test/fixtures/sample/src/index.ts', 'system-test/fixtures/sample/src/index.js'])
```
chingor13, line 46: nit: these could be in the same loop

chingor13 (approved):

owlbot.py
SurferJeffAtGoogle: Duly noted.  I'm not sure that when they release v3, it will be released at exactly the same time for each library.  I wanted to leave a code structure that would not require refactoring when v3 is released.

PR #947, opened 2021-10-06
feat: make INTERNAL RST_STREAM errors retryable

igorbernstein2:
Normally RST_STREAM frames will get wrapped in UNAVAILABLE statuses.
However if the stream is reset higher upstream, it can trigger an INTERNAL error.
ReadRows doesn't really care what was the trigger, so this PR will treat INTERNAL
RST_STREAMs as UNAVAILABLE and retry them

Prior art:
https://github.com/googleapis/java-bigtable/pull/586 
https://github.com/googleapis/java-bigtable/pull/1029
https://github.com/googleapis/java-bigquerystorage/pull/419

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -45,7 +45,9 @@ import {Duplex} from 'stream';
 
 // See protos/google/rpc/code.proto
 // (4=DEADLINE_EXCEEDED, 10=ABORTED, 14=UNAVAILABLE)
-const RETRYABLE_STATUS_CODES = new Set([4, 10, 14]);
+const UNAVAILABLE_STATUS_CODE = 14;
+const INTERNAL_STATUS_CODE = 13;
+const RETRYABLE_STATUS_CODES = new Set([4, 10, UNAVAILABLE_STATUS_CODE]);
 // (1=CANCELLED)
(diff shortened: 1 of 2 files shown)


JustinBeckwith (approved):
LGTM, but I'd like @alexander-fenster and @murgatroid99 to have a look here to make sure there isn't something more broadly applicable to be done

murgatroid99:
> Normally RST_STREAM frames will get wrapped in UNAVAILABLE statuses.

This is simply not true. RST_STREAM frames have an error code, which is analogous to HTTP/1.1 error codes, and gRPC maps RST_STREAM to a gRPC status based on the error code, not on where upstream the RST_STREAM frame originated. Most of those error codes actually map to INTERNAL, and the only one that maps to UNAVAILABLE is REFUSED_STREAM. The full mapping can be found [here](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#errors).

crwilcox:
@murgatroid99 are there any of the codes from rst stream shown at https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#errors that are not retry safe? I think if the underlying call is idempotent they are all alright to retry?

PR #952, opened 2021-10-15
docs: stop publishing autogenerated gapic and protobuf docs

### LINKED ISSUE #951: Update docs to explicitly steer users to use the correct public surface
Currently our docs include the GAPIC generated stubs and protobuf documentation. Unfortunately the GAPIC stubs (ie v2.BigtableClient) do not provide enough functionality to use bigtable correctly (it lacks row merging and proper retries).

It is also unclear to me how much of the autogenerated types are used on the public surface. Ideally any types that aren't used on the public surface should be removed from the autogenerated docs.

igorbernstein2:
The Bigtable data protocol is substantially different from it's api and cannot
be expressed using GAPIC generator alone. Instead a manually written layer 
was created to sit on top of the raw gapic & protobuf apis. Thus our public 
surface is the manual layer, while the autogenerated code is considered an 
internal implementation detail. To avoid confusing users, we should stop 
publishing docs for internal apis.

Fixes #951


kolea2 (approved):
lgtm with one question

danieljbruce:
I found https://stackoverflow.com/questions/56109736/jsdoc-is-not-excluding-paths which may offer a solution to the problem. I'll try tomorrow to change the `excludePattern` to use the `or` operator in a regular expression to perhaps exclude the directory that contains v2 docs.

kolea2 (approved):

PR #986, opened 2022-01-14
fix: add close() in the wrapper

mutianf:

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -887,21 +887,13 @@ export class Bigtable {
    * Terminate grpc channels and close all the clients.
    */
   close(): Promise<void[]> {
-    const combined = [];
-    if (this.api['BigtableInstanceAdminClient']) {
-      combined.push(this.api['BigtableInstanceAdminClient'].close());
-    }
-    if (this.api['BigtableTableAdminClient']) {
-      combined.push(this.api['BigtableTableAdminClient'].close());
-    }
-    if (this.api['BigtableClient']) {
-      combined.push(this.api['BigtableClient'].close());
-    }
+    const combined = Object.keys(this.api).map(clientType =>
+      this.api[clientType].close()
+    );
     if (combined.length === 0) {
       return Promise.resolve([]);
-    } else {
-      return Promise.all(combined);
     }
+    return Promise.all(combined);
   }
 
   /**
--- a/src/index.ts
+++ b/src/index.ts
@@ -897,7 +897,7 @@ export class Bigtable {
     if (this.api['BigtableClient']) {
       combined.push(this.api['BigtableClient'].close());
     }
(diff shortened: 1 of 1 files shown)


igorbernstein2 (approved):
lgtm

src/index.ts
```diff
@@ -883,6 +883,27 @@ export class Bigtable {
     }
   }
 
+  /**
+   * Terminate grpc channels and close all the clients.
+   */
+  close(): Promise<void> | Promise<void[]> {
```
danieljbruce, line 889: The only thing I would consider here is that since `Promise.all(combined)` returns an array then we should replace `Promise.resolve()` with `Promise.resolve([])` so that the output is always the same type. Maybe that's a bad idea though and for some reason we want the types to be different. Not sure.

danieljbruce (commented):
Just one comment on a return value

src/index.ts
```diff
@@ -883,6 +883,27 @@ export class Bigtable {
     }
   }
 
+  /**
+   * Terminate grpc channels and close all the clients.
+   */
+  close(): Promise<void[]> {
+    const combined = [];
```
danieljbruce, line 890: I think to achieve what you want here we can condense the first ten lines of this function to `const combined = Object.keys(this.api).map(clientType => this.api[clientType].close())`.

danieljbruce (commented):
Advice on making this a little more concise.

src/index.ts
danieljbruce: If this isn't exactly right I can try to tweak it later or help with debugging.

mutianf: This works, thank you!

src/index.ts
```diff
@@ -883,6 +883,19 @@ export class Bigtable {
     }
   }
 
+  /**
+   * Terminate grpc channels and close all the clients.
+   */
+  close(): Promise<void[]> {
```
kolea2, line 889: any good way to test this?

mutianf: Maybe we can add a test to make sure request will fail after client is closed, I'll work on it next week!

src/index.ts
```diff
@@ -883,6 +883,19 @@ export class Bigtable {
...
+    const combined = Object.keys(this.api).map(clientType =>
+      this.api[clientType].close()
+    );
+    if (combined.length === 0) {
```
danieljbruce, line 893: Just a small nitpick here. I did a test and it looks like `Promise.resolve` does the same thing as `Promise.all` with an empty array. So that means that these last 4 lines can just be `return Promise.all(combined);` unless there was some specific reason you added the if statement.

image

danieljbruce (commented):
Suggestions for end of close function

PR #992, opened 2022-01-20
test: add test for close()

danieljbruce:
Added a test case for the new close function

### DIFF
--- a/test/index.ts
+++ b/test/index.ts
@@ -1149,4 +1149,17 @@ describe('Bigtable', () => {
       });
     });
   });
+  describe('test the close function', () => {
+    it('request should fail after close is called', done => {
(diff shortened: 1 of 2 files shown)


test/index.ts
```diff
@@ -1149,4 +1149,19 @@ describe('Bigtable', () => {
...
   });
+  describe('close', () => {
+    it('should have failed request after close is called', done => {
+      bigtable.close().then(() => {
+        bigtable.getInstances((err: Error) => {
+          if (err) {
+            done();
+          } else {
+            assert.fail(
+              'The request did not fail, but it should have because the connection closed'
```
mutianf, line 1160: nit: because the connection **is** closed

mutianf (approved):

PR #1007, opened 2022-02-01
feat: Multi cluster routing

danieljbruce:
This PR allows users of the client library to provide a list of clusters in an API request so that routing of requests will stay restricted to that group of clusters.


system-test/common.ts
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as uuid from 'uuid';
+import {Cluster} from '../src/cluster';
+import * as inst from '../src/instance';
+
+export const PREFIX = 'gcloud-tests-';
+
+export function generateId(resourceType: string) {
```
kolea2, line 21: same question here

src/app-profile.ts
```diff
@@ -29,7 +29,7 @@ export interface AppProfileOptions {
    * value is required when creating the app profile and optional when setting
    * the metadata.
    */
-  routing?: 'any' | Cluster;
+  routing?: 'any' | Cluster | Array<Cluster>;
```
kolea2, line 32: The Java client uses a set instead of an array. Is there a compelling reason to have an array here? If no, can we switch to a set?

kolea2 (changes requested):

src/app-profile.ts
danieljbruce: Sounds like a good plan. I can change to a set. The only reason we would want to use an array is if we want for some reason for the order to be preserved every time we make a call to get the clusters in an app profile. I can't see why we would care about that though so a set is fine.

system-test/common.ts
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as uuid from 'uuid';
+import {Cluster} from '../src/cluster';
+import * as inst from '../src/instance';
+
+export const PREFIX = 'gcloud-tests-';
```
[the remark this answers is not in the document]
danieljbruce: These variables were moved into a file called common.js because now two files use them. `app-profile.ts` is one of them and the other is `instance.ts`.

system-test/common.ts
danieljbruce: These variables were moved into a file called common.js because now two files use them. `app-profile.ts` is one of them and the other is `instance.ts`.

system-test/app-profile.ts
```diff
@@ -0,0 +1,79 @@
...
+      });
+      await operation.promise();
+
+      // This is for creating an app profile.
+      const appProfileId = generateId('app-profile');
+      const clusterSubset = clusterIds
+        .slice(1, 3)
+        .map(clusterId => instance.cluster(clusterId));
+      const options = {
+        routing: clusterSubset,
```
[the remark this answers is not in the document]
danieljbruce: Sounds good. I'll go make this change.

src/app-profile.ts
danieljbruce: The only consideration is that in the response object we get an array so maybe we want an array input for consistency, but probably it doesn't matter.

system-test/app-profile.ts
```diff
@@ -0,0 +1,79 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {describe, it} from 'mocha';
+import {generateId} from './common';
+import {Bigtable} from '../src';
+import assert = require('assert');
+
+describe('📦 App Profile', () => {
```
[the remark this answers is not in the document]
danieljbruce: Added these tests. In the process of adding these tests it looks like a lot of restructuring was required in order to eliminate duplicate code fragments so let me know if you have any questions regarding the changes.

src/app-profile.ts
```diff
@@ -29,7 +29,7 @@ export interface AppProfileOptions {
    * value is required when creating the app profile and optional when setting
    * the metadata.
    */
-  routing?: 'any' | Cluster;
+  routing?: 'any' | Cluster | Set<Cluster>;
```
kolea2, line 32: a thought - it might be nice to provide an additional method which takes in a set of strings - we don't need the cluster object anywhere, so this could be easier on the user rather than having to construct a Cluster object.

test/instance.ts
```diff
@@ -263,6 +256,23 @@ describe('Bigtable/Instance', () => {
         };
         instance.createAppProfile(APP_PROFILE_ID, options, assert.ifError);
       });
+
+      it('an array of clusters', done => {
```
kolea2, line 260: this this needs a description and change to "set"

system-test/app-profile.ts
```diff
@@ -0,0 +1,130 @@
...
+        appProfile.metadata?.multiClusterRoutingUseAny?.clusterIds,
+        []
+      );
+    });
+
+    it('should ensure clusters match an updated profile', async () => {
+      const options = {
+        routing: instance.cluster(clusterIds[1]),
+      };
+      const appProfile = await createProfile(instance, options);
```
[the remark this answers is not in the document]
danieljbruce: Good point. Should I now eliminate the test that creates a single cluster since that test is now exactly the same as the first part of this one?

test/instance.ts
danieljbruce: Changed to say `set`. Regarding description, the other tests in the same describe block say `'a cluster value'` and `"an 'any' value"`.  Should those descriptions change along with this to something like `should accept an 'any' value`?

src/app-profile.ts
```diff
@@ -213,7 +235,7 @@ Please use the format 'my-app-profile' or '${instance.name}/appProfiles/my-app-p
         }
       } else {
         throw new Error(
-          'An app profile routing policy can only contain "any" or a `Cluster`.'
+          'An app profile routing policy can only contain "any", a `Cluster` or a set of strings/`Cluster`s.'
```
kolea2, line 238:
```suggestion
          'An app profile routing policy can only contain "any" for multi cluster routing, a `Cluster` for single routing,  or a set of clusterIds as strings or `Clusters` for mutli cluster routing.'
```

system-test/app-profile.ts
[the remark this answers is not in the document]
kolea2: I would keep existing tests as is for now. We can do cleanup later

src/app-profile.ts
```diff
@@ -29,7 +29,7 @@ export interface AppProfileOptions {
    * value is required when creating the app profile and optional when setting
    * the metadata.
    */
-  routing?: 'any' | Cluster;
+  routing?: 'any' | Cluster | Set<Cluster> | Set<string>;
```
kolea2, line 32: let's add a test for this new option `Set<string>`

kolea2 (changes requested):

protos/protos.d.ts
```diff
@@ -330,140 +330,140 @@ export namespace google {
                 namespace BigtableInstanceAdmin {
 
                     /**
-                     * Callback as used by {@link google.bigtable.admin.v2.BigtableInstanceAdmin#createInstance}.
+                     * Callback as used by {@link google.bigtable.admin.v2.BigtableInstanceAdmin|createInstance}.
```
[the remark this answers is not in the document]
danieljbruce: Good point. I think this was generated by `npm run fix` to fix linting issues. Should I go through and manually revert all of these # to | changes?

src/app-profile.ts
danieljbruce: You can find this under the test that says `should create a profile with multiple clusters using strings`.

bcoe:
Just an FYI, if you're working on a branch from a fork, add `owlbot:run` and `kokoro:force-run` to kick off tests/post-processing.

test/instance.ts
```diff
@@ -263,6 +256,23 @@ describe('Bigtable/Instance', () => {
         };
         instance.createAppProfile(APP_PROFILE_ID, options, assert.ifError);
       });
+
+      it('a set of clusters', done => {
```
kolea2, line 260: nit - update this description

kolea2 (commented):
lgtm after last small comment

test/instance.ts
danieljbruce: I updated this to say `a set of cluster objects`, but I think you might have wanted something else. In this describe block our description string says, `"should respect the routing option with"` and then each `it` describes what we pass into the routing option. In this case, on the line that says, `const options = {routing: new Set(clusters)};` we are setting the routing option to a set of clusters.

kolea2 (approved):
lgtm, would appreciate if @bcoe could also sign off before merging :) thanks!

bcoe (approved):
Left a couple non-blocking recommendations.

system-test/common.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export function generateId(resourceType: string) {
+  return PREFIX + resourceType + '-' + uuid.v1().substr(0, 8);
```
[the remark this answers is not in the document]
danieljbruce: I'll create a separate issue for this since I tried it and there was more cleanup for the sake of getting this PR in and keeping things separate.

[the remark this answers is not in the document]
danieljbruce: https://github.com/googleapis/nodejs-bigtable/issues/1084

PR #1011, opened 2022-02-03
fix: manual generate library for ongoing stream on closed client

summer-ji-eng:
Manual update generator [v2.13.0](https://github.com/googleapis/gapic-generator-typescript/releases/tag/v2.13.0).

### DIFF
--- a/samples/generated/v2/snippet_metadata.google.bigtable.v2.json
+++ b/samples/generated/v2/snippet_metadata.google.bigtable.v2.json
@@ -1,0 +1,327 @@
+{
+  "clientLibrary": {
+    "name": "nodejs-bigtable",
+    "version": "0.1.0",
+    "language": "TYPESCRIPT",
+    "apis": [
+      {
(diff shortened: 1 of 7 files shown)


mutianf (approved):

samples/generated/v2/snippet_metadata.google.bigtable.v2.json
```diff
@@ -0,0 +1,327 @@
+{
+  "clientLibrary": {
+    "name": "nodejs-bigtable",
+    "version": "0.1.0",
```
summer-ji-eng, line 4: @sofisl could you take a look? Is the version right to you?

sofisl: This should be the current version of the library, but there's no way to put that information into the generator. I think, for now, we should just change it manually, but then have release-please update this value when a new release is made. I'll open a bug on release-please.

PR #1024, opened 2022-03-10
fix: this makes the full table scan issue go away

danieljbruce:
This fixes the full table scan issue and prevents it from happening when calling readRowStream with a row range as described in buganizer using the emulator.


igorbernstein2:
Please mark incomplete PRs as drafts

PR #1025, opened 2022-03-14
fix: prevent full table scan from happening

danieljbruce:
This change prevents a full table scan from happening unless the user did one intentionally.


src/table.ts
```diff
@@ -895,6 +895,10 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
         reqOpts.rowsLimit = rowsLimit - rowsRead;
       }
 
+      if (!reqOpts.rows && lastRowKey) {
+        return
```
igorbernstein2, line 899: This will just leave the userStream hanging indefinitely, causing the caller's application to hang

PR #1026, opened 2022-03-15
fix: fix createReadStream retry from sending a full table scan

mutianf:

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -733,7 +733,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
     const ranges = options.ranges || [];
     let filter: {} | null;
     const rowsLimit = options.limit || 0;
-    const hasLimit = rowsLimit != 0;
+    const hasLimit = rowsLimit !== 0;
     let rowsRead = 0;
     let numRequestsMade = 0;
 
@@ -836,7 +836,8 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
             !startValue ||
             lessThanOrEqualTo(startValue as string, lastRowKey as string);
           const endKeyIsNotRead =
-            !endValue || (endValue as Buffer).length === 0 ||
+            !endValue ||
+            (endValue as Buffer).length === 0 ||
             lessThan(lastRowKey as string, endValue as string);
           if (startKeyIsRead) {
             if (endKeyIsNotRead) {
--- a/src/table.ts
+++ b/src/table.ts
@@ -733,6 +733,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
     const ranges = options.ranges || [];
     let filter: {} | null;
     let rowsLimit: number;
+    let hasLimit = false;
     let rowsRead = 0;
     let numRequestsMade = 0;
 
@@ -750,6 +751,8 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
 
     if (options.keys) {
       rowKeys = options.keys;
+    } else {
+      rowKeys = null;
     }
 
     if (options.prefix) {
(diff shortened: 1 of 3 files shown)


src/table.ts
```diff
@@ -866,7 +864,22 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
             rowKeys = null;
           }
         }
+
```
igorbernstein2, line 867: for rowKeys above I would check if rowKeys.length == 0

src/table.ts
```diff
@@ -866,7 +864,22 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+        // If there was a row limit in the original request and
+        // we've already read all the rows, end the stream and
+        // do not retry.
+        if (hasLimit && rowsLimit === rowsRead) {
+          userStream.end();
+          return;
+        }
+        // If all the row keys and ranges are read, end the stream
+        // and do not retry.
+        if (!rowKeys && ranges.length === 0) {
```
igorbernstein2, line 877: check rowKeys.length == 0 as well

src/table.ts
```diff
@@ -866,7 +864,22 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+          userStream.end();
+          return;
+        }
       }
+
       if (rowKeys || ranges.length) {
         reqOpts.rows = {};
```
igorbernstein2, line 885: Why do you need the if statements below, the loops will be no-ops for empty arrays

igorbernstein2 (changes requested):

src/table.ts
```diff
@@ -772,12 +775,18 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       });
     }
 
+    // If rowKeys and ranges are both empty, the request is a full table scan
```
danieljbruce, line 778: I think we actually want to remove this block of code. I am not sure what it is trying to accomplish, but note that we don't need it since a full table scan happens anyways under these conditions.

danieljbruce (commented):
Just a comment about the conditions that result in a full table scan.

src/table.ts
```diff
@@ -814,45 +823,34 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+          const range = ranges[index];
+          const startValue = is.object(range.start)
+            ? (range.start as BoundData).value
+            : range.start;
+          const endValue = is.object(range.end)
+            ? (range.end as BoundData).value
+            : range.end;
+          const startKeyIsRead =
+            !startValue ||
+            lessThanOrEqualTo(startValue as string, lastRowKey as string);
```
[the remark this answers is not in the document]
mutianf: startValue could be undefined if it's an empty range or the range doesn't have a startkey

system-test/data/read-rows-retry-test.json
```diff
@@ -151,8 +171,8 @@
         "keys": ["a", "b", "x"]
       },
       "request_options": [
-        { "rowKeys": [ "a", "b", "x" ] },
-        { "rowKeys": [ "x" ], "rowRanges": [ { "startKeyOpen": "c" } ] }
+        { "rowKeys": [ "a", "b", "x" ], "rowRanges": [] },
```
mutianf, line 174: I think this test was wrong, it added a new range for a request that only had row keys right?

src/table.ts
mutianf: This section is incorrect: https://github.com/googleapis/nodejs-bigtable/blob/main/src/table.ts#L820-L827

If the original request only had rowKeys and no ranges, it'll add a new range to the original request. It's only correct when the original request is a full table scan.

To simplify the logic here, we add an empty range if the request was a full table scan. So when we're retrying, we don't need to manually add a new range. When it loops through all the ranges, it'll see that end key of empty range: {} is **after** lastRowKey, so it'll add range (startKey, _) to the retry request.

danieljbruce (commented):
LGTM

system-test/data/read-rows-retry-test.json
danieljbruce: It does seem strange that we would actually add a row range on the second request. What happens if we remove `"rowRanges": []` though?

igorbernstein2 (approved):
lgtm. after the buffer end check

igorbernstein2 (approved):
lgtm

PR #1060, opened 2022-03-25
fix: disable retry-request, add exponential backoff in mutateRows and readRows

mutianf:
This pr has the following changes:

- Add exponential retry backoff in readRows and mutateRows in `table.ts`. For readRows, retry delay and retry attempt count will reset whenever a new row is read (this behavior is consistent with the java client).
- Removed decorateStatus. MutateRows PartialFailureErrors will return grpc status code.
- Add RPC level retries for mutatesRows.
- If there's an RPC level failure, add the RPC status code to mutateRows PartialFailureError.
- Do not retry ABORT errors for mutate rows.
- Ignore retry-request settings to avoid request being retried in multiple places.


src/table.ts
```diff
@@ -1507,13 +1508,26 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
     const onBatchResponse = (
       err: ServiceError | PartialFailureError | null
     ) => {
-      // TODO: enable retries when the entire RPC fails
       if (err) {
-        // The error happened before a request was even made, don't retry.
+        // Retry RPC level errors
+        if (!(err instanceof PartialFailureError)) {
+          const serviceError = err as ServiceError;
```
igorbernstein2, line 1514: should this be `if (err instanceof ServiceError)`?

src/table.ts
```diff
@@ -1507,13 +1508,26 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+          if (
+            numRequestsMade <= maxRetries &&
+            IDEMPOTENT_RETRYABLE_STATUS_CODES.has(serviceError.code)
+          ) {
+            console.log('RETRYING ' + err.code);
```
igorbernstein2, line 1519: should this be debug?

src/table.ts
```diff
@@ -1507,13 +1508,26 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+            makeNextBatchRequest();
```
igorbernstein2, line 1520: Im a bit concerned that there is no backoff...it can result in a death spiral for the service

mutianf: Should we add a fixed delay for now?

src/table.ts
mutianf: Sorry this should be cleaned up

src/table.ts
mutianf: I couldn't do `if (err instanceof ServiceError)` because ServiceError is an interface so I can't do instanceof on it 🤔

src/table.ts
```diff
@@ -940,13 +963,21 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
             numRequestsMade <= maxRetries &&
             RETRYABLE_STATUS_CODES.has(error.code)
           ) {
-            makeNewRequest();
+            const backOffSettings =
+              options.gaxOptions?.retry?.backoffSettings ||
+              DEFAULT_BACKOFF_SETTINGS;
+            const nextRetryDelay = getNextDelay(
+              numRequestsMade,
```
igorbernstein2, line 970: should this be numConsecutiveErrors? In other words should backoff be rest when receiving a successful result? Maybe add a TODO to figure this out

src/table.ts
```diff
@@ -1504,23 +1535,27 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
-    const onBatchResponse = (
-      err: ServiceError | PartialFailureError | null
-    ) => {
-      // TODO: enable retries when the entire RPC fails
-      if (err) {
-        // The error happened before a request was even made, don't retry.
-        callback(err);
-        return;
+    const isRetryable = (err: ServiceError | null) => {
+      if (pendingEntryIndices.size === 0 || numRequestsMade > maxRetries) {
```
igorbernstein2, line 1539: should it numRequestsMade >= maxRetries?

src/table.ts
```diff
@@ -1504,23 +1535,27 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+        return false;
       }
-      if (pendingEntryIndices.size !== 0 && numRequestsMade <= maxRetries) {
-        makeNextBatchRequest();
+      return !err || IDEMPOTENT_RETRYABLE_STATUS_CODES.has(err.code);
+    };
+
+    const onBatchResponse = (err: ServiceError | null) => {
+      if (isRetryable(err as ServiceError)) {
```
igorbernstein2, line 1546: is the cast still necessary here?

src/table.ts
```diff
@@ -1541,6 +1576,12 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
 
       const retryOpts = {
         currentRetryAttempt: numRequestsMade,
+        // Handling retries in this client. Specify the retry options to
+        // make sure nothing is retried in retry-request.
+        noResponseRetries: 0,
+        shouldRetryFn: function (response: any) {
```
igorbernstein2, line 1582: s/response/_

src/table.ts
```diff
@@ -1552,8 +1593,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
           retryOpts,
         })
         .on('error', (err: ServiceError) => {
-          // TODO: this check doesn't actually do anything, onBatchResponse
-          // currently doesn't retry RPC errors, only entry failures
+          // The error happened before a request was even made, don't retry.
           if (numRequestsMade === 0) {
             callback(err); // Likely a "projectId not detected" error.
             return;
```
igorbernstein2, line 1599: consider moving this check to onBatchResponse...no reason to split it

src/table.ts
```diff
@@ -1504,23 +1535,27 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
       }
 
       if (mutationErrorsByEntryIndex.size !== 0) {
         const mutationErrors = Array.from(mutationErrorsByEntryIndex.values());
-        err = new PartialFailureError(mutationErrors);
+        callback(new PartialFailureError(mutationErrors));
+        return;
       }
 
       callback(err);
```
igorbernstein2, line 1561: I think you need to massage this a bit. If there are no mutations left, then it should clear the err. ie race between deadline_exceeded after sending all of the results

igorbernstein2 (changes requested):

src/table.ts
mutianf: Yes, I think numRequestsMade > maxRetries is more correct, but currently it's implemented as `numRequestsMade > maxRetries`, so I'm wondering if I should change that?

src/table.ts
mutianf: I think in normal case it's returning callback(undefined) https://github.com/googleapis/nodejs-bigtable/blob/main/src/table.ts#L1584

src/table.ts
```diff
@@ -1504,23 +1538,44 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+          DEFAULT_BACKOFF_SETTINGS;
+        const nextDelay = getNextDelay(numRequestsMade, backOffSettings);
+        setTimeout(makeNextBatchRequest, nextDelay);
         return;
       }
 
-      if (mutationErrorsByEntryIndex.size !== 0) {
-        const mutationErrors = Array.from(mutationErrorsByEntryIndex.values());
-        err = new PartialFailureError(mutationErrors);
+      if (numRequestsMade !== 0) {
```
igorbernstein2, line 1564: move this up and bail early

src/table.ts
```diff
@@ -1504,23 +1538,44 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+        // If there's a race condition where all the mutations
+        // succeeded, but the server returned an error (like deadline
+        // exceeded), set error to null
+        if (pendingEntryIndices.size === 0 && err) {
+          err = null;
+        }
```
igorbernstein2, line 1570: just simplify to:
// No work to be done, exit successfully
if (pendingEntryIndices.size === 0) {
  err = null;
}

src/table.ts
```diff
@@ -1504,23 +1538,44 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+
+        if (mutationErrorsByEntryIndex.size !== 0) {
+          const mutationErrors = Array.from(
+            mutationErrorsByEntryIndex.values()
+          );
+          callback(new PartialFailureError(mutationErrors));
```
igorbernstein2, line 1576: This loses the rpcError. I would try to add an extra attribute to PartialFailureError. Or just return the rpcError

src/table.ts
```diff
@@ -937,19 +958,30 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
-            makeNewRequest();
+            const backOffSettings =
+              options.gaxOptions?.retry?.backoffSettings ||
+              DEFAULT_BACKOFF_SETTINGS;
+            const nextRetryDelay = getNextDelay(
+              numConsecutiveErrors,
+              backOffSettings
+            );
+            retryTimer = setTimeout(makeNewRequest, nextRetryDelay);
+            numConsecutiveErrors++;
```
igorbernstein2, line 972: I think this should be moved to before the if statement. Probably the first thing in the error callback.  (and `numConsecutiveErrors <` changed to `numConsecutiveErrors <=`)

src/table.ts
```diff
@@ -937,19 +958,30 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
           } else {
             userStream.emit('error', error);
           }
         })
+        .on('data', _ => {
+          numConsecutiveErrors = 0;
+        })
         .on('end', () => {
           activeRequestStream = null;
+          retryTimer = null;
```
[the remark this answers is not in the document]
igorbernstein2: Do you need to unpipe here?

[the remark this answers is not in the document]
mutianf: Yeah, I don't think we need to set it again since we reset it to null in the beginning of a new request. 

I think it's unpiped at line 796? 

```
    userStream.end = () => {
      rowStream?.unpipe(userStream);
```

src/table.ts
mutianf: I think it should still be numConsecutiveErrors < because we're not counting the first request we sent

igorbernstein2: ok, please move it :)

src/table.ts
```diff
@@ -937,19 +957,29 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+              numConsecutiveErrors,
+              backOffSettings
+            );
+            retryTimer = setTimeout(makeNewRequest, nextRetryDelay);
           } else {
             userStream.emit('error', error);
           }
         })
+        .on('data', _ => {
+          numConsecutiveErrors = 0;
```
igorbernstein2, line 977: Please add a comment explaining that we reset retries when we get a successful result

src/index.ts
```diff
@@ -842,7 +841,11 @@ export class Bigtable {
           currentRetryAttempt: 0,
           noResponseRetries: 0,
           objectMode: true,
-          shouldRetryFn: shouldRetryRequest,
+          // TODO: re-enable shouldRetryFn when the retry
+          // logic is fixed in gax / retry-request
+          shouldRetryFn: (_: any) => {
+            return false;
+          },
```
igorbernstein2, line 848: Why does this have to happen here? don't you overwrite there in mutateRows & readRows methods?

src/table.ts
mutianf: I wait I think i misunderstood your comment. Updated it

igorbernstein2 (approved):
lgtm!

igorbernstein2 (approved):
lgtm!

PR #1062, opened 2022-04-01
feat: enable channel pooling

mutianf:
Enabling channel pooling with grpc-gcp-node. Max channel pool size is set to 10. It'll create a new channel when a channel has more than 30 concurrent streams.


igorbernstein2:
closing in favor of #1065

PR #1068, opened 2022-04-12
feat: send retry attempt header to ease debugging

igorbernstein2:
This is the equivilent of https://github.com/googleapis/java-bigtable/pull/935

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -920,7 +920,8 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
 
       options.gaxOptions = populateAttemptHeader(
         numRequestsMade,
-        options.gaxOptions);
+        options.gaxOptions
+      );
 
       const requestStream = this.bigtable.request({
         client: 'BigtableClient',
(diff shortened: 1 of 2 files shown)


src/table.ts
```diff
@@ -2072,6 +2084,14 @@ function getNextDelay(numConsecutiveErrors: number, config: BackoffSettings) {
   return Math.min(calculatedNextRetryDelay, config.maxRetryDelayMillis);
 }
 
+function populateAttemptHeader(attempt: number, gaxOpts?: CallOptions) {
```
danieljbruce, line 2087: Looks like a good approach to me. I think a good TODO eventually would be to create a function for these next three lines of code because we repeat it in `mutateRows` and `readRows`, but we can do that later.

test/table.ts
```diff
@@ -2707,6 +2733,36 @@ describe('Bigtable/Table', () => {
           done();
         });
       });
+
+      it('should send attempt header', done => {
+        const error = new Error('retryable') as ServiceError;
+        error.code = 14; // Unavailable
```
danieljbruce, line 2739: In the future, I think there is an enum that we should use instead of `14`, but not super critical.

danieljbruce (approved):
A few small style comments, but a great feature to have overall.

src/table.ts
igorbernstein2: This is already extracted to its own function?

test/table.ts
igorbernstein2: agreed, but thats out of scope for this PR

PR #1070, opened 2022-04-13
feat: enable retries for RESOURCE_EXHAUSTED an simplify error codes

igorbernstein2:
Unify retriable status codes for ReadRows and MutateRows to be DEADLINE_EXCEEDED -  RESOURCE_EXHAUSTED, ABORTED, UNAVAILABLE.

### DIFF
--- a/test/table.ts
+++ b/test/table.ts
@@ -2485,8 +2485,8 @@ describe('Bigtable/Table', () => {
           {
             index: 1,
             status: {
-              code: 10,
-              message: 'ABORTED',
+              code: 3,
+              message: 'INVALID_ARGUMENT',
             },
           },
         ];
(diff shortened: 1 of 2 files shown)


test/table.ts
```diff
@@ -2682,7 +2682,7 @@ describe('Bigtable/Table', () => {
 
       it('should not retry unretriable errors', done => {
         const unretriableError = new Error('not retryable') as ServiceError;
-        unretriableError.code = 10; // Aborted
+        unretriableError.code = 3; // INVALID_ARGUMENT
```
danieljbruce, line 2685: I see. Looks like this change had to be made because of the new retry code we are allowing.

danieljbruce (approved):
LGTM. I'm not too sure why idempotent requests would retry based on a different error set before so it seems reasonable to retry based on the same set of errors no matter what type of call we make as per the changes.

PR #1077, opened 2022-04-21
feat: Autoscaler

danieljbruce:
This feature allows the user to set properties when creating a cluster or updating a cluster that turn on the autoscaler or turn it off to use manual scaling.


src/cluster.ts
```diff
@@ -75,7 +76,10 @@ export type GetClustersCallback = (
   apiResponse?: google.bigtable.admin.v2.IListClustersResponse
 ) => void;
 export interface SetClusterMetadataOptions {
-  nodes: number;
+  nodes?: number;
```
kolea2, line 79: is this considered breaking if we change something from required to optional? I'm guessing it would be the other way around, but want to confirm :)

src/cluster.ts
```diff
@@ -690,35 +697,42 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
+      );
+    };
+    if (this.metadata && this.metadata.location) {
+      setMetadataWithLocation(this.metadata.location, this.name);
+    } else {
+      this.getMetadata(gaxOptions, (err, res) => {
+        if (err || !this.metadata || !this.metadata.location) {
+          callback(err, null);
+        } else {
+          setMetadataWithLocation(this.metadata.location, this.name);
```
kolea2, line 732: why do we need to reset the metadata in this case?

src/instance.ts
```diff
@@ -415,17 +422,6 @@ Please use the format 'my-instance' or '${bigtable.projectName}/instances/my-ins
...
     }
 
-    if (options.location) {
-      reqOpts.cluster!.location = Cluster.getLocation_(
-        this.bigtable.projectId,
-        options.location
-      );
-    }
-
-    if (options.nodes) {
```
kolea2, line 425: why is this deleted?

src/cluster.ts
danieljbruce: Here we are changing something from required to optional so that would make it not a breaking change since users can still pass just `nodes` in. This is where we make `nodes` optional because instead the user can provide autoscaling parameters. These parameters will be validated against a new validation function and an error will be thrown if an invalid combination of parameters are provided.

src/instance.ts
danieljbruce: `cluster.serveNodes` is now provided by the call to `ClusterUtils.getClusterBaseConfig` above. This change is a refactor so that this function doesn't get too verbose.

system-test/cluster.ts
```diff
@@ -0,0 +1,313 @@
...
+          try {
+            await cluster2.create({
+              location: 'us-west1-c',
+            });
+            assert.fail();
+          } catch (e) {
+            assert.equal(e.message, ClusterUtils.noConfigError);
+          }
+        });
+        it('By providing too much cluster configurations', async () => {
```
kolea2, line 105: nit - reword to express that the user entered manual AND autoscaling configurations

system-test/cluster.ts
```diff
@@ -0,0 +1,313 @@
...
+          await localCluster.getMetadata({});
+        assert.strictEqual(metadata[0].serveNodes, nodes);
+      }
+      beforeEach(async () => {
+        clusterId = generateId('cluster');
+        await createStandardNewInstance(clusterId, 2);
+        cluster = instance.cluster(clusterId);
+      });
+      it('Create an instance with clusters for manual scaling', async () => {
+        await checkMetadata(cluster, 2);
```
kolea2, line 78: should there be assertions here and below?

kolea2 (changes requested):

system-test/cluster.ts
danieljbruce: checkMetadata contains an assertion.

danieljbruce (commented):
Initial follow-up to first review.

system-test/cluster.ts
```diff
@@ -0,0 +1,313 @@
...
+      });
+      it('Create an instance and then create a cluster for manual scaling', async () => {
```
kolea2, line 80: nit - here and in all other `it(..` descriptions, I think they should be reworded slightly; for example, here I would change it to say:

`it('should create an instance and then create a cluster with manual scaling')`

Let me know your thoughts :)

danieljbruce: Very good point. I'll look at some of the other PRs and make sure that they have these descriptions as well.

src/utils/cluster.ts
```diff
@@ -0,0 +1,154 @@
...
+  static getClusterBaseConfig(
+    metadata: SetClusterMetadataOptions | BasicClusterConfig,
+    location: string | undefined,
+    name: string | undefined
+  ): google.bigtable.admin.v2.ICluster {
+    let clusterConfig;
+    if (
+      metadata.cpuUtilizationPercent ||
+      metadata.minServeNodes ||
+      metadata.maxServeNodes
```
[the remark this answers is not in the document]
danieljbruce: When creating instances we now validate inputs.

danieljbruce (commented):
Addressed two more outstanding comments

system-test/cluster.ts
```diff
@@ -0,0 +1,326 @@
...
+              location: 'us-west1-c',
+              nodes: 2,
+              minServeNodes: 3,
+            });
+            assert.fail();
+          } catch (e) {
+            assert.equal(e.message, ClusterUtils.allConfigError);
+          }
+        });
+        it('should throw an error providing all autoscaling configurations', async () => {
```
kolea2, line 134: should this say "should throw an error when missing some autoscaling configurations"?

system-test/cluster.ts
```diff
@@ -0,0 +1,326 @@
...
+
+      beforeEach(async () => {
+        const clusterId = generateId('cluster');
+        await createStandardNewInstance(clusterId, startingNodes);
+        cluster = instance.cluster(clusterId);
+      });
+
+      it('should change nodes for manual scaling', async () => {
+        const updateNodes = 5;
+        assert.notEqual(startingNodes, updateNodes);
```
kolea2, line 204: this check is comparing two int variables - if you want to have this, I would instead get the creater cluster's nodes, then compare the created cluster's nodes to the updateNodes value.

system-test/cluster.ts
```diff
@@ -0,0 +1,326 @@
...
+            await cluster.setMetadata({
+              nodes: 2,
+              minServeNodes: 3,
+            });
+            assert.fail();
+          } catch (e) {
+            assert.equal(e.message, ClusterUtils.allConfigError);
+          }
+        });
+        it('should throw an error providing all autoscaling configurations', async () => {
```
kolea2, line 254: i think this is the wrong description for this test?

kolea2 (changes requested):

system-test/cluster.ts
danieljbruce: `checkMetadata` does the second thing you mentioned though it checks `metadata.serveNodes` equals `updateNodes` rather than checking `metadata.serveNodes` does not equal `startingNodes`. This `assert.notEqual` is more of a maintainability thing to ensure the programmer doesn't set `updateNodes` to equal `startingNodes` and then have the test pass when `setMetadata` doesn't actually work properly. If you think it's confusing though then we can remove it.

system-test/cluster.ts
```diff
@@ -0,0 +1,326 @@
...
+            ...createClusterOptions,
+            nodes: minServeNodes,
+          },
+          true
+        );
+      });
+    });
+  });
+  describe('Update cluster', () => {
+    describe('Starting from manual scaling', () => {
```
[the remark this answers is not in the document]
danieljbruce: Done

system-test/cluster.ts
danieljbruce: Yeah. This is a confusing error. Related to the comment above. Changing this now.

system-test/cluster.ts
kolea2: I don't think this is providing much in terms of autoscaling assertions, so I would remove it in favor of the checks you already mentioned.

src/cluster.ts
```diff
@@ -690,35 +697,42 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
...
-        client: 'BigtableInstanceAdminClient',
-        method: 'updateCluster',
-        reqOpts,
-        gaxOpts: gaxOptions,
-      },
-      (err, resp) => {
-        callback(err, resp);
-      }
-    );
+    const setMetadataWithLocation = (location: string, name: string) => {
```
bcoe, line 707: I would remove this logic if we think it's currently not needed based on your testing.

src/utils/cluster.ts
```diff
@@ -0,0 +1,154 @@
...
+    location: string,
+    name: string
+  ): google.bigtable.admin.v2.ICluster {
+    const cluster: ICluster = Object.assign(
+      {},
+      this.getClusterBaseConfig(metadata, location, name),
+      metadata
+    );
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    delete (cluster as any).nodes;
```
bcoe, line 137: I don't think you should have to case to `any`, assuming that these are properties on `ICluster`.

bcoe (changes requested):

src/cluster.ts
bcoe: I would not consider switching from required to optional for TypeScript breaking 👍

bcoe (approved):

kolea2 (approved):

PR #1090, opened 2022-05-19
feat: Errors from gax layer

danieljbruce:
This PR contains tests for ensuring that errors which pass through the gax layer are presented to the user properly using unit tests from a grpc service.


test/util/mock-server.ts
```diff
@@ -0,0 +1,65 @@
...
+      callback();
+    });
+    netServer.once('listening', () => {
+      // close the server if listening doesn't fail
+      netServer.close();
+    });
+    // netServer.listen(port);
+    netServer.listen(`localhost:${port}`);
+  }
+  describe('Ensure server shuts down properly when destroyed', () => {
```
danieljbruce, line 51: If I add a before hook with the following in it:

`checkPort(inputPort, false, done);`

the `assert.notEqual(err.code, 'EADDRINUSE');` fails, but I am not sure why at this time.

test/util/mock-server.ts
```diff
@@ -0,0 +1,49 @@
...
+import {MockServer} from '../../src/util/mock-servers/mock-server';
+import * as assert from 'assert';
+
+const tcpPortUsed = require('tcp-port-used');
+
+describe('Bigtable/Mock-Server', () => {
+  const inputPort = '1234';
+  let server: MockServer;
+  function checkPort(port: string, inUse: boolean, callback: () => void) {
+    tcpPortUsed.check(parseInt(port), 'localhost').then((isInUse: boolean) => {
```
bcoe, line 29: my advice would be to always use an async/await interface when possible, I think you could replace this with:

```
await checkPort()
```

bcoe (changes requested):
Left a nit, I would always opt for an async/await appraoch, rather than Promise.then these days.

test/util/mock-server.ts
danieljbruce: Done

danieljbruce: It actually looks like this now:

```
  async function checkPort(port: string, inUse: boolean, callback: () => void) {
    const isInUse: boolean = await tcpPortUsed.check(
      parseInt(port),
      'localhost'
    );
    assert.strictEqual(isInUse, inUse);
    callback();
  }
```

bcoe (approved):

PR #1092, opened 2022-06-01
test: Add autogenerated tests to prepare for a refactor PR

danieljbruce:
The only thing this PR does is add tests with automatically generated snapshots that capture the data that will be sent to the generated layer. There should be no functionality changes to code in the `src` folder.

The main reason for creating these autogenerated snapshots is that next I will open a technical debt PR which refactors code and we want to be sure that when we do that refactor, we don't actually change what the code does. Comparing variable values to these previously generated snapshots will allow us to do this to ensure data being passed to the generated layer remains unchanged. If there are subtle changes then we will see the difference in the snapshot in the tech debt PR and we can check with team members that manage code server side to ensure the changes won't break anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
     "pack-n-play": "^1.0.0-2",
     "proxyquire": "^2.0.0",
     "sinon": "^14.0.0",
+    "snap-shot-it": "^7.9.1",
     "ts-loader": "^9.0.0",
     "typescript": "^4.6.4",
     "uuid": "^8.0.0",
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "system-test": "mocha build/system-test --timeout 600000",
     "pretest": "npm run compile",
     "test": "c8 mocha build/test",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -86,6 +86,7 @@
     "pack-n-play": "^1.0.0-2",
     "proxyquire": "^2.0.0",
     "sinon": "^14.0.0",
+    "snap-shot-it": "^7.9.1",
```
bcoe, line 89: I would add an additional npm script that allows you to take snapshots retake snapshots, otherwise if the structure of responses changes in the future someone needs to read the docs for `snap-shot-it`, something like:

```
{"scripts: {
  "test:snap": "COMMAND TO TAKE SNAPSHOT"
}}
```

bcoe (approved):
This seems like a great use of snapshots, I left one suggestion that I think will make maintenance easier.

package.json
danieljbruce: Script added.

danieljbruce (commented):
First PR follow-up

kolea2 (approved):

PR #1093, opened 2022-06-01
feat: Remove need to pass location parameter along

danieljbruce:
This PR is the 2nd step towards refactoring code pertaining to autoscaler and cluster update work. This should be merged after the following PR:

https://github.com/googleapis/nodejs-bigtable/pull/1092

### DIFF
--- a/src/cluster.ts
+++ b/src/cluster.ts
@@ -80,6 +80,7 @@ export interface SetClusterMetadataOptions {
   minServeNodes?: number;
   maxServeNodes?: number;
   cpuUtilizationPercent?: number;
+  location?: string;
 }
 export type SetClusterMetadataCallback = GenericOperationCallback<
   Operation | null | undefined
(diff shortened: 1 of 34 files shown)


src/cluster.ts
```diff
@@ -693,7 +693,7 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
    * region_tag:bigtable_api_cluster_set_meta
    */
   setMetadata(
-    metadata: SetClusterMetadataOptions,
+    metadata: BasicClusterConfig,
```
danieljbruce, line 696: This is not a breaking change because `BasicClusterConfig` includes all the parameters that `SetClusterMetadataOptions` includes. We need to do this to allow the user to pass in `location` parameter, which shouldn't be required, but seems to be accepted in other codebases when we update clusters.

bcoe: I believe this will be a compilation error for upstream users because `location` is `require3d` on `BasicClusterConfig`, what I might do instead is add `location` as an optional parameter on `SetClusterMetadataOptions`.

bcoe (changes requested):

src/cluster.ts
danieljbruce: Great point. By adding this optional parameter to the interface it also means we make fewer changes too which is good because it keeps this PR small. Done.

danieljbruce (commented):
Comments addressed.

bcoe (approved):

PR #1094, opened 2022-06-01
refactor: Pull code to compute location into one function

danieljbruce:
This pull request is dependent on https://github.com/googleapis/nodejs-bigtable/pull/1093. After that PR and its predecessor are merged this one will just be a PR where we pull repeated functionality from two functions into a single utility function.


test/index.ts
```diff
@@ -493,7 +493,10 @@ describe('Bigtable', () => {
     });
 
     it('should respect the clusters option', done => {
-      const fakeLocation = 'a/b/c/d';
+      const fakeLocation = Cluster.getLocation_(
```
bcoe, line 496: I'm surprised that tests had to be updated, since it looks like this is just a refactor. Does the value of `fakeLocation` actually matter?

It may be better leave the existing tests as is, since it makes it more clear that this is simply a refactor.

src/utils/cluster.ts
```diff
@@ -85,6 +91,21 @@ export class ClusterUtils {
     return updateMask;
   }
 
+  static getClusterBaseConfigWithFullLocation(
```
bcoe, line 94: I think we should try to add a test that specifically asserts against this expected behaviour:

```
it('populates full location path if location parameter provided')
```

What format is it expected that metadata.location is in? Is this value provided by a user.

bcoe (changes requested):
If I understand correctly, this is a cleanup and refactor of existing behaviour, vs., a new feature?

I would title the PR in this case:

`refactor: pull code to compute location into one function`.

---

I like this cleanup 🥳 , but left a couple thoughts.

src/utils/cluster.ts
bcoe: I would be tempted to add a test for the user behavior that would cause these different code paths to be hit.

One test for  `location` being undefined, one test for `location` being set.

test/index.ts
danieljbruce: It looks like we decided just to get rid of the mocks that weren't being used.

bcoe (approved):

PR #1103, opened 2022-06-20
fix: Integration test close function again

danieljbruce:
This fixes an issue with the close function where promises of promises were once returned, but now we simply return a promise so that when we use `await` then the code is executed properly.


src/index.ts
```diff
@@ -930,7 +930,7 @@ export class Bigtable {
  * that a callback is omitted.
  */
 promisifyAll(Bigtable, {
-  exclude: ['instance', 'operation', 'request'],
+  exclude: ['close', 'instance', 'operation', 'request'],
```
danieljbruce, line 933: In the absence of this change, the test just seems to hang forever.

danieljbruce (commented):
A comment to clarify the original problem.

src/index.ts
bcoe: `promisifyAll` turns a `cb` based function into a promise based function, it looks like close was already a promise, and did not expect a `callback` as an argument:

```js
close(): Promise<void[]>
```

bcoe (approved):

src/index.ts
danieljbruce: Yes. I am making sure the reader of this PR knows why this change is necessary.

PR #1117, opened 2022-06-29
feat: Generate end to end tests for retries

danieljbruce:
This PR is a work in progress, but contains a test framework that is entirely necessary for end to end testing when we decide to move retry logic into the gax layer. The idea is that we collect snapshots to ensure that the incoming requests match the snapshots that are recorded.


src/util/mock-servers/service-testers/service-handlers/implementation/read-rows-handler.ts
```diff
@@ -0,0 +1,126 @@
...
+    responses: ReadRowsResponse[],
+    message?: any
+  ) {
+    super(service, endpoint);
+    this.responses = responses;
+    this.message = Object.assign({}, message);
+  }
+
+  // TODO: Create interface for this.
+  callHandler(call: any) {
```
bcoe, line 61: It would b good to use a better type for `call`, you could potentially just create your own interface:

```js
interface {
  emit: () => {}
}
```

But I wouldn't be surprised if there's a type available in the `grpc` library.

src/util/mock-servers/service-testers/service-handlers/implementation/read-rows-handler.ts
```diff
@@ -0,0 +1,126 @@
+// Copyright 2022 Google LLC
```
bcoe, line 1: I should have caught this in the last review, since these servers (I believe) will only be used in testing, I'd be tempted to have them in `test/util`, rather than `src/util`. I think of `src` as being components of the library itself, vs., components used during testing.

src/util/mock-servers/service-testers/service-handlers/implementation/read-rows-handler.ts
```diff
@@ -0,0 +1,126 @@
...
+//
+// ** This file is automatically generated by gapic-generator-typescript. **
+// ** https://github.com/googleapis/gapic-generator-typescript **
+// ** All changes to this file may be overwritten. **
+
+import {SameCallHandler} from './same-call-handler';
+import {MockService} from '../../../mock-service';
+import {Mutation} from '../../../../../mutation';
+
+function rowResponse(rowKey: {}) {
```
bcoe, line 23: Is there going to be a different "implementation" file for different test cases, e.g, ensuring that streaming errors are gracefully recovered from?

If so, I would be tempted to name the implementation file after the specific behaviour it's a test for. Then you'll have a bunch of implementation files for a bunch of different regression tests.

test/server/read-rows.ts
```diff
@@ -0,0 +1,320 @@
...
+
+  function getStreamTester(
+    serviceHandler: ServiceHandler,
+    opts?: GetRowsOptions
+  ) {
+    const streamFetcher = new ReadRowsFetcher(table, opts);
+    return new StreamTester(serviceHandler, streamFetcher);
+  }
+  /*
+  describe('with a mock server that always sends an error back', () => {
```
bcoe, line 62: remember to comment these out before you land.

src/util/mock-servers/service-testers/stream-tester.ts
```diff
@@ -0,0 +1,59 @@
...
+export class StreamTester {
+  serviceHandler: ServiceHandler;
+  streamFetcher: StreamFetcher;
+
+  constructor(serviceHandler: ServiceHandler, streamFetcher: StreamFetcher) {
+    this.serviceHandler = serviceHandler;
+    this.streamFetcher = streamFetcher;
+  }
+
+  checkSnapshots(callback: () => void): void {
```
bcoe, line 35: I like that you've pulled the snapshotting into a helper 😄

src/util/mock-servers/service-testers/stream-tester.ts
```diff
@@ -0,0 +1,59 @@
+// Copyright 2022 Google LLC
```
bcoe, line 1: I would have anything related to testing in the `test` folder rather than `src` (which should exclusively be application logic).

danieljbruce:
Readrows retries has been merged so this is not necessary anymore.

PR #1118, opened 2022-06-30
refactor: first refactor of createReadStream

danieljbruce:
This is the first of many PRs to refactor createReadStream. We need to reduce the size of this function so that we can work with it and so that adding retry features is easier.

API surface affected:
- table.createReadStream:
- https://github.com/googleapis/nodejs-bigtable/blob/main/src/table.ts#L734
- incoming parameters to `request` function in index.ts:
- https://github.com/googleapis/nodejs-bigtable/blob/main/src/index.ts#L779

For tests that look at the API surface:
https://github.com/googleapis/nodejs-bigtable/blob/main/test/table.ts#L539
https://github.com/googleapis/nodejs-bigtable/blob/main/system-test/read-rows.ts#L154

In this PR, we pull out code to calculate the `ranges` variable in `createReadStream` into a separate function. The main motivation for this besides reducing the `createReadStream` function size is to take code, which calculates `ranges` initially and move it into a separate function since this initial calculation has nothing to do with what happens in the rest of the function after that. This conceptually tells the code reader that only `options` is needed to calculate ranges initially and that all code used to calculate `ranges` initially can be ignored if the reader is only interested in what happens in the rest of the function.

So.... 
https://github.com/googleapis/nodejs-bigtable/pull/1118/files#diff-6400d8602f0855266c743f4fb031e97ac9cf07f9dcddfba03833b6f04d547a5eL750 to https://github.com/googleapis/nodejs-bigtable/pull/1118/files#diff-6400d8602f0855266c743f4fb031e97ac9cf07f9dcddfba03833b6f04d547a5eL780)

should just be a cut/paste of
https://github.com/googleapis/nodejs-bigtable/pull/1118/files#diff-2490acae1e166b5d2477465b3ee761d0857374c409d215d0363d626c665e80d6R20 to
https://github.com/googleapis/nodejs-bigtable/pull/1118/files#diff-2490acae1e166b5d2477465b3ee761d0857374c409d215d0363d626c665e80d6R48

The only other two minor changes are the changes to the test stub and the change to `createPrefixRange`. The test stub changes which include all the changes in `test/table.ts` are necessary as a result of pulling out `ranges` into a utility function and the way that sinon works in the tests.  `table.createPrefixRange` delegates its functionality to `TableUtils.createPrefixRange` which does the exact same thing `table.createPrefixRange` did (we are just cut/pasting the code), but is something we need to do because of sinon stubs. Basically, sinon gets in the way of just being able to pull code out because it mocks `createPrefixRange` from `Table`, but `TableUtils` has a different copy of `Table` so while a test for prefixes should work, in this case the test needs to be slightly modified because it also demands that everything happens in `Table` which shouldn't be necessary. But if this is undesirable then we can consider using a private method instead of a utility class.


igorbernstein2 (changes requested):
I think this needs a bit more  refactoring. Currently this creates a circular dependency between TableUtils and Table, which is a bad code smell

danieljbruce:
More refactoring is done in the next two subsequent PRs.

danieljbruce:
If you want to see a larger refactor then take a look at this PR:

https://github.com/googleapis/nodejs-bigtable/pull/1121

igorbernstein2 (approved):
lgtm

PR #1129, opened 2022-07-13
test: Add a logger to print the instance id.

### LINKED ISSUE #1128: Add a log in the asserts for system tests on clusters

danieljbruce:
This PR is used to help with debugging cluster issues.

Fixes #1128


kolea2 (approved):
please rename PR to "test: ..." prefix

system-test/cluster.ts
```diff
@@ -67,6 +67,7 @@ describe('Cluster', () => {
         time_created: Date.now(),
       },
     });
+    console.log(`Test Instance Id: ${instanceId}`);
     await operation.promise();
```
shantstepanian, line 71: Can we log the cluster we query in checkMetadata (e.g. method starting line 40)?

system-test/cluster.ts
```diff
@@ -67,6 +67,7 @@ describe('Cluster', () => {
         time_created: Date.now(),
       },
     });
```
shantstepanian, line 69: Is it also possible to add more IDs on the cluster that is being compared in the assertion? i.e. to easily show which cluster is mismatching the ID

```
  1) Cluster << would be nice to have the cluster ID here >>
       Update cluster
         Starting from autoscaling
           should change cluster to manual scaling:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

3 !== 5

      + expected - actual

      -3
      +5

      at checkMetadata (build/system-test/cluster.js:31:16)
          -> /workspace/system-test/cluster.ts:42:12
      at async Context.<anonymous> (build/system-test/cluster.js:273:17)
          -> /workspace/system-test/cluster.ts:306:9
```

danieljbruce: Hmm. I'm not sure how to make a dynamic title in the mocha framework, but we can definitely print out the cluster id.

system-test/cluster.ts
danieljbruce: Sure. We can log the cluster id.

system-test/cluster.ts
shantstepanian: No worries, yea, whatever is easier to do, as I'm not too familiar w/ Typescript and its frameworks. So long as we see the cluster ID logged earlier in the log that corresponds w/ the test failure, we should be alright

PR #1153, opened 2022-09-09
feat: Add feature for copying backups

danieljbruce:
This PR includes changes in the veneer layer that implement the new copy backup feature which allows users to copy a backup of a table to a cluster of their choice.

A config interface is introduced which is used to specify details about the copied backup. On a backup object, a copy function is added which copies the backup to a destination specified by the config passed into the function parameter. Code that parses expiry time is reused for the copy backup function.

Unit tests are added to make sure source and destination information get passed along to the Gapic layer. System tests are added to check that copying a backup works when copying to the same cluster, different cluster and different project. System tests are also added to be sure that expiry times can be passed in using different formats and tests are added so that we can confirm restoring the copied backup works properly.


system-test/bigtable.ts
```diff
@@ -1417,6 +1417,66 @@ describe('Bigtable', () => {
 
       Object.keys(policy).forEach(key => assert(key in updatedPolicy));
     });
+    describe('copying backups', () => {
+      async function testCopyBackup(backup: Backup, newBackup: Backup) {
```
kolea2, line 1421: I would like to see some more assertions here. I am thinking something along the lines of:
1. copy the backup, wait for operation to complete
2. list backups
3. assert the copied backup is in the list, and assert it has correct name, expire time, etc

test/backup.ts
```diff
@@ -223,6 +224,38 @@ describe('Bigtable/Backup', () => {
     });
   });
 
+  describe('copy', () => {
+    it.only('should correctly copy backup from the Cluster', done => {
+      const backupId = generateId('backup');
+      const newBackupId = generateId('backup');
+      const oldBackup = new Backup(CLUSTER, backupId);
```
kolea2, line 231: nit - for clarity, I would rename these `backup` and `copiedBackup`

kolea2 (changes requested):

system-test/bigtable.ts
danieljbruce: list backups contains name and id so I made an assert that compares those two and ensures exactly one new backup gets created that matches.

test/backup.ts
danieljbruce: Done

src/utils/cluster.ts
```diff
@@ -163,4 +165,10 @@ export class ClusterUtils {
       updateMask: {paths: this.getUpdateMask(metadata)},
     };
   }
+
+  static setExpiryTime(backup: ModifiableBackupFields) {
```
kolea2, line 169: rename - `setBackupExpiryTime`

danieljbruce: Done

danieljbruce:
https://github.com/googleapis/nodejs-bigtable/pull/1372 will address the failing test in the system tests.

src/utils/cluster.ts
```diff
@@ -163,4 +165,10 @@ export class ClusterUtils {
       updateMask: {paths: this.getUpdateMask(metadata)},
     };
   }
+
+  static setBackupExpiryTime(backup: ModifiableBackupFields) {
+    if (backup.expireTime instanceof Date) {
```
kolea2, line 170: should we throw if the condition is not met here?

danieljbruce: No. This function is just pulling out the exact code that is being used in createBackup at https://github.com/googleapis/nodejs-bigtable/pull/1153/files#diff-9ce792749808a4cb9055a011d90ec2c35709e9deeee913947cb43f7ca394c911L317-L321 as this code is now used twice to parse a user provided date. backup.expireTime is of BackupTimestamp type and this means it can be a Date, PreciseDate or google.protobuf.ITimestamp as you can see by the definition in https://github.com/googleapis/nodejs-bigtable/blob/d30e5133a55e1b8f269b3aa6a69b49415361ba34/src/backup.ts#L44. Throwing an exception here would not let users provide a google.protobuf.ITimestamp value for their request.

system-test/bigtable.ts
```diff
@@ -1414,6 +1421,268 @@ describe('Bigtable', () => {
...
+        // Create the copy and test the copied backup
+        await testCopyBackup(
+          backup,
+          {
+            cluster: new Cluster(INSTANCE, destinationClusterId),
+            id: generateId('backup'),
+            expireTime: copyExpireTime,
+          },
+          INSTANCE
+        );
```
cindy-peng, line 1594: Delete the instance here after testing?

danieljbruce: This instance is already deleted in https://github.com/googleapis/nodejs-bigtable/blob/d30e5133a55e1b8f269b3aa6a69b49415361ba34/system-test/bigtable.ts#L98-L115.

src/backup.ts
```diff
@@ -37,6 +37,15 @@ import {
 } from './cluster';
 import {CallOptions, LROperation, Operation, ServiceError} from 'google-gax';
 import {Instance} from './instance';
+import {ClusterUtils} from './utils/cluster';
+
+export type CopyBackupResponse = GenericBackupPromise<Operation>;
+export type CopyBackupCallback = GenericBackupCallback<Operation>;
+export interface CopyBackupConfig extends ModifiableBackupFields {
+  cluster: Cluster;
+  gaxOptions?: CallOptions;
```
kolea2, line 46: rename to `callOptions`?

src/backup.ts
```diff
@@ -243,6 +253,45 @@ Please use the format 'my-backup' or '${cluster.name}/backups/my-backup'.`);
     });
   }
 
+  /**
+   * Copy this backup
```
kolea2, line 257: "Copy this backup..." to where? please add to the documentation

src/backup.ts
```diff
@@ -243,6 +253,45 @@ Please use the format 'my-backup' or '${cluster.name}/backups/my-backup'.`);
...
+   *
+   * @param {CopyBackupConfig} [config] The config that specifies all of the
+   * information about the destination backup.
```
kolea2, line 260: what is the "destination backup"? Is that the backup that will be copied, or the backup to be copied? Please add some more info here.

src/utils/cluster.ts
kolea2: I think this method is misleading then. If I call `setBackupExpiryTime` and I pass in a non Date object (like an `ITimestamp` as you mention), there is no information to me that I am sending in something invalid, and nothing is getting set in that case. Can we rename the method to something more specific to the use case, or not extract it out?

kolea2 (commented):
added some initial comments, need to come back to review the tests

src/backup.ts
danieljbruce: I added a paragraph here. Some functions in the client library have a long explanation of what they do and some don't, but I extended this one so its purpose is more clear.

src/backup.ts
danieljbruce: I added some more information here about this parameter.

src/utils/cluster.ts
danieljbruce: I changed the name to formatBackupExpiryTime. We should still extract it out so that we don't repeat code. Feel free to let me know if you have a different preferred name in mind.

kolea2: LGTM

system-test/bigtable.ts
```diff
@@ -1414,6 +1421,268 @@ describe('Bigtable', () => {
...
+          expireTime: sourceExpireTime,
+        });
+        {
+          await op.promise();
+          // Check the expiry time.
+          await backup.getMetadata();
+          assert.deepStrictEqual(backup.expireDate, sourceExpireTime);
+        }
+        // Create client, instance, cluster for second project
+        const bigtable = new Bigtable(
```
kolea2, line 1607: nit - I would rename this to something that signifies it's the second project. Maybe `bigtableSecondaryProject`?

system-test/bigtable.ts
```diff
@@ -1414,6 +1421,268 @@ describe('Bigtable', () => {
...
+          process.env.GCLOUD_PROJECT2
+            ? {projectId: process.env.GCLOUD_PROJECT2}
+            : {}
+        );
+        const instanceId = generateId('instance');
```
kolea2, line 1612: nit - I think we can inline this to the line below

system-test/bigtable.ts
```diff
@@ -1414,6 +1421,268 @@ describe('Bigtable', () => {
...
+        const instance = bigtable.instance(instanceId);
```
kolea2, line 1613: as above, I would rename this to show it's the second project instance

system-test/bigtable.ts
kolea2: or `bigtableDestinationProject`?

danieljbruce: Done

system-test/bigtable.ts
danieljbruce: Done

system-test/bigtable.ts
danieljbruce: Done

kolea2 (approved):

system-test/bigtable.ts
```diff
@@ -1414,6 +1421,269 @@ describe('Bigtable', () => {
...
+        const newBackups = backupsAfterCopy.filter(
+          backup => !backupIdsBeforeCopy.includes(backup.id)
+        );
+        assert.strictEqual(newBackups.length, 1);
+        const [fetchedNewBackup] = newBackups;
+        // Ensure the fetched backup matches the config
+        assert.strictEqual(fetchedNewBackup.id, id);
+        assert.strictEqual(fetchedNewBackup.name, backupPath);
+        // Delete the copied backup
+        await config.cluster.backup(fetchedNewBackup.id).delete();
```
leahecole, line 1477: If this test fails, do you need to worry about backups being leftover? Or is the entire resource destroyed no matter what?

test/backup.ts
```diff
@@ -223,6 +225,68 @@ describe('Bigtable/Backup', () => {
     });
   });
 
+  describe('copy', () => {
+    beforeEach(() => {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
leahecole, line 231: A few questions
1) What's the reasoning behind the eslint-disable-next-line plus the ts-ignore? In my mind, they kind of cancel each other out, but I'm sure that there's a deeper reason I just don't know here 🙂 
2) Is there a way we can not ignore the linter here?
3) if we're ignoring the linter, can we get away with only the eslint-disable-next-line and whatever the error is it's throwing? Or some kind of ts-expect-error? (I can't remember our exact settings for both eslint and the ts checker)

src/cluster.ts
```diff
@@ -314,11 +314,7 @@ Please use the format 'my-cluster' or '${instance.name}/clusters/my-cluster'.`);
       },
     };
 
-    if (reqOpts.backup.expireTime instanceof Date) {
-      reqOpts.backup.expireTime = new PreciseDate(
-        reqOpts.backup.expireTime
-      ).toStruct();
-    }
+    ClusterUtils.formatBackupExpiryTime(reqOpts.backup);
```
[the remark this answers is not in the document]
danieljbruce: Yes!

system-test/bigtable.ts
```diff
@@ -1414,6 +1421,269 @@ describe('Bigtable', () => {
 
       Object.keys(policy).forEach(key => assert(key in updatedPolicy));
     });
+    describe('copying backups', () => {
+      const sourceExpireTimeMilliseconds =
+        PreciseDate.now() + (8 + 300) * 60 * 60 * 1000;
```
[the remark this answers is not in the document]
danieljbruce: My local system tests are hitting quotas so it's hard to recall exactly by testing it out. But if I recall correctly, the create backup expire time has to be sufficiently far ahead of the current time and the copied backup expire time has to be sufficiently ahead of the create backup expire time to avoid server errors specific to the copy backup feature. I added a comment for this.

system-test/bigtable.ts
danieljbruce: I added better cleanup for the backups both created from scratch and copied.

test/backup.ts
danieljbruce: These comments may have been leftover from a copy/paste. They are not needed anymore.

system-test/bigtable.ts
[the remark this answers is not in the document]
leahecole: Thanks for the comment! Makes sense as to why it needs to be a large number. My remaining confusion is why the arithmetic needs to be there rather than adding a fixed integer to PreciseDate.now()?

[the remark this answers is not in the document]
danieljbruce: To your point, I think we can change this to `308 * 60 * 60 * 1000` and `608 * 60 * 60 * 1000`, but I want to keep the 60 * 60 * 1000 in there because it is beneficial to the reader to know that 308 is the number of hours (milliseconds/second * seconds/minute * minutes/hour).

system-test/bigtable.ts
```diff
@@ -1414,6 +1421,296 @@ describe('Bigtable', () => {
 
       Object.keys(policy).forEach(key => assert(key in updatedPolicy));
     });
+    describe('copying backups', () => {
+      // The server requires the source backup time to be sufficiently ahead
+      // of the current time to avoid an error.
+      // The server requires the copy backup time to be sufficiently ahead of
+      // the create backup time to avoid an error.
+      const sourceExpireTimeMilliseconds =
+        PreciseDate.now() + (8 + 300) * 60 * 60 * 1000;
```
leahecole, line 1430:
```suggestion
      // The server requires the copy backup time to be sufficiently ahead of
      // the create backup time to avoid an error.
      // Set it to 308 hours ahead
      const sourceExpireTimeMilliseconds =
        PreciseDate.now() + (8 + 300) * 60 * 60 * 1000;
```

system-test/bigtable.ts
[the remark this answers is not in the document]
leahecole: ah, thank you! I agree with the 60 * 60 * 1000 making sense. I made a suggestion to modify the comment

system-test/bigtable.ts
```diff
@@ -1414,6 +1421,296 @@ describe('Bigtable', () => {
...
+      const sourceExpireTime = new PreciseDate(sourceExpireTimeMilliseconds);
+      const copyExpireTimeMilliseconds =
+        PreciseDate.now() + (8 + 600) * 60 * 60 * 1000;
```
leahecole, line 1433:
```suggestion
      // 608 hours ahead of now, 300 hours ahead of sourceExpireTimeMilliseconds
      const copyExpireTimeMilliseconds =
        PreciseDate.now() + (8 + 600) * 60 * 60 * 1000;
```

leahecole (approved):
approving once suggestions with comments are added and with the assumption that the failing conformance tests are unrelated

system-test/bigtable.ts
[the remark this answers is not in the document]
danieljbruce: Suggestions applied

PR #1160, opened 2022-09-16
fix: Test is less restrictive to allow changes

danieljbruce:
Sometimes cluster metadata returned changes slightly and this causes tests to fail. We want to make sure that these tests only check for what really needs to be there.


Mariatta (approved):
LGTM

PR #1174, opened 2022-09-20
fix: Remove docs test for now

danieljbruce:
This change removes the docs test which is blocking other PRs for the time being.


danieljbruce:
Add this later if docs tests block again.

Issue #991, opened 2022-01-20
Bubble up 'metadata' event for mutate request


mutianf:
mutate request should emit 'metadata' event in the manual layer so applications can get useful metrics from response headers like gfe 'server-timing'

danieljbruce:
I looked at what I get back from an error on the following test when using mutate instead of insert:

https://github.com/googleapis/nodejs-bigtable/blob/main/test/errors.ts#L105

It appears that the error includes metadata:

err.metadata = `{"grpc-server-stats-bin":[{"type":"Buffer","data":[0,0,116,73,159,3,0,0,0,0]}]}`

Do you mean `metadata` should be available when data is sent back as well?

danieljbruce:
Clarification: the bug is not talking about the metadata in the error. It is talking about the metadata grpc response which indeed includes headers like `'server-timing'`.

danieljbruce:
https://github.com/googleapis/nodejs-bigtable/pull/1178 is a possible fix.

danieljbruce:
https://github.com/googleapis/nodejs-bigtable/pull/1180 contains the changes that include making metadata available to the user, but it seems like we are satisfied with providing it on the returned stream.

danieljbruce:
https://github.com/googleapis/nodejs-bigtable/pull/1178 merged. Let me know if any more changes are needed.

PR #1178, opened 2022-09-23
fix: pipe metadata along

danieljbruce:
Work towards https://github.com/googleapis/nodejs-bigtable/issues/991


system-test/bigtable.ts
```diff
@@ -698,6 +698,25 @@ describe('Bigtable', () => {
...
+            data: {
+              nonExistentColumnFamily: {
+                jadams: 1,
+              },
+            },
+          },
+        ];
+        try {
+          await TABLE.insert(rows);
+          assert.fail('Inserts should not work on non-existent column family.');
```
ruyadorno, line 714: `assert.fail` is always going to be throwing an `AssertionError` thus always triggering the following `catch` block

```suggestion
          // Inserts should not work on non-existent column family.
          await TABLE.insert(rows);
```

That message will live better as a comment or something similar

system-test/bigtable.ts
```diff
@@ -698,6 +698,25 @@ describe('Bigtable', () => {
...
+        } catch (e: any) {
+          assert(e.metadata);
```
ruyadorno, line 716: I believe it's better to try and match the exact expected values of this metadata object here rather than just a blank non-falsy assertion.

ruyadorno (commented):
This is looking like a good start. 👍  I'm curious to understand better what is the mechanism that makes it so that the emitted `metadata` makes it way into the `mutationErrorsByEntryIndex` map.

Given that the test is always throwing at the moment I'm not entirely convinced that the implementation is working as intended. Let's start with improving the tests and feel free to elaborate a bit more on the mechanism!

Having working tests that assert the exact expected values for the metadata will make me more confident on this PR 😊

system-test/bigtable.ts
danieljbruce: Are you sure? I want the test to fail if the insert succeeds. In this case it will go to the catch block, but then the other assertion will fail.

system-test/bigtable.ts
danieljbruce: Added a check for two of the metadata properties that don't change.

src/table.ts
```diff
@@ -1612,6 +1614,9 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
             mutationErrorsByEntryIndex.set(originalEntriesIndex, errorDetails);
           });
         })
+        .on('metadata', metadata => {
```
mutianf, line 1617: I don't think we need to save the metadata for now. When we were debugging the customer's issue, we wanted to see the value of server-timing that's published by GFE. In the future we'll also add this as a metric. But for now I think the changes in tables.ts could be reverted.

mutianf (changes requested):

src/table.ts
danieljbruce: Ok. I guess the original intention was to make it so that the metadata is available on the stream, but doesn't need to reach the user?

mutianf (approved):

PR #1217, opened 2023-01-26
test: Add await

danieljbruce:
This should fix a flakey test issue where we wait for an instance to be created.

### DIFF
--- a/samples/test/filters.js
+++ b/samples/test/filters.js
@@ -19,11 +19,13 @@ const snapshot = require('snap-shot-it');
 const {assert} = require('chai');
 const {describe, it, before, after} = require('mocha');
 const cp = require('child_process');
-const {obtainTestInstance} = require('./util');
+const {obtainTestInstance, createTestInstance} = require('./util');
 
 const execSync = cmd => cp.execSync(cmd, {encoding: 'utf-8'});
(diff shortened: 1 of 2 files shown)


kolea2 (dismissed):
please prefix this with "test" rather than "fix" so it does not show up in the release notes as a bugfix

danieljbruce:
> please prefix this with "test" rather than "fix" so it does not show up in the release notes as a bugfix

Done!

samples/test/filters.js
```diff
@@ -19,12 +19,14 @@ const snapshot = require('snap-shot-it');
...
 const {describe, it, before, after} = require('mocha');
 const cp = require('child_process');
-const {obtainTestInstance} = require('./util');
+const {obtainTestInstance, createTestInstance} = require('./util');
 
 const execSync = cmd => cp.execSync(cmd, {encoding: 'utf-8'});
 const runId = uuid.v4().split('-')[0];
 const TABLE_ID = `mobile-time-series-${runId}`;
 
+createTestInstance();
```
bcoe, line 28: This will create an `unhandledRejection`, so I think we shouldn't do this in the top level.

danieljbruce:
Closed in favour of moving `obtainTestInstance` instead.

PR #1219, opened 2023-01-27
Simple start and end example

danieljbruce:
Not to be merged. Just console logs to show program flow.


danieljbruce:
Tests recorded and observations discussed. This is of no further value.

PR #1225, opened 2023-01-30
test: Move obtain test instance

danieljbruce:
This PR will reduce the number of failures due to flakey tests by fixing the flakey tests.

### DIFF
--- a/samples/test/reads.js
+++ b/samples/test/reads.js
@@ -24,13 +24,16 @@ const execSync = cmd => cp.execSync(cmd, {encoding: 'utf-8'});
 const TABLE_ID = `mobile-time-series-${uuid.v4()}`.substr(0, 30); // Bigtable naming rules
 
 describe('reads', async () => {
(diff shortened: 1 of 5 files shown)


samples/test/reads.js
```diff
@@ -24,13 +24,16 @@ const execSync = cmd => cp.execSync(cmd, {encoding: 'utf-8'});
...
 describe('reads', async () => {
-  const instance = await obtainTestInstance();
-  const INSTANCE_ID = instance.id;
+  let INSTANCE_ID;
   let table;
-  const TIMESTAMP = new Date(2019, 5, 1);
-  TIMESTAMP.setUTCHours(0);
 
   before(async () => {
+    const instance = await obtainTestInstance();
```
bcoe, line 31: This change makes more sense to me. It's a good cautionary tale that we should avoid `await` outside the `before`, `after`, `beforeEach`, `afterEach` blocks.

bcoe (approved):

PR #1230, opened 2023-02-01
chore(testproxy): read row should return a single row

danieljbruce:
To get one of the specific test cases passing for readRow, we need a slight fix to the way returned data is parsed.

### DIFF
--- a/testproxy/services/read-row.js
+++ b/testproxy/services/read-row.js
@@ -27,10 +27,11 @@ const readRow = ({clientMap}) =>
     const table = getTableInfo(bigtable, tableName);
     const row = table.row(rowKey);
     const res = await row.get(columns);
+    const firstRow = getRowResponse(res[0]);
 
     return {
       status: {code: grpc.status.OK, details: []},
-      row: res.map(getRowResponse),
+      row: firstRow,
     };
   });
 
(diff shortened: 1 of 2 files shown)


ruyadorno:
it'd be nice for the commit message to mention that this is related to the test proxy and not the implementation itself 😊 maybe prepend `testproxy: ` to the title?

ruyadorno (commented):
LGTM 👍

danieljbruce:
> it'd be nice for the commit message to mention that this is related to the test proxy and not the implementation itself 😊 maybe prepend `testproxy: ` to the title?

Unfortunately for the conventional commits to pass we must prefix with a specific word. Probably `test:` in this case.

https://www.conventionalcommits.org/en/v1.0.0/

Maybe set to `test: testproxy: Slight change in data returned by readRow`?

ruyadorno:
maybe we can use: `test(testproxy): read row should return a single row` so that it's also a bit more descriptive of the change

ruyadorno:
for reference, I used `chore` in the commit I introduced the testproxy, ref: 0b3d996a667029cf2259dcc8f22d0dc537bf33f0

ruyadorno (approved):

PR #1231, opened 2023-02-03
fix: Out of order read rows fix

danieljbruce:
Client changes:

Emits an error if the chunk transformer receives keys that are out of order. The chunk transformer is a stream that data passes through before it reaches the stream that has the handlers for errors and retries. 

Also pulls some comparison functions into a utility file because now those functions are used by the chunk transformer to check to see if rows are in order. They are still used by the `table.ts` file, but are available in both places now.

Test proxy changes:

Makes it so that when the client emits an error as a result of rows that are out of order, that the error is sent back to the test runner as an rpc message instead of an rpc error. This is done to stay consistent with the test proxy in Java and allows the TestReadRows_NoRetry_OutOfOrderError test case to pass.

### DIFF
--- a/src/chunktransformer.ts
+++ b/src/chunktransformer.ts
@@ -264,7 +264,7 @@ export class ChunkTransformer extends Transform {
       typeof lastRowKey !== 'undefined' &&
       TableUtils.lessThanOrEqualTo(newRowKey as string, lastRowKey as string)
     ) {
-      errorMessage = 'key must be strictly increasing';
+      errorMessage = 'A row key must be strictly increasing';
     } else if (!chunk.familyName) {
       errorMessage = 'A family must be set';
     } else if (chunk.qualifier === null || chunk.qualifier === undefined) {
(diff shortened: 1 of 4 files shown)


src/chunktransformer.ts
```diff
@@ -259,6 +260,11 @@ export class ChunkTransformer extends Transform {
       errorMessage = 'A new row cannot be reset';
     } else if (lastRowKey === newRowKey) {
       errorMessage = 'A commit happened but the same key followed';
+    } else if (
+      typeof lastRowKey !== 'undefined' &&
+      TableUtils.lessThanOrEqualTo(newRowKey as string, lastRowKey as string)
+    ) {
+      errorMessage = 'key must be strictly increasing';
```
ruyadorno, line 267: What about a more friendly message, e.g: `A row key must be strictly increasing`

ruyadorno (approved):

PR #1234, opened 2023-02-07
chore(testproxy): Mutate rows bug fix

danieljbruce:
This PR makes a change to the test proxy that allows the test proxy to use the veneer layer instead to make its call to mutateRows. Using the veneer layer to make the call allows additional parsing to be done on the request and the results which ensures the right results are returned to the test runner enabling the test runner to pass its test.

The test in question is TestMutateRows_NoRetry_NonTransientErrors.


testproxy/services/bulk-mutate-rows.js
```diff
@@ -14,38 +14,44 @@
...
+      };
+    } catch (error) {
+      if (error.name === 'PartialFailureError') {
+        return {
+          status: {code: grpc.status.OK, message: error.message},
+          entry: Array.from(error.errors.entries()).map(([index, entry]) => ({
+            index: index + 1,
+            status: entry,
+          })),
+        };
```
ruyadorno, line 47: why is the proxy turning partial failures into an "OK" response here?

danieljbruce: It meets the needs of the test case and is required for the test case to pass, but this is a good start to a discussion about what the test proxy *should* do.

danieljbruce: Maybe https://github.com/googleapis/nodejs-bigtable/blob/000a353f05fa65bc0a7dc551c777ac55c6c08875/src/table.ts#L1540 should change, but then we have a breaking change?

liujiongxin: We shouldn't change the node js client. 

You can return the status as is, just ensure the entries are filled correctly. I will update the test case as well as the guide of test proxy writeup.

ruyadorno (approved):
LGTM 👍

PR #1249, opened 2023-03-03
chore(testproxy): fix close generated (v2) clients

ruyadorno:
Tests for closing the client are failing since a new client object is created on every operation that uses the generated (v2) layer. This changeset fixes it by moving the v2 client creation to `testproxy/services/create-client.js` alongside the veneer client creation and properly closing that v2 instance on `testproxy/services/close-client.js`.

Every command using the v2 layer is updated to read the client from the client map instead of creating a new instance.

Three previosly failing tests are now passing with this fix:
- TestCheckAndMutateRow_Generic_CloseClient
- TestMutateRow_Generic_CloseClient
- TestReadModifyWriteRow_Generic_CloseClient

### DIFF
--- a/testproxy/services/close-client.js
+++ b/testproxy/services/close-client.js
@@ -15,6 +15,8 @@
 
 const normalizeCallback = require('./utils/normalize-callback.js');
 
+const v2 = Symbol.for('v2');
+
 const closeClient = ({clientMap}) =>
   normalizeCallback(async rawRequest => {
     const request = rawRequest.request;
@@ -22,7 +24,8 @@ const closeClient = ({clientMap}) =>
     const bigtable = clientMap.get(clientId);
 
     if (bigtable) {
-      bigtable.close();
+      await bigtable[v2].close();
+      await bigtable.close();
       return {};
     }
   });
--- a/testproxy/services/remove-client.js
+++ b/testproxy/services/remove-client.js
@@ -15,6 +15,8 @@
 
 const normalizeCallback = require('./utils/normalize-callback.js');
 
+const v2 = Symbol.for('v2');
+
 const removeClient = ({clientMap}) =>
   normalizeCallback(async rawRequest => {
     const request = rawRequest.request;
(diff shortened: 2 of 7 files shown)


testproxy/services/close-client.js
```diff
@@ -15,14 +15,17 @@
...
+
 const closeClient = ({clientMap}) =>
   normalizeCallback(async rawRequest => {
     const request = rawRequest.request;
     const {clientId} = request;
     const bigtable = clientMap.get(clientId);
 
     if (bigtable) {
-      bigtable.close();
+      await bigtable[v2].close();
```
danieljbruce, line 27: I actually think this line might not be necessary.

testproxy/services/remove-client.js
```diff
@@ -15,14 +15,17 @@
...
+
 const removeClient = ({clientMap}) =>
   normalizeCallback(async rawRequest => {
     const request = rawRequest.request;
     const {clientId} = request;
     const bigtable = clientMap.get(clientId);
 
     if (bigtable) {
-      bigtable.close();
+      await bigtable[v2].close();
```
danieljbruce, line 27: Same comment as above

danieljbruce (commented):
A couple of suggestions

danieljbruce (approved):

testproxy/services/close-client.js
ruyadorno: Good point but I just confirmed it's necessary. I tried running the tests without closing the v2 instance (removing this line) and tests fail again without it.

My understanding is that veneer and v2 are completely independent instances so it makes sense that it needs to be closed separately.

danieljbruce: Got it. If that is the case then it looks like the code contains a client bug because closing the veneer is supposed to close the clients on the v2 layer. Oh, but the client never gets initialized because we never make calls with the veneer layer and clients are initialized lazily. I suspect though that when we use the veneer layer for the `mutateRow` calls when `mutateRow` is exposed then we will not need this line anymore.

Thanks for checking.

PR #1251, opened 2023-03-13
fix: Always set the retry attempt to 0 for now

danieljbruce:
If the retry attempt is greater than 1 then a library called retryRequest creates a large delay. If we force the retry count to be 0 then we skip this delay and the requests run in the required timeline.

### DIFF
--- a/src/table.ts
+++ b/src/table.ts
@@ -776,7 +776,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       } as google.bigtable.v2.IReadRowsRequest;
 
       const retryOpts = {
-        currentRetryAttempt: numConsecutiveErrors,
+        currentRetryAttempt: 0, // was numConsecutiveErrors
         // Handling retries in this client. Specify the retry options to
         // make sure nothing is retried in retry-request.
         noResponseRetries: 0,


src/table.ts
```diff
@@ -776,7 +776,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       } as google.bigtable.v2.IReadRowsRequest;
 
       const retryOpts = {
-        currentRetryAttempt: numConsecutiveErrors,
+        currentRetryAttempt: 0, // was numConsecutiveErrors
```
liujiongxin, line 779: Do you know why "numConsecutiveErrors" doesn't work here even though its value is 0 according to Line 730?

danieljbruce: `numConsecutiveErrors` works, but if it is greater than 0 then that will make the request take longer. Note that the number of consecutive errors is incremented at https://github.com/googleapis/nodejs-bigtable/blob/44fab5cdbb6289f08be17c55f5d7402f89b5d46f/src/table.ts#L924 on each retry.

liujiongxin: I found this

`calculatedNextRetryDelay =
    config.initialRetryDelayMillis *
      Math.pow(config.retryDelayMultiplier, numConsecutiveErrors)`

It implies that when it's executed for the first time (i.e. the original request has failure), `numConsecutiveErrors` should be 0 such that the retry delay is ` config.initialRetryDelayMillis`

liujiongxin (approved):

PR #1272, opened 2023-05-05
fix: Require stream destruction to be done with error

danieljbruce:
This fix will ensure that users receive all the data they request even if the chunk transformer stream is unintentionally destroyed.

### DIFF
--- a/src/chunktransformer.ts
+++ b/src/chunktransformer.ts
@@ -169,11 +169,11 @@ export class ChunkTransformer extends Transform {
    */
   destroy(err?: Error): this {
     if (this._destroyed) return this;
-    this._destroyed = true;
     if (err) {
+      this._destroyed = true;
       this.emit('error', err);
+      this.emit('close');
     }
-    this.emit('close');
     return this;
   }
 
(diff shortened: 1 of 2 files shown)


src/chunktransformer.ts
```diff
@@ -167,13 +167,15 @@ export class ChunkTransformer extends Transform {
    * @public
    * @param {error} err error if any
    */
+  // When this function is called in nodejs-bigtable, ensure an error is passed
+  // in to destroy the stream properly.
   destroy(err?: Error): this {
     if (this._destroyed) return this;
-    this._destroyed = true;
     if (err) {
+      this._destroyed = true;
```
igorbernstein2, line 175: I dont think this is correct, as the stream will never be destroyed in a non-error case

igorbernstein2 (changes requested):

src/chunktransformer.ts
danieljbruce: Note that this `_destroyed` variable is just a local variable that informs createReadStream to stop passing data along when set to true. What if we move `this.emit('close');` back to where it was so that the stream gets destroyed, but data still continues to pass through in `createReadStream` until it's done?

danieljbruce: If we move `this.emit('close');` outside the if block then that will destroy the stream. Does that address this concern?

PR #1273, opened 2023-05-08
Require stream destroy without fix

danieljbruce:
This PR illustrates a test that passes if the chunks get sent back whole so that we can more easily explore solutions that will work and solutions that don't.

### DIFF
--- a/system-test/bigtable.ts
+++ b/system-test/bigtable.ts
@@ -539,21 +539,23 @@ describe('Bigtable', () => {
           rowStatus: 'commitRow',
         };
       }
-      const chunks = [];
-      for (let i = 0; i < 209; i++) {
-        chunks.push(getChunk(i));
-      }
-      const data = {
-        chunks,
-        lastScannedRowKey: Buffer.from('a'),
-      };
-      stream.push(data);
-      setTimeout(() => {
-        try {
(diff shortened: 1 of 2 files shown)


system-test/bigtable.ts
```diff
@@ -488,6 +491,83 @@ describe('Bigtable', () => {
...
+        for (let i = 0; i < numChunks; i++) {
+          chunks.push(getChunk(i));
+        }
+        const data = {
+          chunks,
+          lastScannedRowKey: Buffer.from('a'),
+        };
+        stream.push(data);
+      }
+      setTimeout(() => {
```
alexander-fenster, line 554: Change this to `process.nextTick`.

alexander-fenster: Actually, maybe even emit each chunk at `nextTick` and then `end` at `nextTick` as well.

danieljbruce: I gave your former suggestion a try. For your latter suggestion do you mean something like this?

```
let chunksPushed = 0;
process.nextTick(() => {
    if (chunksPushed > 83) {
        pushChunks(chunkSize);
        chunksPushed++;
    } else {
        stream.emit('end');
    }
});
```

PR #1277, opened 2023-05-12
fix: Parsing for qualifiers with colon characters

KDB223:
`parseColumnName()` uses `.split(':')` to find the qualifier, but this parses qualifiers with `:` chars in it incorrectly, since they also are picked up by the split. This causes it to ignore everything after `:` in the qualifier.

e.g. consider a column name `"foo:bar:baz"`
`parseColumnName()` parses this incorrectly as 
```
{
  family: 'foo', 
  qualifier: 'bar'
}
```

Using `slice()` instead with the index of the first `:` character fixes this:
```
{
  family: 'foo', 
  qualifier: 'bar:baz'
}
```

### DIFF
--- a/src/mutation.ts
+++ b/src/mutation.ts
@@ -386,7 +386,7 @@ export class Mutation {
       // columnName does not contain ':'
       return {
         family: columnName,
-        qualifier: undefined,
+        qualifier: null,
       };
     }
       
--- a/src/mutation.ts
+++ b/src/mutation.ts
@@ -381,15 +381,15 @@ export class Mutation {
    */
   static parseColumnName(columnName: string): ParsedColumn {
     const colonIdx = columnName.indexOf(':');
-      
-    if (colonIdx == -1) {
+
+    if (colonIdx === -1) {
       // columnName does not contain ':'
       return {
         family: columnName,
         qualifier: null,
       };
     }
-      
+ 
     return {
       family: columnName.slice(0, colonIdx),
       qualifier: columnName.slice(colonIdx + 1),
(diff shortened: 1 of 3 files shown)


danieljbruce:
Please fix the tests `should array-ify the input` and `should parse a family name` and modify the last commit so that the conventional commits github action passes.

KDB223:
All done. Apologies for the noise.

Changed the qualifier assertion to check for `null` instead of `undefined` since we can't return `undefined` from the method.

src/mutation.ts
```diff
@@ -380,11 +380,19 @@ export class Mutation {
...
    */
   static parseColumnName(columnName: string): ParsedColumn {
-    const parts = columnName.split(':');
+    const colonIdx = columnName.indexOf(':');
+
+    if (colonIdx === -1) {
+      // columnName does not contain ':'
+      return {
+        family: columnName,
+        qualifier: null,
```
danieljbruce, line 389: I recommend setting `qualifier` to `undefined` so that this code works the same way as before when the string doesn't contain a colon.

KDB223: I tried that too, however the `ParsedColumn` interface does not declare `undefined` as an acceptable type for `qualifier` at https://github.com/googleapis/nodejs-bigtable/blob/main/src/mutation.ts#L35.

Adding `undefined` here would solve this:

```typescript
export interface ParsedColumn {
  family: string | null;
  qualifier: string | null | undefined;
}
```

danieljbruce (approved):

PR #1284, opened 2023-05-19
fix: properly handle asynchronous read from stream

### LINKED ISSUE #607: createReadStream didn't returns all data when stream is pause and resumed
Hello,

I have an issue regarding createReadStream method in nodejs bigtable client, in may situation I need to have a way to get only one item from the stream and do some process on it then get the next and so on, and I  don't want to store all the data in memory, and bigtable client in nodejs didn't provide any thing to do it, I use the createReadStream and pause the stream once I get an Item/Row from the table.

This worked fine for small data, but when I tried to do it on large data about 500k rows I didn't retrieve all the rows that I have in the table for the query I made (the 'end' event is emitted before I get all the result from the table).

My code use class similar to this code, the class is supposed to return promise resolves to the next item from the stream or undefined if the stream is finished

`export default class Iterable {
    stream: any;
    buffer: any[] = [];
    public cnt: number = 0;
    id = 0;

    constructor(id: number, stream: any) {
        this.id = id;
        this.stream = stream;

        this.stream.on('end', () => {
            console.log('end', this.cnt, this.buffer.length);
            this.buffer.push(undefined);
        });
        this.stream.once('data', (d: any) => {
            this.buffer.push(d);
            this.stream.pause();
        });
    }

    async next(): Promise<any> {
        this.stream.resume();
        return new Promise(resolve => {
            if(this.cnt < this.buffer.length) {
                resolve(this.buffer[this.cnt++]);
            } else {
                let id = setInterval(() => {
                    if (this.cnt < this.buffer.length) {
                        clearInterval(id);
                        resolve(this.buffer[this.cnt++]);
                    }
                }, 1);
            }
        });
    }
}`

I also try to use once to only listen on data event on time each time I call next() but I still also didn't get all the data that I should get.

I really hope you solve this issue or create another way to get only one item from the bigtable at each step.

#### Environment details

  - OS: Ubuntu 19.10
  - Node.js version: 10.15.2
  - npm version: 6.12.0
  - `@google-cloud/bigtable` version: 2.3.1

I also use bigtable emulator

#### Steps to reproduce

  1. Have a big table with large number of rows
  2. Try to use the above class to get all the data

Thanks!

alexander-fenster:
Fixes #607.

~Note: I'm currently comparing this pull request to #1282, I will change the base to `main` after #1282 is merged. This way it's easier to review.~

Several changes here:

1. Remove the logic in `table.ts` that would skip the unprocessed records if a `ChunkTransformer` has `_destroyed` flag set. There are two separate cases when the `ChunkTransformer` gets destroyed:

    1. if it's a network problem or anything else that forced gRPC to emit an `error` event. In such case, it always makes sense to process all the rows we have received before giving up.

    2. if it's a user cancellation (by calling `stream.end()` from the user code). In this case we need to stop emitting new rows.

~So let's introduce a new flag, `_canceled`, and ignore the incoming rows only if the stream is canceled from `.end()`.~ Let's have the new flag, `userCanceled`, and use it in `userStream` `transform` method to stop emitting new rows when the user don't need them, but we'll still keep emitting all rows if there was no user cancellation. This change fixes #607.

2. ~When a network error happens, we need to update the `lastRowKey` only if the row processing has completed (the commit bit was set in the last chunk). Until then, don't update `lastRowKey`. The current code makes recovery skip one record - the one that was incomplete.~ This was actually a test problem, there is no problem here.

3. Minor change to make modern TypeScript compiler happy when comparing `chunk.timestampMicros` to zero. `chunk.timestampMicros` can theoretically be an instance of `Long`, so let's handle this accordingly. It was showing as an error in my editor and it makes sense so I'm fixing it in the safest way possible. In the default case, the `||` expression will short circuit in the `typeof chunk.timestampMicros === 'number'` branch.


alexander-fenster:
Note: the system and samples tests won't run here until I rebase this PR to be against the `main` branch, but I ran them locally and they pass.

Also note: I cherry-picked a commit from #1285 to avoid an unrelated Windows test failure.

src/chunktransformer.ts
```diff
@@ -91,10 +92,15 @@ export class ChunkTransformer extends Transform {
     super(options);
     this.options = options;
     this._destroyed = false;
+    this._userCanceled = false;
     this.lastRowKey = undefined;
     this.reset();
   }
 
+  get canceled() {
```
igorbernstein2, line 100: It seems like we expose ChunkTransformer as part of our public api. Which I think is incorrect to begin with. To minimize further leakage of implementation details, can we mark this method as internal?

src/chunktransformer.ts
```diff
@@ -226,7 +221,14 @@ export class ChunkTransformer extends Transform {
...
       (chunk.value && chunk.value.length !== 0) ||
-      chunk.timestampMicros! > 0;
+      // if it's a number
+      (typeof chunk.timestampMicros === 'number' &&
+        chunk.timestampMicros! > 0) ||
+      // if it's an instance of Long
+      (typeof chunk.timestampMicros === 'object' &&
+        'compare' in chunk.timestampMicros &&
+        typeof chunk.timestampMicros.compare === 'function' &&
+        chunk.timestampMicros.compare(0) === 1);
```
igorbernstein2, line 231: This is a bit strange, chunk is a protobuf, so shouldnt it have a stable type?

src/table.ts
```diff
@@ -882,7 +885,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       const toRowStream = new Transform({
         transform: (rowData, _, next) => {
           if (
-            chunkTransformer._destroyed ||
+            chunkTransformer.canceled ||
```
igorbernstein2, line 888: It seems like there are only 2 things that care about the cancelled flag:
- the closure for userstream.end()
- the closure for toRowStream

ChunkTransformer seems like an innocent bystander here. Why not move the userCancelled flag as a local var in createReadStream(), whose scope is shared by the 2 closure that care about the flag?

src/chunktransformer.ts
```diff
@@ -147,17 +153,6 @@ export class ChunkTransformer extends Transform {
...
-      }
-    }
-    if (data.lastScannedRowKey && data.lastScannedRowKey.length > 0) {
-      this.lastRowKey = Mutation.convertFromBytes(
-        data.lastScannedRowKey as Bytes,
-        {
-          userOptions: this.options,
-        }
-      );
     }
```
igorbernstein2, line 156: I think this breaks request resumption logic in makeNewRequest

`const lastRowKey = chunkTransformer ? chunkTransformer.lastRowKey : '';`

igorbernstein2 (changes requested):

igorbernstein2:
Also I think we still have a problem of end() on the userStream being delayed. I think we can deal with it in a separate PR but I think that in the case where the consumer is slower than the grpc producer (ie. the user transformer calls the callback in a setTimeout()). Then rows can get buffered in the userStream. When the end user calls cancels the stream early via an end(), the rows buffered in the userStream will still be emitted.

I'm not sure how to deal with this. We would need the userStream to signal the stream end on read() when cancelled()

src/table.ts
alexander-fenster: Moving the flag closer to the consumer makes sense to me, I'll try to do it. I only put it here to replace the `_destroyed` logic, but you are right, it will make more sense downstream.

src/chunktransformer.ts
alexander-fenster: Fixed this

src/table.ts
alexander-fenster: Done

src/chunktransformer.ts
igorbernstein2: please leave a comment in the code explaining this

src/table.ts
```diff
@@ -745,22 +745,35 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+        if (userCanceled) {
+          callback();
+          return;
+        }
+        callback(null, row);
+      },
+    });
+    const originalEnd = userStream.end.bind(userStream);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    userStream.end = (chunk?: any, encoding?: any, cb?: () => void) => {
```
igorbernstein2, line 764: please add a comment explaining that this is needed to fulfill a pre-existing contract with the caller, where the caller can prematurely cancel the stream by calling end. And we need to be able to disambiguate between a user cancellation and the a normal end. So we capture the originalEnd to be used for normal termination and the overwritten end is meant for end users

src/table.ts
```diff
@@ -745,22 +745,35 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
       rowStream?.unpipe(userStream);
+      rowStream.removeListener('end', originalEnd);
```
igorbernstein2, line 766: can you add 2 helper functions rowStreamPipe & rowStreamUnpipe  and add comments that we need to do special handling of the end event to re-direct the handler to the originalEnd

src/table.ts
```diff
@@ -914,6 +927,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       rowStream
         .on('error', (error: ServiceError) => {
           rowStream.unpipe(userStream);
+          rowStream.removeListener('end', originalEnd);
```
igorbernstein2, line 930: please delegate to a helper

igorbernstein2 (changes requested):

src/table.ts
```diff
@@ -745,22 +745,51 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+      rowStream.on('end', originalEnd);
+    };
+    const rowStreamUnpipe = (rowStream: Duplex, userStream: PassThrough) => {
       rowStream?.unpipe(userStream);
+      rowStream?.removeListener('end', originalEnd);
+    };
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    userStream.end = (chunk?: any, encoding?: any, cb?: () => void) => {
+      rowStreamUnpipe(rowStream, userStream);
```
danieljbruce, line 782: Does this prevent the data from being passed on from rowStream to userStream? It looks like we are using both `userCanceled` and unpipe to stop passing data along. Should we also unpipe the data being passed from the chunk transformer to `toRowStream`?

danieljbruce (commented):
Looks good. A couple of questions/comments about specific details.

src/table.ts
alexander-fenster: We don't really care about the data passed from `chunkTransformer` to `toRowStream` because this is all inside `rowStream`. We only need to stop that data from flowing into the user stream. Since it's an end of the user stream, it's not really important if more or less data has flown inside the `rowStream` pipeline.

src/table.ts
```diff
@@ -745,22 +745,51 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+
+    let userCanceled = false;
+    const userStream = new PassThrough({
+      objectMode: true,
+      transform(row, _encoding, callback) {
+        if (userCanceled) {
+          callback();
+          return;
+        }
+        callback(null, row);
```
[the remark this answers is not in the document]
alexander-fenster: Yes. The reason we also have the condition in `toRowStream.transform` is just because it does not make much sense to pipe more data through `toRowStream` after the user explicitly asked us to stop.

igorbernstein2 (approved):

danieljbruce (approved):

PR #1313, opened 2023-08-09
fix: remove the watermarks

### LINKED ISSUE #1286: readStream.end() called asynchronously returns 16 extra rows before stopping
In the current implementation of `createReadStream` in `table.ts`, the end user is supposed to stop the streaming synchronously by calling `readStream.end()`. If they instead call it asynchronously, `setTimeout(() => { readStream.end(); }, 0);`, a few records (16 to be precise, that's the default value of `highWaterMark` for object streams) will still be received.

See the skipped test `should be able to stop reading from the read stream when reading asynchronously`:

https://github.com/googleapis/nodejs-bigtable/blob/5c7c5aecae3f4c2355e3871ba4dc9890d8a4d3ca/test/readrows.ts#L226

danieljbruce:
Removing the watermarks allows data not to be stuck at the last transform so that data doesn’t get stuck due to the watermarks.

Fixes #1286


test/readrows.ts
```diff
@@ -252,6 +252,7 @@ describe('Bigtable/ReadRows', () => {
     // Transform stream
     const transform = new Transform({
       objectMode: true,
+      writableHighWaterMark: 0,
```
ruyadorno, line 255: why is `writableHighWaterMark` only set in the test and not in the implementation (in `src/table.ts`) similar to `readableHighWaterMark` ?

igorbernstein2 (approved):

ruyadorno (approved):

PR #1349, opened 2023-10-03
test: Add conformance tests

danieljbruce:
Add conformance tests that use the test proxy to ensure the client is running correctly. This PR should be very similar to [https://github.com/googleapis/java-bigtable/pull/1699](https://www.google.com/url?q=https://github.com/googleapis/java-bigtable/pull/1699&sa=D&source=buganizer&usg=AOvVaw3uY2epO8IEddjTnUwg00ez) which adds conformance tests to java-bigtable. A list of known test failures is also provided so that we only test against the test proxy tests that pass and our continuous integration pipeline passes. The idea is that we want to reduce this list over time as the test proxy improves.


bhshkh (approved):

.github/workflows/conformance.yaml
```diff
@@ -0,0 +1,41 @@
...
+  push:
+    branches:
+    - main
+  pull_request:
+name: conformance
+jobs:
+  conformance:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
```
daniel-sanche, line 26: Is this needed? Looks like a duplicate of the step below

.github/workflows/conformance.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - uses: actions/checkout@v3
+      with:
+        repository: googleapis/cloud-bigtable-clients-test
+        ref: main
+        path: cloud-bigtable-clients-test
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 14
```
daniel-sanche, line 34: Do you think it would be useful to test with multiple node versions?

testproxy/known_failures.txt
```diff
@@ -0,0 +1 @@
+TestCheckAndMutateRow_NoRetry_TransientError\|TestCheckAndMutateRow_Generic_Headers\|TestCheckAndMutateRow_Generic_DeadlineExceeded\|TestMutateRow_Generic_Headers\|TestMutateRow_Generic_DeadlineExceeded\|TestMutateRows_Generic_CloseClient\|TestReadModifyWriteRow_Generic_Headers\|TestReadModifyWriteRow_NoRetry_TransientError\|TestReadModifyWriteRow_Generic_DeadlineExceeded\|TestReadRow_Generic_DeadlineExceeded\|TestReadRows_Retry_PausedScan\|TestReadRows_Retry_StreamReset\|TestMutateRows_Generic_DeadlineExceeded\|TestSampleRowKeys_Generic_Headers\|TestReadRows_Generic_DeadlineExceeded\|TestReadRows_Generic_CloseClient\|TestSampleRowKeys_Generic_DeadlineExceeded
```
daniel-sanche, line 1: I'm a little unsure if we should be skipping all these tests, maybe we should just let the test stay red until these are fixed?

daniel-sanche (approved):

.github/workflows/conformance.yaml
danieljbruce: I bet it's not needed, but I'm not completely sure. The PR in Java has this twice so I figured there must be a reason it is there https://github.com/googleapis/java-bigtable/pull/1699/files. I'll investigate further.

.github/workflows/conformance.yaml
danieljbruce: Yes

testproxy/known_failures.txt
danieljbruce: Let's look into this further.

danieljbruce (commented):
Answer to review comments.

.github/workflows/conformance.yaml
danieljbruce: I've been informed by @liujiongxin this check is probably for the Java client repo itself.

testproxy/known_failures.txt
triplequark: We should have two test paths: default and all tests enabled.

.github/workflows/conformance.yaml
daniel-sanche: Right, that makes sense. I forgot there were two repos involved!

.kokoro/conformance.sh
```diff
@@ -0,0 +1,44 @@
...
+
+set -eo pipefail
+
+export NPM_CONFIG_PREFIX=${HOME}/.npm-global
+
+## cd to the parent directory, i.e. the root of the git repo
+cd $(dirname $0)/..
+
+# Build and start the proxy in a separate process
+pushd .
```
daniel-sanche, line 25: The pushd and popd shouldn't be necessary if you stay at the root

.kokoro/conformance.sh
```diff
@@ -0,0 +1,44 @@
...
+npm install
+nohup npm run testproxy &
+proxyPID=$!
+popd
+
+# Run the conformance test
+pushd .
+pwd
+ls
+cd cloud-bigtable-clients-test/tests
```
daniel-sanche, line 35: shouldn't this be in pushd?

.kokoro/conformance.sh
```diff
@@ -0,0 +1,44 @@
...
+pushd .
+npm install
+nohup npm run testproxy &
+proxyPID=$!
+popd
+
+# Run the conformance test
+pushd .
+pwd
+ls
```
daniel-sanche, line 34: was pwd and ls added for debugging? Can probably remove them

danieljbruce: Yes. These debugging commands have been removed now.

.kokoro/conformance.sh
danieljbruce: This code is copied straight from the https://github.com/googleapis/java-bigtable/pull/1699/files#diff-08b007a16e10891ada78edba8173c7323e74500fa6e1600c5e1dc7a721df5388R51-R56. Do you mean `pushd ./cloud-bigtable-clients-test/tests`? here and then eliminating the `pushd .` line?

That seems like a good idea that would condense the code, sure.

.kokoro/conformance.sh
danieljbruce: What exactly would you suggest here if we eliminate pushd and popd? I tried subshells locally https://unix.stackexchange.com/questions/13802/execute-a-specific-command-in-a-given-directory-without-cding-to-it, but the command hangs.

Looking at this further, I think in the Java example we wanted to go back to the root because we wanted to run https://github.com/googleapis/java-bigtable/pull/1699/files#diff-08b007a16e10891ada78edba8173c7323e74500fa6e1600c5e1dc7a721df5388R62. Since we don't need to run anything else in this script after the golang tests however, I think we can eliminate `pushd .` and `popd` and not worry about staying at the root.

.kokoro/conformance.sh
danieljbruce: Now that  `pushd .` has been removed, no further action is required here.

.kokoro/conformance.sh
```diff
@@ -0,0 +1,40 @@
...
+proxyPID=$!
+popd
+
+# Run the conformance test
+cd cloud-bigtable-clients-test/tests
+eval "go test -v -proxy_addr=:9999 -skip `cat ../../testproxy/known_failures.txt`"
+RETURN_CODE=$?
+
+# Stop the proxy
+kill $proxyPID
```
bhshkh, line 37: This should be executed irrespective of when / how script exits. Exit traps can be used here. E.g. https://github.com/googleapis/google-cloud-go/blob/5de394be14cf852b99dde65fbeba3d02714f6845/storage/emulator_test.sh#L64-L70

danieljbruce: Sure. Makes sense. So would you suggest adding `trap kill $proxyPID` somewhere at the top of the file and removing this line? In that case would any exit signal still get propogated up to the user?

bhshkh: Add a cleanup function.
Trap on EXIT signal which would catch any exit signal.

Similar to https://github.com/googleapis/google-cloud-go/blob/5de394be14cf852b99dde65fbeba3d02714f6845/storage/emulator_test.sh#L70

danieljbruce: Done

.kokoro/conformance.sh
daniel-sanche: pushd and popd are only needed if you're changing directories. In the case of this specific block , it seems like a no-op, so yeah it makes sense to remove it

Alternatively, you could replace all the `cd` statements with `pushd` statements, which can be a lot cleaner if you have to move around the filesystem. But since this script is pretty simple, you can probably just stick with the `cds` you have

.kokoro/conformance.sh
daniel-sanche: Was the trap change pushed? I don't see the changes here

danieljbruce: In my git client at about 1:30 I made the changes and pushed them. Not sure why they don't show up.

image

danieljbruce: It is called "Add a cleanup operation"

triplequark (approved):

PR #1401, opened 2024-04-12
test: rewrite createreadstream tests

danieljbruce:
When retries are moved to google-gax from the handwritten layer of `nodejs-bigtable` then the current tests mentioned in the PR will not work. They need to be rewritten to still test retry behavior, but account for the fact that the retries will be done in a different layer.

Here is what is done to modify [`Tests in [read-rows.ts]`](https://github.com/googleapis/nodejs-bigtable/blob/17838eda19b001e322765c33a83a756eeeb68963/system-test/read-rows.ts#L126) just to clarify further:
1. This [sinon stub mock](https://github.com/googleapis/nodejs-bigtable/blob/17838eda19b001e322765c33a83a756eeeb68963/system-test/read-rows.ts#L153) currently mocks out the request function and sends responses back to createreadstream where retries are currently done. Since retries will not be done in `createreadstream` anymore and this test intends to test retries then responses must be mocked in a layer below where retries are going to be done in google-gax in a later change. The current sinon stub mock should be removed and instead the [mock grpc server](https://github.com/googleapis/nodejs-bigtable/blob/main/src/util/mock-servers/mock-server.ts) should send [responses](https://github.com/googleapis/nodejs-bigtable/blob/17838eda19b001e322765c33a83a756eeeb68963/system-test/data/read-rows-retry-test.json#L33-L36) back in order to test retry behavior.
2. As is done with the current mock, the new mock server mock should check each request to ensure it matches [request_options](https://github.com/googleapis/nodejs-bigtable/blob/17838eda19b001e322765c33a83a756eeeb68963/system-test/data/read-rows-retry-test.json#L27-L32)

### DIFF
--- a/system-test/read-rows.ts
+++ b/system-test/read-rows.ts
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {Bigtable, Table} from '../src';
+import {Bigtable, protos, Table} from '../src';
 import {Mutation} from '../src/mutation.js';
 const {tests} = require('../../system-test/data/read-rows-retry-test.json') as {
   tests: Test[];
@@ -28,6 +28,7 @@ import {PassThrough} from 'stream';
 import {MockServer} from '../src/util/mock-servers/mock-server';
 import {MockService} from '../src/util/mock-servers/mock-service';
 import {BigtableClientMockService} from '../src/util/mock-servers/service-implementations/bigtable-client-mock-service';
+import {ServerWritableStream} from '@grpc/grpc-js';
 
 const {grpc} = new GrpcClient();
 
@@ -76,6 +77,33 @@ function rowResponse(rowKey: {}) {
     commitRow: true,
     value: 'value',
   };
+}
+
+function getRequestOptions(request: any): google.bigtable.v2.IRowSet {
+  const requestOptions = {} as google.bigtable.v2.IRowSet;
+  if (request.rows && request.rows.rowRanges) {
+    requestOptions.rowRanges = request.rows.rowRanges.map(
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (range: any) => {
+        const convertedRowRange = {} as {[index: string]: string};
+        Object.keys(range).forEach(
+          key => (convertedRowRange[key] = range[key].asciiSlice())
+        );
+        return convertedRowRange;
+      }
+    );
+  }
+  if (request.rows && request.rows.rowKeys) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    requestOptions.rowKeys = request.rows.rowKeys.map((rowKeys: any) =>
+      rowKeys.asciiSlice()
+    );
+  }
+  if (request.rowsLimit) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (requestOptions as any).rowsLimit = request.rowsLimit;
+  }
+  return requestOptions;
 }
 
 describe('Bigtable/Table', () => {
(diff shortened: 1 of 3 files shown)


system-test/read-rows.ts
```diff
@@ -12,69 +12,76 @@
...
-    timestampMicros: 0,
-    labels: [],
+    qualifier: {value: Buffer.from('qualifier').toString('base64')},
     commitRow: true,
-    value: 'value',
+    value: Buffer.from(rowKey).toString('base64'),
   };
 }
 
+function getRequestOptions(request: any): google.bigtable.v2.IRowSet {
```
danieljbruce, line 41: From the old test, this contains all the code which was used for building a request.

system-test/read-rows.ts
```diff
@@ -123,97 +130,89 @@ describe('Bigtable/Table', () => {
...
+          assert.deepStrictEqual(requestedOptions, test.request_options);
+          done();
         }
-        assert.deepStrictEqual(rowKeysRead, test.row_keys_read);
-        assert.strictEqual(
-          responses.length,
-          0,
-          'not all the responses were used'
-        );
-        assert.deepStrictEqual(requestedOptions, test.request_options);
```
danieljbruce, line 216: These asserts are moved into `checkResults` which evaluates response/request data when the call to createreadstream results in an error or ended stream.

danieljbruce (commented):
Added a few explanations throughout the PR

system-test/read-rows.ts
```diff
@@ -12,69 +12,76 @@
...
+    value: Buffer.from(rowKey).toString('base64'),
   };
 }
 
+function getRequestOptions(request: any): google.bigtable.v2.IRowSet {
+  const requestOptions = {} as google.bigtable.v2.IRowSet;
+  if (request.rows && request.rows.rowRanges) {
+    requestOptions.rowRanges = request.rows.rowRanges.map(
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (range: any) => {
```
leahecole, line 46: Is there another type option than `any` that can be used here?

system-test/read-rows.ts
```diff
@@ -12,69 +12,76 @@
...
+        }
+        Object.keys(range).forEach(
+          key => (convertedRowRange[key] = range[key].asciiSlice())
+        );
+        return convertedRowRange;
+      }
+    );
+  }
+  if (request.rows && request.rows.rowKeys) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
leahecole, line 67: Is there another type we can use besides any?

system-test/read-rows.ts
```diff
@@ -12,69 +12,76 @@
...
+  }
+  // The grpc protocol sets rowsLimit to '0' if rowsLimit is not provided in the
+  // grpc request.
+  //
+  // Do not append rowsLimit to collection of request options if received grpc
+  // rows limit is '0' so that test data in read-rows-retry-test.json remains
+  // shorter.
+  if (request.rowsLimit && request.rowsLimit !== '0') {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (requestOptions as any).rowsLimit = parseInt(request.rowsLimit);
```
leahecole, line 80: once again asking about the `any` 🙂

system-test/read-rows.ts
```diff
@@ -123,97 +130,93 @@ describe('Bigtable/Table', () => {
...
+            rowKeysRead.push([]);
+            requestedOptions.push(getRequestOptions(stream.request));
+            if (response.row_keys) {
+              stream.write({
+                chunks: response.row_keys.map(rowResponseFromServer),
+              });
+            }
+            if (response.end_with_error) {
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              const error: any = new Error();
```
leahecole, line 201: Is this an error coming from the server? If so is it a GoogleError not an any?

system-test/testTypes.ts
```diff
@@ -46,3 +47,23 @@ export interface Test {
...
+}
+export interface ReadRowsTest {
+  createReadStream_options?: GetRowsOptions;
+  max_retries: number;
+  responses: CreateReadStreamResponse[];
+  request_options: CreateReadStreamRequest[];
+  error: number;
+  row_keys_read: string[][];
+  name: string;
+}
```
leahecole, line 69: In here there is an interesting mix of snake_case and camelCase - is there a reason why both have to be used? Or can they all be migrated to a more typical JS-y camelCase?

leahecole (changes requested):

leahecole:
Is the failure of the conformance tests expected?

danieljbruce:
> Is the failure of the conformance tests expected?

Yup. Right now the conformance tests are supposed to be running, but a whole bunch of them fail.

system-test/read-rows.ts
danieljbruce: Yup. I made this more specific and solved the resulting compiler errors too.

system-test/read-rows.ts
danieljbruce: I got rid of this comment and made it compile. This was copy/paste from the original test along with some other any's. Now the types are more specific, but a lot of interfaces and guards needed to be added to solve compiler errors up and down the stack.

system-test/read-rows.ts
danieljbruce: This one was a good one to fix. Helped catch some incorrect types downstream.

system-test/read-rows.ts
danieljbruce: Yup. Changed to GoogleError.

system-test/testTypes.ts
danieljbruce: They can be migrated, but it means that the json file containing all the test data in https://github.com/googleapis/nodejs-bigtable/blob/3f5db964062502cd70235714eb7b5fb62a211d4f/system-test/data/read-rows-retry-test.json#L112 needs to be updated to match.

PR #1403, opened 2024-04-22
refactor: Move retries finish making createreadstream smaller

danieljbruce:
This PR is written to reduce the amount of code in createreadstream so it is easier to work with when moving the retries. Before merging to main we can always revert these changes if we prefer the code to be inline.

Code from createreadstream is pulled out into 3 functions:
- getRowKeys
- spliceRanges
- #readRowsReqOpts


src/util/mock-servers/mock-server.ts
```diff
@@ -37,7 +37,6 @@ export class MockServer {
       `localhost:${this.port}`,
       grpc.ServerCredentials.createInsecure(),
       () => {
-        server.start();
```
danieljbruce, line 40: As stated before. start is deprecated for grpc servers.

src/table.ts
```diff
@@ -741,10 +741,6 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       ranges.push({});
     }
 
-    if (options.filter) {
-      filter = Filter.parse(options.filter);
-    }
-
```
danieljbruce, line 747: This block along with the `if (filter)` line below is moved to `#readRowsReqOpts`.

src/table.ts
```diff
@@ -800,11 +796,6 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       chunkTransformer = new ChunkTransformer({decode: options.decode} as any);
 
-      const reqOpts = {
-        tableName: this.name,
-        appProfileId: this.bigtable.appProfileId,
-      } as google.bigtable.v2.IReadRowsRequest;
-
```
danieljbruce, line 807: This block is moved to #readRowsReqOpts

src/table.ts
```diff
@@ -870,25 +825,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
-        Filter.createRange(
-          range.start as BoundData,
-          range.end as BoundData,
-          'Key'
-        )
-      );
-
-      if (filter) {
-        reqOpts.filter = filter;
-      }
```
danieljbruce, line 891: This whole block above is moved to #readRowsReqOpts

src/table.ts
```diff
@@ -816,44 +807,8 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
-              range.start = {
-                value: lastRowKey,
-                inclusive: false,
-              };
-            } else {
-              // EndKey is read, remove this range
-              ranges.splice(index, 1);
-            }
-          }
-        }
```
danieljbruce, line 851: This block of code is moved to the new spliceRanges function.

src/table.ts
```diff
@@ -816,44 +807,8 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
-
-        // Remove rowKeys already read.
-        rowKeys = rowKeys.filter(rowKey =>
-          TableUtils.greaterThan(rowKey, lastRowKey as string)
-        );
```
danieljbruce, line 856: This block of code above is moved to the new TableUtils.getRowKeys function.

src/table.ts
```diff
@@ -2049,7 +2016,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
  * that a callback is omitted.
  */
 promisifyAll(Table, {
-  exclude: ['family', 'row'],
+  exclude: ['family', 'row', '#readRowsReqOpts'],
```
danieljbruce, line 2019: This makes it so our new helper function #readRowsReqOpts works as a normal function and doesn't return a promise.

danieljbruce (commented):
Adding some comments to clearly explain where all the code was moved to.

src/table.ts
```diff
@@ -719,11 +719,10 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
    * region_tag:bigtable_api_table_readstream
    */
   createReadStream(opts?: GetRowsOptions) {
-    const options = opts || {};
+    const options: GetRowsOptions = opts || {};
     const maxRetries = is.number(this.maxRetries) ? this.maxRetries! : 10;
     let activeRequestStream: AbortableDuplex | null;
     let rowKeys: string[];
-    let filter: {} | null;
```
danieljbruce, line 726: Not used anymore.

leahecole: This doesn't affect at all what users pass in, right? If this doesn't affect what users pass in, then cool. If it does, we have to proceed with caution

src/table.ts
```diff
@@ -870,25 +824,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
-          range.start as BoundData,
-          range.end as BoundData,
-          'Key'
-        )
-      );
-
-      if (filter) {
-        reqOpts.filter = filter;
-      }
+      const reqOpts: any = this.#readRowsReqOpts(ranges, rowKeys, options);
```
leahecole, line 827: Please address this any!

src/table.ts
```diff
@@ -1640,6 +1576,40 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
     makeNextBatchRequest();
   }
 
+  #readRowsReqOpts(
```
leahecole, line 1579: nit that I'm surprised our linter didn't catch - I think you shouldn't use #private notation per our [ts style guide](https://google.github.io/styleguide/tsguide.html#class-members). Does this helper function for sure need to be private? (It might - I just want to know more about your thought process! 😁 )

leahecole: Additional question - does this need to stay within the table class, or could it be a helper function similar to `populateAttemptHeader` down below? Tell me more! 🙂

src/table.ts
leahecole: This is fine for now, but depending on what we decide on for this eventually, I want to leave this comment open in case we need to change it later

leahecole (commented):
just a couple of nits!

src/table.ts
danieljbruce: I think doing what `populateAttemptHeader` is a good option. The con is that the function also needs to be able to accept table so needs a fourth argument. The pro is that we don't have to have a private ECMA script function that is hard to debug. It's a good suggestion overall that I was wondering about.

We definitely should hide this function from the user either way.

src/table.ts
danieljbruce: If we pull `#readRowsReqOpts` out of this class like `populateAttemptHeader` then we don't need this anymore and since modifying this `exclude` variable is not preferred, I think this is a good argument for the change you suggested above which is pulling `#readRowsReqOpts` out of this class like `populateAttemptHeader`.

src/table.ts
danieljbruce: Made this change.

PR #1442, opened 2024-07-10
fix: Fix missing rows with test plus more fixes and watermarks

danieljbruce:
Added here for logging:
https://github.com/googleapis/nodejs-bigtable/compare/main...danieljbruce:fix-missing-rows-with-test-plus-more-fixes-and-watermarks?expand=1 plus watermarks removed.


igorbernstein2:
closing in favor of #1444

PR #1443, opened 2024-07-10
fix: Fix missing rows with test and fix for node 14

danieljbruce:
Took https://github.com/googleapis/nodejs-bigtable/pull/1440
Added the test that guarantees 150 rows are all sent back and in the right order.
Modified tests to make mocks schedule last event emitted later.
Fixed test to work with Node v14


igorbernstein2:
closing in favor of #1444

PR #1444, opened 2024-07-10
fix: ensure that during resumption of a scan, rows that have not been observed by the caller are re-requested

danieljbruce:
Took https://github.com/googleapis/nodejs-bigtable/pull/1440:

createReadStream() creates a pipeline of streams that converts a stream of row chunks into a stream of logical rows. It also has logic to handle stream resumption when a single attempt fails. The pipeline can be split into 2 parts: the persistent operation stream that the caller sees and the transient per attempt segment. When a retry attempt occurs, the per attempt segment is unpiped from the operation stream and is discarded. Currently this includes any buffered data that each stream might contain. Unfortunately, when constructing the retry request, createReadStream() will use the last row key from the last buffered row. This will cause the buffered rows to be omitted from the operation stream.

This PR fixes the missing rows part by only referencing the row keys that were seen by the persistent operation stream when constructing a retry attempt. In other words, this will ensure that we only update the lastSeenRow key once the row has been "committed" to the persistent portion of the pipeline

If my understanding is correct, this should be sufficient to fix the correctness issue. However the performance issue of re-requesting the dropped buffered data remains. This should be addressed separately

Plus:

Added the test that guarantees 150 rows are all sent back and in the right order.
Modified tests to make mocks schedule last event emitted later.
Fixed test to work with Node v14
Removed watermarks

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -819,7 +819,13 @@ export class Bigtable {
     };
 
     if (isStreamMode) {
-      stream = streamEvents(new PassThrough({objectMode: true}));
+      stream = streamEvents(
+        new PassThrough({
+          objectMode: true,
+          readableHighWaterMark: 0,
+          writableHighWaterMark: 0,
+        })
+      );
       stream.abort = () => {
         if (gaxStream && gaxStream.cancel) {
           gaxStream.cancel();
--- a/src/index.ts
+++ b/src/index.ts
@@ -819,13 +819,7 @@ export class Bigtable {
     };
 
     if (isStreamMode) {
-      stream = streamEvents(
-        new PassThrough({
-          objectMode: true,
-          readableHighWaterMark: 0,
-          writableHighWaterMark: 0,
-        })
-      );
+      stream = streamEvents(new PassThrough({objectMode: true}));
       stream.abort = () => {
         if (gaxStream && gaxStream.cancel) {
           gaxStream.cancel();
--- a/src/index.ts
+++ b/src/index.ts
@@ -819,11 +819,7 @@ export class Bigtable {
     };
 
     if (isStreamMode) {
-      stream = streamEvents(
-        new PassThrough({
-          objectMode: true,
-        })
-      );
+      stream = streamEvents(new PassThrough({objectMode: true}));
       stream.abort = () => {
         if (gaxStream && gaxStream.cancel) {
           gaxStream.cancel();
--- a/src/index.ts
+++ b/src/index.ts
@@ -822,8 +822,6 @@ export class Bigtable {
       stream = streamEvents(
         new PassThrough({
           objectMode: true,
-          readableHighWaterMark: 0,
-          writableHighWaterMark: 0,
         })
       );
       stream.abort = () => {
--- a/src/table.ts
+++ b/src/table.ts
@@ -719,6 +719,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
    * region_tag:bigtable_api_table_readstream
    */
   createReadStream(opts?: GetRowsOptions) {
+    console.log('in createReadStream');
     const options = opts || {};
     const maxRetries = is.number(this.maxRetries) ? this.maxRetries! : 10;
     let activeRequestStream: AbortableDuplex | null;
--- a/src/table.ts
+++ b/src/table.ts
@@ -756,7 +756,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
     let rowsRead = 0;
     const userStream = new PassThrough({
       objectMode: true,
-      readableHighWaterMark: 0, // We need to disable readside buffering to allow for sane behavior when the end user cancels the stream early
+      readableHighWaterMark: 0, // We need to disable readside buffering to allow for acceptable behavior when the end user cancels the stream early.
       writableHighWaterMark: 0, // We need to disable writeside buffering because in nodejs 14 the call to _transform happens after write buffering. This creates problems for tracking the last seen row key.
       transform(row, _encoding, callback) {
         if (userCanceled) {
--- a/src/table.ts
+++ b/src/table.ts
@@ -917,11 +917,13 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
 
       const toRowStream = new Transform({
         transform: (rowData, _, next) => {
+          console.log(`in toRowStream ${rowData.key}`);
           if (
             userCanceled ||
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
             (userStream as any)._writableState.ended
           ) {
+            console.log('has been cancelled');
             return next();
           }
           const row = this.row(rowData.key);
--- a/src/table.ts
+++ b/src/table.ts
@@ -719,7 +719,6 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
    * region_tag:bigtable_api_table_readstream
    */
   createReadStream(opts?: GetRowsOptions) {
-    console.log('in createReadStream');
     const options = opts || {};
     const maxRetries = is.number(this.maxRetries) ? this.maxRetries! : 10;
     let activeRequestStream: AbortableDuplex | null;
@@ -918,13 +917,11 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
 
       const toRowStream = new Transform({
         transform: (rowData, _, next) => {
-          console.log(`in toRowStream ${rowData.key}`);
           if (
             userCanceled ||
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
             (userStream as any)._writableState.ended
           ) {
-            console.log('has been cancelled');
             return next();
           }
           const row = this.row(rowData.key);
--- a/src/table.ts
+++ b/src/table.ts
@@ -807,8 +807,6 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       chunkTransformer = new ChunkTransformer({
         decode: options.decode,
-        readableHighWaterMark: 0,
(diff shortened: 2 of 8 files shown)


src/index.ts
```diff
@@ -819,7 +819,11 @@ export class Bigtable {
     };
 
     if (isStreamMode) {
-      stream = streamEvents(new PassThrough({objectMode: true}));
+      stream = streamEvents(
+        new PassThrough({
+          objectMode: true,
+        })
+      );
```
igorbernstein2, line 826: please revert

src/table.ts
```diff
@@ -749,14 +749,22 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
     let userCanceled = false;
+    // The key of the last row that was emitted by the per attempt pipeline
+    // Note: this must be updated from the operation level userStream to avoid referencing buffered rows that will be
+    // discarded in the per attempt subpipeline (rowStream)
+    let lastRowKey = '';
+    let rowsRead = 0;
     const userStream = new PassThrough({
       objectMode: true,
       readableHighWaterMark: 0,
+      writableHighWaterMark: 0,
```
igorbernstein2, line 760: Add a comment:
// We need to disable readside buffering to allow for sane behavior when the end user cancels the stream early 
// We need to disable writeside buffering because in nodejs 14 the call to _transform happens after write buffering. This creates problems for tracking the last seen row key.

src/table.ts
```diff
@@ -796,9 +804,10 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       // cancelled the stream in the middle of a retry
       retryTimer = null;
 
-      const lastRowKey = chunkTransformer ? chunkTransformer.lastRowKey : '';
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      chunkTransformer = new ChunkTransformer({decode: options.decode} as any);
+      chunkTransformer = new ChunkTransformer({
+        decode: options.decode,
+      } as any);
```
igorbernstein2, line 810: please revert

test/readrows.ts
```diff
@@ -317,6 +318,52 @@ describe('Bigtable/ReadRows', () => {
...
+      return new Promise(resolve => setTimeout(resolve, ms));
+    };
+    (async () => {
+      try {
+        // Setup
+        // const bigtableClient = new Bigtable();
+        // const instance = bigtableClient.instance(INSTANCE_ID);
+        // const table = await prepareTable({instance: instance, N: 150});
+
+        // console.log('read rows');
```
igorbernstein2, line 342: please dont forget to clean this up

test/readrows.ts
```diff
@@ -317,6 +318,52 @@ describe('Bigtable/ReadRows', () => {
...
+        const stream = table.createReadStream({
+          start: '00000000',
+          end: '00000150',
+        });
+
+        for await (const row of stream) {
+          dataResults.push(row.id);
+          console.log('logging row');
+          console.log(row.id, row.data);
```
igorbernstein2, line 351: please remove before merging

test/readrows.ts
```diff
@@ -317,6 +318,52 @@ describe('Bigtable/ReadRows', () => {
...
+          await sleep(50);
+        }
+        const expectedResults = Array.from(Array(150).keys())
+          .map(i => '00000000' + i.toString())
+          .map(i => i.slice(-8));
+        assert.deepStrictEqual(dataResults, expectedResults);
+        console.log('No more data in the stream');
```
igorbernstein2, line 358: please remove before merging

test/readrows.ts
```diff
@@ -317,6 +318,52 @@ describe('Bigtable/ReadRows', () => {
...
+        done();
+      } catch (error) {
+        console.error('Something went wrong:', error);
```
igorbernstein2, line 361: please remove before merging

test/table.ts
```diff
@@ -1262,7 +1262,9 @@ describe('Bigtable/Table', () => {
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
           ((stream: any) => {
             stream.push([{key: 'a'}]);
-            stream.emit('error', makeRetryableError());
+            setImmediate(() => {
+              stream.emit('error', makeRetryableError());
+            });
```
igorbernstein2, line 1267: please extract to a helper function and add a comment why this is necessary

test/utils/readRowsImpl2.ts
```diff
@@ -0,0 +1,328 @@
...
+
+const DEBUG = process.env.BIGTABLE_TEST_DEBUG === 'true';
+
+export function debugLog(text: string) {
+  if (DEBUG) {
+    console.log(text);
+  }
+}
+
+function prettyPrintRequest(
```
igorbernstein2, line 32: please add a todo to simplify this. A lot of this complexity can be simplified, but we dont have time to do that now

igorbernstein2 (changes requested):

test/utils/readRowsImpl2.ts
danieljbruce: Done

test/table.ts
danieljbruce: Done

test/readrows.ts
danieljbruce: Done

test/readrows.ts
danieljbruce: Done

test/readrows.ts
danieljbruce: Done

test/readrows.ts
danieljbruce: Done

src/table.ts
danieljbruce: Done

src/table.ts
danieljbruce: Done.

src/index.ts
danieljbruce: Done

test/readrows.ts
```diff
@@ -317,6 +318,42 @@ describe('Bigtable/ReadRows', () => {
     });
   });
 
+  it('should return row data in the right order', done => {
+    // 1000 rows must be enough to reproduce issues with losing the data and to create backpressure
```
leahecole, line 322: is this comment about backpressure still true given that the highwatermark is set to 0? if not, please remove, otherwise, feel free to resolve

test/readrows.ts
```diff
@@ -317,6 +318,42 @@ describe('Bigtable/ReadRows', () => {
...
+    const keyFrom = undefined;
+    const keyTo = undefined;
+    // the server will error after sending this chunk (not row)
+    const errorAfterChunkNo = 100;
+    const dataResults = [];
+
+    service.setService({
+      ReadRows: readRowsImpl2(keyFrom, keyTo, errorAfterChunkNo) as any,
+    });
+    const sleep = (ms: any) => {
```
leahecole, line 332:
```suggestion
    const sleep = (ms: number) => {
```

test/readrows.ts
```diff
@@ -317,6 +318,42 @@ describe('Bigtable/ReadRows', () => {
...
+  it('should return row data in the right order', done => {
+    // 1000 rows must be enough to reproduce issues with losing the data and to create backpressure
+    const keyFrom = undefined;
+    const keyTo = undefined;
+    // the server will error after sending this chunk (not row)
+    const errorAfterChunkNo = 100;
+    const dataResults = [];
+
+    service.setService({
+      ReadRows: readRowsImpl2(keyFrom, keyTo, errorAfterChunkNo) as any,
```
leahecole, line 330: Can this be changed to something that is not "any" pretty please?

test/utils/readRowsImpl2.ts
leahecole: once this is merged, an issue should be opened to track it as well

leahecole (changes requested):

igorbernstein2 (approved):

test/readrows.ts
danieljbruce: It's still true because we want to create the scenario where there is backpressure in the chunk transformer and other streams in order to reproduce the issue that occurs when these transforms are thrown away from before the fix. Note that this fix only applies a highwatermark of 0 to the user stream.

However, this comment still does need an adjustment from 1000 to 150 :)

test/readrows.ts
sofisl: Sometimes types can be difficult in tests where we're mocking types, so maybe it can be coerced into a Type or it can be a <Partial>, or at least a comment saying the type!

test/utils/readRowsImpl2.ts
```diff
@@ -0,0 +1,330 @@
...
+      stream?.request?.rows?.rowRanges &&
+      stream?.request?.rows?.rowRanges[0] &&
+      stream?.request?.rows?.rowRanges[0]?.startKeyOpen?.toString()
+    ) {
+      keyFromRequestOpen =
+        stream?.request?.rows?.rowRanges[0]?.startKeyOpen?.toString();
+    }
+    let keyToRequestClosed: any;
+    if (
+      stream?.request?.rows?.rowRanges &&
```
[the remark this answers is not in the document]
igorbernstein2: Can we defer cleaning up the test code until after we land the fix?
I think this entire generator can be simplified significantly but I really dont want us to block a fix to a data loss bug due to test code hygiene. Daniel already added a TODO comment to simplify this

test/readrows.ts
danieljbruce: I added a TODO here. When you make this type more specific to actually match the service then there are some more complex compiler errors downstream in the mock server definition to make this work.

danieljbruce: Just made this work now with a more specific type.

leahecole (approved):
Approval, please create an issue for the cleanup concerns @sofisl and I have

test/utils/readRowsImpl2.ts
danieljbruce: Done

PR #1447, opened 2024-07-16
test: begin to refactor the mock service

danieljbruce:
**Summary:**

The ultimate goal is to address the TODOs mentioned in https://github.com/googleapis/nodejs-bigtable/blob/34d138dc5424f428d9569e62c0d05f6443dd63c4/test/utils/readRowsImpl2.ts#L161-L163. That is, we want to make the ReadRows mock service more modular and first there should ideally just be one copy of it. This PR contains initial steps towards that goal with the intent of merging more PRs in the future to arrive at a more modular mock service that is unit tested.

We are going to stack PRs onto [this branch](https://github.com/googleapis/nodejs-bigtable/tree/3527322442-to-main) and then make one final merge into main.

**Plan:**

Currently there are 2 ReadRows mock services with a bunch of code that is duplicated. Before we address the excessive amount of if statements we need to take steps to eliminate similar services `ReadRowsImpl1` and `ReadRowsImpl2`. And before we eliminate duplicate code in `ReadRowsImpl` and `ReadRowsImpl2` we should take some basic steps toward trying to eliminate `ReadRowsImpl2` entirely by allowing for additional configurations to be made to `ReadRowsImpl` so that it can be reused.

**In this PR:**

1. Change the existing mock service so that it accepts parameters for chunk size, value size and chunks per response and pass these parameters in each time a service is defined. The two services `ReadRowsImpl`, `ReadRowsImpl2` each rely on a different set of constants for chunk size, value size and chunks per response. Allowing them to be configurable in `ReadRowsImpl` generalizes `ReadRowsImpl` and is one step toward our plan of replacing `ReadRowsImpl2` with a generalized version of `ReadRowsImpl`.
2. To achieve 1 without having functions with 3+ parameters which is bad engineering practice we need to create new interfaces `ReadRowsServiceParameters` and `ChunkGeneratorParameters` so that `readRowsImpl` and  `generateChunks` only need one parameter passed in.
3. In the tests we define `const keyFrom = 0;` and `const keyTo = 1000;` many times and we can just rely on a constant for these defined at the top of the file.
4. We begin to trim code from `ReadRowsImpl2` by removing `prettyPrintRequest` because it is an exact duplicate of the `prettyPrintRequest` function in `ReadRowsImpl`.

**Next steps:**

1. Eliminate more duplicate code in `ReadRowsImpl2`.
2. Try to replace all usages of `ReadRowsImpl2` with a generalized version of `ReadRowsImpl`.
3. Address issue with too many `if` statements in `ReadRowsImpl`. Introduce classes, generate unit tests, make the service more modular.

### DIFF
--- a/test/readrows.ts
+++ b/test/readrows.ts
@@ -75,7 +75,9 @@ describe('Bigtable/ReadRows', () => {
     this.timeout(60000);
 
     service.setService({
-      ReadRows: readRowsImpl(STANDARD_SERVICE_WITHOUT_ERRORS) as any,
+      ReadRows: readRowsImpl(
+        STANDARD_SERVICE_WITHOUT_ERRORS
+      ) as ServerImplementationInterface,
     });
 
     let receivedRowCount = 0;
@@ -103,7 +105,9 @@ describe('Bigtable/ReadRows', () => {
 
   it('should create read stream and read synchronously using Transform stream', done => {
     service.setService({
-      ReadRows: readRowsImpl(STANDARD_SERVICE_WITHOUT_ERRORS) as any,
+      ReadRows: readRowsImpl(
+        STANDARD_SERVICE_WITHOUT_ERRORS
+      ) as ServerImplementationInterface,
     });
 
     let receivedRowCount = 0;
@@ -151,7 +155,9 @@ describe('Bigtable/ReadRows', () => {
       this.timeout(60000); // it runs much slower on Windows!
     }
     service.setService({
-      ReadRows: readRowsImpl(STANDARD_SERVICE_WITHOUT_ERRORS) as any,
+      ReadRows: readRowsImpl(
+        STANDARD_SERVICE_WITHOUT_ERRORS
+      ) as ServerImplementationInterface,
     });
 
     let receivedRowCount = 0;
@@ -201,7 +207,9 @@ describe('Bigtable/ReadRows', () => {
     const stopAfter = 42;
 
     service.setService({
-      ReadRows: readRowsImpl(STANDARD_SERVICE_WITHOUT_ERRORS) as any,
+      ReadRows: readRowsImpl(
+        STANDARD_SERVICE_WITHOUT_ERRORS
+      ) as ServerImplementationInterface,
     });
 
     let receivedRowCount = 0;
@@ -241,7 +249,9 @@ describe('Bigtable/ReadRows', () => {
     const stopAfter = 420;
 
     service.setService({
-      ReadRows: readRowsImpl(STANDARD_SERVICE_WITHOUT_ERRORS) as any,
+      ReadRows: readRowsImpl(
+        STANDARD_SERVICE_WITHOUT_ERRORS
+      ) as ServerImplementationInterface,
     });
 
     let receivedRowCount = 0;
@@ -301,7 +311,7 @@ describe('Bigtable/ReadRows', () => {
         chunkSize: CHUNK_SIZE,
         chunksPerResponse: CHUNKS_PER_RESPONSE,
         errorAfterChunkNo: 423,
-      }) as any,
+      }) as ServerImplementationInterface,
     });
 
     let receivedRowCount = 0;
--- a/test/readrows.ts
+++ b/test/readrows.ts
@@ -302,41 +302,53 @@ describe('Bigtable/ReadRows', () => {
     pipeline(readStream, transform, passThrough, () => {});
   });
 
-  it('should silently resume after server or network error', done => {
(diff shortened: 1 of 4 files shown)


test/readrows.ts
```diff
@@ -81,19 +95,15 @@ describe('Bigtable/ReadRows', () => {
...
   });
 
   it('should create read stream and read synchronously using Transform stream', done => {
-    // 1000 rows must be enough to reproduce issues with losing the data and to create backpressure
-    const keyFrom = 0;
-    const keyTo = 1000;
-
     service.setService({
-      ReadRows: readRowsImpl(keyFrom, keyTo) as any,
+      ReadRows: readRowsImpl(STANDARD_SERVICE_WITHOUT_ERRORS) as any,
```
leahecole, line 106: I would love it if we could have no typecasting to `any` in this PR

test/readrows.ts
```diff
@@ -22,9 +22,27 @@ import {MockServer} from '../src/util/mock-servers/mock-server';
...
 import {debugLog, readRowsImpl} from './utils/readRowsImpl';
-import {ServerWritableStream, UntypedHandleCall} from '@grpc/grpc-js';
+import {ServerWritableStream} from '@grpc/grpc-js';
 import {readRowsImpl2} from './utils/readRowsImpl2';
 
+import {ReadRowsServiceParameters} from '../test/utils/readRowsServiceParameters';
+
+// Define parameters for a standard Bigtable Mock service
+const VALUE_SIZE = 1024 * 1024;
+// we want each row to be splitted into 2 chunks of different sizes
```
leahecole, line 32:
```suggestion
// we want each row to be split into 2 chunks of different sizes
```
grammar nit

test/readrows.ts
```diff
@@ -295,14 +293,15 @@ describe('Bigtable/ReadRows', () => {
...
-
     service.setService({
-      ReadRows: readRowsImpl(keyFrom, keyTo, errorAfterChunkNo) as any,
+      ReadRows: readRowsImpl({
+        keyFrom: STANDARD_KEY_FROM,
+        keyTo: STANDARD_KEY_TO,
+        valueSize: VALUE_SIZE,
+        chunkSize: CHUNK_SIZE,
+        chunksPerResponse: CHUNKS_PER_RESPONSE,
+        errorAfterChunkNo: 423,
```
leahecole, line 303: any reason why we're using 423 or is it just an arbitrary spot to error?

test/utils/readRowsImpl2.ts
```diff
@@ -15,6 +15,7 @@
 import {ServerWritableStream} from '@grpc/grpc-js';
 import {protos} from '../../src';
 import {GoogleError, Status} from 'google-gax';
+import {prettyPrintRequest} from './readRowsImpl';
```
leahecole, line 18: (feel free to resolve) YAY FOR REUSING THINGS 😁

leahecole (changes requested):
small nits mostly. Would love @sofisl to take a glance too  if feasible

test/readrows.ts
danieljbruce: This was just the value that Alex used for this test. It looks like it is just a value that is as random as possible to make the test realistic.

sofisl: Curious if we could use `Math.random()` to test it more realistically.

test/readrows.ts
```diff
@@ -22,9 +22,27 @@ import {MockServer} from '../src/util/mock-servers/mock-server';
...
 import {MockService} from '../src/util/mock-servers/mock-service';
 import {debugLog, readRowsImpl} from './utils/readRowsImpl';
-import {ServerWritableStream, UntypedHandleCall} from '@grpc/grpc-js';
+import {ServerWritableStream} from '@grpc/grpc-js';
 import {readRowsImpl2} from './utils/readRowsImpl2';
 
+import {ReadRowsServiceParameters} from '../test/utils/readRowsServiceParameters';
+
+// Define parameters for a standard Bigtable Mock service
+const VALUE_SIZE = 1024 * 1024;
```
sofisl, line 31: what is value size represent?

danieljbruce: I added a comment for this. It is `// An upper bound on the amount of data included in the chunks`

test/readrows.ts
danieljbruce: I added an additional test to emit an error at a random position. I want to keep the old test in place too so that if it breaks then we have predictable code we can use to debug the issue.

sofisl (approved):

PR #1450, opened 2024-07-18
fix: Pin the grpc version

danieljbruce:
Pin the grpc version to address test failures in the data pipeline.


leahecole (commented):
Does it have to be in both dev dependencies and overrides?

PR #1453, opened 2024-07-19
fix: throw away excess data in order to avoid delivering duplicate data

danieljbruce:
Changes:
-  Throw away excess data if its value doesn't exceed the row key
- Fix the mock in the Table class so that the row function actually returns a more realistic value


src/table.ts
```diff
@@ -759,7 +759,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       readableHighWaterMark: 0, // We need to disable readside buffering to allow for acceptable behavior when the end user cancels the stream early.
       writableHighWaterMark: 0, // We need to disable writeside buffering because in nodejs 14 the call to _transform happens after write buffering. This creates problems for tracking the last seen row key.
       transform(row, _encoding, callback) {
-        if (userCanceled) {
+        if (userCanceled || TableUtils.lessThanOrEqualTo(row.id, lastRowKey)) {
```
igorbernstein2, line 762: Please split these into separate if statements and add a comment explaining why this is necessary

danieljbruce: Done

leahecole (approved):

PR #1457, opened 2024-07-22
test: Eliminate the second ReadRows Service and use a generalized version of the first ReadRows

danieljbruce:
**Summary:**

This PR is a follow-up to https://github.com/googleapis/nodejs-bigtable/pull/1447. It eliminates the second ReadRows service entirely.

**Background:**

To solve a high priority issue previously I created `readRowsImpl2.ts` by copy/pasting code from `readRowsImpl.ts` and making tweaks. There wasn't enough time to investigate if `readRowsImpl.ts` could be generalized to do what `readRowsImpl2.ts` was doing for us so we are addressing that technical debt now.

**Changes:**

- The second read rows service known as `readRowsImpl2.ts` is deleted entirely.
- `readRowsImpl.ts` has been generalized so that when `keyFrom` and `keyTo` are not provided then the keys from the request parameters will be used. This was how `readRowsImpl2.ts` worked.
- Any code previously using `readRowsImpl2` will now use the generalized version of `readRowsImpl` and when not providing the `keyFrom` and `keyTo` parameters then the key range in the generated chunks will be determined by the values passed into the request.
- Much like the other tests - increase the timeout so that the Windows environment has enough time to complete the test runs

**Next steps:**

1. Break down the readRowsImpl into smaller parts.
2. Address issue with too many if statements in ReadRowsImpl. Introduce classes, generate unit tests, make the service more modular.


test/utils/readRowsImpl.ts
```diff
@@ -228,7 +231,60 @@ export function readRowsImpl(
     });
 
     let chunksSent = 0;
```
danieljbruce, line 233: The entire block of changes shown below is copy/pasted from ReadRowsImpl2.ts. The next PR is going to get rid of the repetition in these if statements.

test/utils/readRowsImpl.ts
```diff
@@ -166,6 +166,9 @@ function isKeyInRowSet(
 // Returns an implementation of the server streaming ReadRows call that would return
 // monotonically increasing zero padded rows in the range [keyFrom, keyTo).
 // The returned implementation can be passed to gRPC server.
```
danieljbruce, line 168: The TODOs below are copied verbatim from readRowsImpl2.

test/utils/readRowsServiceParameters.ts
```diff
@@ -22,8 +22,8 @@ interface SharedReadRowsParameters {
 }
```
danieljbruce, line 23: `ReadRowsServiceParameters` interface is generalized so that `keyFrom` and `keyTo` do not have to be provided. When they are not provided, the parameters from the request are used.

test/utils/readRowsImpl.ts
```diff
@@ -228,7 +231,60 @@ export function readRowsImpl(
     });
 
     let chunksSent = 0;
-    const chunks = generateChunks(serviceParameters);
+    let keyFromRequestClosed: any;
```
sofisl, line 234: please do not use `any`s

sofisl (changes requested):

test/utils/readRowsImpl.ts
danieljbruce: I made changes so that the entire file doesn't contain `any` type. This entire repeated block of if statements has already been replaced with one function call in the next branch/PR anyways.

sofisl (approved):

test/readrows.ts
```diff
@@ -339,37 +338,44 @@ describe('Bigtable/ReadRows', () => {
         done();
       });
     }
-    it('with an error at a fixed position', done => {
+    it('with an error at a fixed position', function (done) {
+      if (process.platform === 'win32') {
+        this.timeout(60000); // it runs much slower on Windows!
+      }
```
leahecole, line 344: nit - could this be in a beforeeach?

leahecole (approved):
one non blocking nit

test/readrows.ts
danieljbruce: I grouped this repeated code into a `setTestTimeout` function. Unfortunately, we can't use `beforeEach` because you will notice the code uses `this` which is a reference to the test context. If we put this code into a `beforeEach` hook then `this.timeout` would set the timeout on the `beforeEach` and not on the `it`.

test/readrows.ts
```diff
@@ -71,8 +71,14 @@ describe('Bigtable/ReadRows', () => {
     service = new BigtableClientMockService(server);
   });
 
+  function setTestTimeout(test: mocha.Context) {
```
leahecole, line 74:
```suggestion
  // helper function because some tests run slower
  // on Windows and need a longer timeout
  function setWindowsTestTimeout(test: mocha.Context) {
```

test/readrows.ts
```diff
@@ -71,8 +71,14 @@ describe('Bigtable/ReadRows', () => {
...
+    if (process.platform === 'win32') {
+      test.timeout(60000); // it runs much slower on Windows!
+    }
+  }
+
   it('should create read stream and read synchronously', function (done) {
-    this.timeout(60000);
+    setTestTimeout(this);
```
leahecole, line 81:
```suggestion
    setWindowsTestTimeout(this);
```

test/readrows.ts
```diff
@@ -151,9 +157,7 @@ describe('Bigtable/ReadRows', () => {
   });
 
   it('should create read stream and read asynchronously using Transform stream', function (done) {
-    if (process.platform === 'win32') {
-      this.timeout(60000); // it runs much slower on Windows!
-    }
+    setTestTimeout(this);
```
leahecole, line 160:
```suggestion
    setWindowsTestTimeout(this);
```

test/readrows.ts
```diff
@@ -242,9 +246,7 @@ describe('Bigtable/ReadRows', () => {
 
   // TODO: enable after https://github.com/googleapis/nodejs-bigtable/issues/1286 is fixed
   it('should be able to stop reading from the read stream when reading asynchronously', function (done) {
-    if (process.platform === 'win32') {
-      this.timeout(600000); // it runs much slower on Windows!
-    }
+    setTestTimeout(this);
```
leahecole, line 249:
```suggestion
    setWindowsTestTimeout(this);
```

test/readrows.ts
```diff
@@ -339,37 +341,38 @@ describe('Bigtable/ReadRows', () => {
         done();
       });
     }
-    it('with an error at a fixed position', done => {
+    it('with an error at a fixed position', function (done) {
+      setTestTimeout(this);
```
leahecole, line 345:
```suggestion
      setWindowsTestTimeout(this);
```

test/readrows.ts
```diff
@@ -339,37 +341,38 @@ describe('Bigtable/ReadRows', () => {
...
       // Emits an error after enough chunks have been pushed to create back pressure
       runTest(done, 423);
     });
-    it('with an error at a random position', done => {
+    it('with an error at a random position', function (done) {
+      setTestTimeout(this);
```
leahecole, line 350:
```suggestion
      setWindowsTestTimeout(this);
```

test/readrows.ts
```diff
@@ -339,37 +341,38 @@ describe('Bigtable/ReadRows', () => {
...
     });
   });
-  it('should return row data in the right order', done => {
-    // 150 rows must be enough to reproduce issues with losing the data and to create backpressure
-    const keyFrom = undefined;
-    const keyTo = undefined;
-    // the server will error after sending this chunk (not row)
-    const errorAfterChunkNo = 100;
+  it('should return row data in the right order', function (done) {
+    setTestTimeout(this);
```
leahecole, line 357:
```suggestion
    setWindowsTestTimeout(this);
```

leahecole (changes requested):

leahecole (approved):

PR #1460, opened 2024-07-29
test: Eliminate repeated if statements in the ReadRows mock service reducing the size of it significantly

danieljbruce:
**Summary:**

This PR reduces all of those custom `if` statements down to one line of code in the ReadRows mock implementation cutting the ReadRows mock implementation down to about half the size so that in the next PR it is easy to make a modular version of the ReadRows mock implementation.

**Background:**

As requested, the eventual goal is to address [these TODOs](https://github.com/googleapis/nodejs-bigtable/blob/ed942c553f717d3734624b8877ee08d8eb425a46/test/utils/readRowsImpl2.ts#L161-L163). So far, two PRs have already been merged to take steps toward this:
1. [this PR](https://github.com/googleapis/nodejs-bigtable/pull/1447) eliminates duplicate code and adds interfaces to simplify data flows
2. [this PR](https://github.com/googleapis/nodejs-bigtable/pull/1456) eliminated the duplicate copy of the ReadRows service 

**Changes:**

- Instead of having a global debug log, define a debugLog and pass the debugLog into the readrows service when creating the readrows service. This involves adding a debug log to the ReadRowsServiceParameters.
- Add a DebugLog interface as it is now needed to specify the type for the debugLog dependency passed in to various functions now that the global debugLog has been eliminated.
- Add an interface `ReadRowsWritableStream` for the `ServerWritableStream<x,x>` type used everywhere for conciseness. Then replace all references to `ServerWritableStream<x,x>` with the `ReadRowsWritableStream` type.
- In readRowsImpl.ts replace all those repeated if statements with a call to `generateChunksFromRequest`. `generateChunksFromRequest` then calls a new `generateChunks` function with calculated `keyTo` and `keyFrom` values. `keyTo` and `keyFrom` are both calculated using a new `getSelectedKey` method which uses a default key if it is provided and otherwise uses the key provided in the open/close properties of the request.
- Every time a ReadRows service is defined a debugLog is passed in as now this parameter is required.

**Next Steps:**

The `readRowsImpl` function will be broken down into classes as required by the TODO.


test/readrows.ts
```diff
@@ -13,20 +13,30 @@
...
 import * as mocha from 'mocha';
 
+const DEBUG = process.env.BIGTABLE_TEST_DEBUG === 'true';
+
+function debugLog(text: string) {
+  if (DEBUG) {
+    console.log(text);
+  }
+}
+
```
danieljbruce, line 39: DebugLog is just being moved from the service to the test

test/utils/readRowsImpl.ts
```diff
@@ -12,28 +12,25 @@
...
+  ReadRowsWritableStream,
 } from './readRowsServiceParameters';
-
-const DEBUG = process.env.BIGTABLE_TEST_DEBUG === 'true';
-
-export function debugLog(text: string) {
-  if (DEBUG) {
-    console.log(text);
-  }
-}
```
danieljbruce, line 29: This has just been moved to the test and will be passed into the service.

test/utils/readRowsImpl.ts
```diff
@@ -231,60 +307,7 @@ export function readRowsImpl(
...
-        : keyToRequestClosed
-          ? parseInt(keyToRequestClosed as string)
-          : parseInt(keyToRequestOpen as string) + 1;
-    const chunks = generateChunks({
-      keyFrom: keyFromUsed,
-      keyTo: keyToUsed,
-      chunkSize: serviceParameters.chunkSize,
-      valueSize: serviceParameters.valueSize,
-    });
+    const chunks = generateChunksFromRequest(stream.request, serviceParameters);
```
danieljbruce, line 310: This single line of code is a substitute for all of the code deleted previous to it.

danieljbruce: The code for each if statement now lives in `getKeyValue`. `generateChunksFromRequest` calls `generateChunks` making 2 calls to `getSelectedKey` each making 2 calls to `getKeyValue`.

bhshkh (approved):

leahecole (approved):

danielbankhead (approved):

PR #1462, opened 2024-08-19
test: Make the ReadRows service in tests more modular

danieljbruce:
**Summary**

This PR makes the ReadRows service in tests more modular.

**Background**

This PR is made up of 4 already reviewed PRs merged onto this feature branch that are now being merged onto main.

1. [this PR](https://github.com/googleapis/nodejs-bigtable/pull/1447) eliminates duplicate code and adds interfaces to simplify data flows
2. [this PR](https://github.com/googleapis/nodejs-bigtable/pull/1456) eliminated the duplicate copy of the ReadRows service 
3. [this PR](https://github.com/googleapis/nodejs-bigtable/pull/1460) gets rid of a complex block of if statements
4. [this PR](https://github.com/googleapis/nodejs-bigtable/pull/1461) breaks the ReadRow service down into classes.


test/utils/readRowsImpl.ts
```diff
@@ -154,80 +170,216 @@ function isKeyInRowSet(
...
+  const keyRequestClosed = getKeyValue(
+    request,
+    keySelectionParameters.keyClosedProperty
+  );
+  const defaultKey = keySelectionParameters.defaultKey;
+  return defaultKey !== undefined
+    ? defaultKey
+    : keyRequestClosed
+      ? parseInt(keyRequestClosed as string)
+      : parseInt(keyRequestOpen as string) + 1;
```
sofisl, line 218: These combined ternary expressions are hard to read. Is this saying if the defaultKey is not undefined, then use default key, else use keyRequestClosed, and if that is undefined use it as opened, or else as closed? I think this might achieve the same thing and be easier to read, or if not maybe you can rewrite it in a way that preserves the logic:
```
return defaultKey || keyRequestClosed ? parseInt(keyRequestClosed as string) : parseInt(keyRequestOpen as string) + 1;
```

test/utils/readRowsImpl.ts
```diff
@@ -154,80 +170,216 @@ function isKeyInRowSet(
...
+        }
+        if (response.lastScannedRowKey) {
+          const binaryKey = Buffer.from(
+            response.lastScannedRowKey as string,
+            'base64'
+          );
+          const stringKey = binaryKey.toString();
+          debugLog(`sent lastScannedRowKey = ${stringKey}`);
+        }
+        if (!canSendMore) {
```
sofisl, line 295: what's the difference between lastScannedRowKey and canSendMore? Maybe a comment describing the branches would be useful. Also, is it possible to have a lastScannedRowKey and !canSendMore? Otherwise perhaps an else might be useful.

danieljbruce: From [this](https://nodejs.org/api/stream.html#writablewritechunk-encoding-callback), `canSendMore` is the value returned indicating whether or not we want the stream to drain. `lastScannedRowKey` is actually part of the response so I think both the if statements are independent.

danieljbruce: This code is pretty much an exact copy of what we have on main: https://github.com/googleapis/nodejs-bigtable/blob/ed942c553f717d3734624b8877ee08d8eb425a46/test/utils/readRowsImpl.ts#L191-L210

test/utils/readRowsImpl.ts
danieljbruce: We can't do it quite like this because when defaultKey is 0 it is falsy so it doesn't get used, but we want it to get used when it is 0. What I did instead was did an equal check instead of a not equal check.

test/utils/readRowsImpl.ts
danieljbruce: I added comments for the if blocks

danielbankhead (approved):

cindy-peng (approved):

bhshkh (approved):

PR #1463, opened 2024-08-22
feat: Bigtable authorized views - move the code over to the TabularApiSurface class

danieljbruce:
**Summary:**

This PR is the first in the series to support the plan from [go/node-cbt-authorized-views](https://goto.google.com/node-cbt-authorized-views). Code from the `Table` class is moved over to the `TabularApiSurface` class in preparation for future PRs that will make changes to this `TabularApiSurface` code.

**Changes:**

- Code has been moved over from `table.ts` to `tabular-api-service.ts` from the `Table` class to the `TabularApiService` class. `TabularApiSurface` now contains the `createReadStream`, `getRows`, `insert`, `mutate`, `row`, `sampleRowKeysStream`, `sampleRowKeys` methods. `Table` inherits from `TabularApiSurface` so `Table` still does the same thing as before, but when we make the code changes later we will be able to see the changes in isolation because the code from `Table` has been moved over already.
- `src/row.ts` changes - The object passed into the Row is now a TabularAPISurface so row.ts gets interface changes to work with this kind of object
- `src/utils/table.ts` changes - some items now live in `TabularAPISurface` so they get imported from there instead
- `test/table.ts` changes - Some mocks using `proxyquire` must now be on `tabular-api-service.js` instead of `../src/table.js` in order to emulate the same behaviour using the mocks.

**Next Steps:**

Next we will add all the functions that will be needed on the `TabularAPISurface` class. `TabularAPISurface` class contains the old `Table` functions it needs to contain, but doesn't contain functions that will use `Row` that will be necessary for the `View`. In the next PR we won't introduce the `View` yet, we will just add all the missing functions to `TabularAPISurface` so that we know it is complete.


src/row.ts
```diff
@@ -156,13 +157,13 @@ export class RowError extends Error {
...
   bigtable: Bigtable;
-  table: Table;
+  table: TabularApiService;
   id: string;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   data: any;
   key?: string;
   metadata?: {};
-  constructor(table: Table, key: string) {
+  constructor(table: TabularApiService, key: string) {
```
daniel-sanche, line 166: Is this meant to be `TabularAPISurface` or `TabularAPIService`?

danieljbruce: Good catch. It should be TabularAPISurface. I changed this everywhere now.

test/table.ts
```diff
@@ -110,14 +110,21 @@ describe('Bigtable/Table', () => {
...
       pumpify,
       './row.js': {Row: FakeRow},
       './chunktransformer.js': {ChunkTransformer: FakeChunkTransformer},
+    }).TabularApiSurface;
+    Table = proxyquire('../src/table.js', {
+      '@google-cloud/promisify': fakePromisify,
+      './family.js': {Family: FakeFamily},
+      './mutation.js': {Mutation: FakeMutation},
+      './row.js': {Row: FakeRow},
+      './tabular-api-surface': {TabularApiService: FakeTabularApiSurface},
```
daniel-sanche, line 127: still a couple "services" here

daniel-sanche (approved):
LGTM after replacing service with surface

test/table.ts
danieljbruce: Yup. I noticed the tests failing and then adjusted these.

danieljbruce: I'll have to investigate the reason the most recent tests are failing.

daniel-sanche: It looks like there's still at least one :) 

(TabularApiService)

danieljbruce: Like you pointed out... It looks like `TabularApiService` still needs to be renamed.

PR #1501, opened 2024-10-09
feat: Create the unit tests for Authorized Views ensuring that requests are consistent

danieljbruce:
**Summary**

This PR adds unit tests for the Authorized Views implementation and fixes some bugs that the unit tests reveal. All changes to the source code are annotated with a comment that explains the change that was made. The unit tests ensure that when a grpc request is made for an authorized view that it would be exactly the same as it would be for a table with the only change being that it would target an authorized view instead.

**Changes**

- The AuthorizedView class is promisified much like the other classes in the client library are so that users can get results from a call by providing a callback or by using the returned promise.
- view which is the function that gets an AuthorizedView is not promisified
- The `append` property does not need to be provided anymore when calling createRules
- TabularApiSurface will now make requests that target Authorized Views
- A new file `test/authorized-views` is added that ensures that for each method requests made for tables are consistent with requests made for authorized views .


src/authorized-view.ts
```diff
@@ -253,3 +254,7 @@ export class AuthorizedView extends TabularApiSurface {
     }
   }
 }
+
+promisifyAll(AuthorizedView, {
+  exclude: ['initializeRow', 'generateProperties'],
+});
```
danieljbruce, line 260: Much like the `Table` and `Row` classes, `AuthorizedView` must use `promisify`. This is because the body of the methods in `AuthorizedView` don't return anything and they use a callback to send back results, but the methods for `AuthorizedView` should return promises like the methods for `Table` or `Row` do.

`initializeRow` and `generateProperties` are just helper methods and exist to eliminate duplicate code so should not be transformed to return promises.

src/instance.ts
```diff
@@ -1501,6 +1501,7 @@ promisifyAll(Instance, {
     'getBackupsStream',
     'getTablesStream',
     'getAppProfilesStream',
+    'view',
```
danieljbruce, line 1504: This needs to be added because calling `view` should return an Authorized View. It should not return a promis that resolves to an authorized view.

src/row.ts
```diff
@@ -34,7 +34,7 @@ import {TabularApiSurface} from './tabular-api-surface';
 
 export interface Rule {
   column: string;
-  append: string;
+  append?: string;
```
danieljbruce, line 37: append shouldn't actually be required and the reason I know this is that while developing, I noticed the increment method which works properly calls createRules providing a `Rule` object without an `append` property.

src/tabular-api-surface.ts
```diff
@@ -306,10 +306,17 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+        this.viewName
+          ? {
+              authorizedViewName: `${this.name}/authorizedViews/${this.viewName}`,
+              appProfileId: this.bigtable.appProfileId,
+            }
+          : {
+              tableName: this.name,
+              appProfileId: this.bigtable.appProfileId,
+            }
+      ) as google.bigtable.v2.IReadRowsRequest;
```
danieljbruce, line 319: This change makes it so that when viewName is defined, ie. when the object is an authorized view that the request will be made for an authorized view and not a table.

src/tabular-api-surface.ts
```diff
@@ -674,13 +681,21 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+      const baseReqOpts = (
+        this.viewName
+          ? {
+              authorizedViewName: `${this.name}/authorizedViews/${this.viewName}`,
+            }
+          : {
+              tableName: this.name,
+            }
+      ) as google.bigtable.v2.IReadRowsRequest;
+      const reqOpts = Object.assign(baseReqOpts, {
```
danieljbruce, line 693: This change makes it so that when viewName is defined, ie. when the object is an authorized view that the request will be made for an authorized view and not a table.

danieljbruce (commented):
Add comments inline to explain code.

src/tabular-api-surface.ts
daniel-sanche: Do you think a comment like this would be helpful in the code?

src/tabular-api-surface.ts
daniel-sanche: Same as above: maybe explain this with a comment?

daniel-sanche (approved):
new tests LGTM, but consider adding some extra comments in the code where you added comments here. I'm not sure how obvious some of this will be in the future

src/tabular-api-surface.ts
danieljbruce: Yup. Added a comment.

src/tabular-api-surface.ts
danieljbruce: Done

PR #1504, opened 2024-10-17
feat: Bigtable authorized views integration tests

danieljbruce:
**Summary**

This PR adds integration tests for the Authorized Views feature. The integration tests ensure that for Authorized Views the calls to `readRows`, `mutateRows`, `sampleRowKeys`, `checkAndMutateRows` and `readModifyWriteRows` perform the right read/write operations on the authorized view. There are also a few tests in place that try to modify data and user doesn't have access to and ensure an error is thrown.

In addition to the tests there are a couple of minor changes to the source code mentioned inline that were found when writing the tests.


src/table.ts
```diff
@@ -260,7 +260,7 @@ export type SampleRowKeysCallback = (
   err: ServiceError | null,
   keys?: string[]
 ) => void;
-export type SampleRowsKeysResponse = [string[]];
+export type SampleRowsKeysResponse = [{key: Uint8Array; offset: string}[]];
```
danieljbruce, line 263: The response `[string[]]` does not actually match what the server sends back.

src/row.ts
```diff
@@ -58,6 +58,7 @@ export interface Family {
 export interface FilterConfigOption {
   method?: string;
   data?: Data;
+  key?: string;
```
danieljbruce, line 61: [this code sample](https://github.com/googleapis/nodejs-bigtable/blob/86c8f4b96668a91f01bda2c75356b1ed12828269/samples/deleteSnippets.js#L88-L103) demonstrates that `key` should be a property of this interface.

daniel-sanche (approved):

PR #1511, opened 2024-11-01
test: log conformance test failures in failure logs

kevkim-codes:
Add conformance known test failure logs and clean up kokoro script


danieljbruce (approved):
This seems like a value add, but I wonder if there is a way in each PR we could actually automatically log and see which conformance tests are failing so that we can see the progress over time without any more manual changes.

PR #1522, opened 2024-11-26
test: Investigate the TestReadRows_Retry_StreamReset test

danieljbruce:
**Summary**

This pull request should not be merged. It is created so that I can share logs about what happens when the TestReadRows_Retry_StreamReset test is run. It looks like Gapic gives us an error with a grpc code of 1 instead of a retryable grpc code. The test needs a retryable grpc code in order to pass.


danieljbruce:
This was opened for experimental purposes

PR #1529, opened 2024-11-29
test: Prepare helper function for readWriteModifyRow test proxy service

danieljbruce:
**Summary:**

To prepare for future fixes to the readModifyWriteRow service in the test proxy, we create a `getRMWRRequest ` function that builds a readModifyWrite Gapic request using the same logic from the client library. We also create a `getRMWRRequestInverse` function that does the opposite of `getRMWRRequest ` and we verify that these two functions are opposites/inverses using tests. `getRMWRRequestInverse` will be used in the test proxy to change Gapic requests to handwritten layer requests.

**Background:**

As part of a [broader effort to address conformance tests](https://docs.google.com/document/d/1LU7aY756rBEEbHHVw7213HfI6EfKBa_ql6e-Q5e1Oo4/edit?tab=t.0#heading=h.xzptrog8pyxf), the readModifyWriteRow test proxy service needs to [call the handwritten in order for its tests to be valid](https://docs.google.com/document/d/1AAaH08Nb33GnYBOfYHbh9TsCNsZuCZqv8rrg2J6cVuU/edit?tab=t.0#heading=h.xzptrog8pyxf). To call the handwritten layer, it needs a request that is suitable for the handwritten layer so it needs a function created in this PR called `getRMWRRequestInverse` to transform a Gapic ready request passed into the test proxy service into a request for the handwritten layer. When this request is passed into the handwritten layer and a gapic ready request reaches the readModifyWriteRow method in the Gapic layer then the Gapic ready request should be equivalent to the request passed into the readModifyWriteRow test proxy service. This is necessary in order for the readModifyWriteRow test proxy service to work as intended.

**Changes**

**`testproxy/services/utils/request/readModifyWriteRow.ts`**: A `getRMWRRequest` method is created for building a Gapic request from data passed into the handwritten layer that would be sent to the `readModifyWriteRow` Gapic method. This contains code equivalent to code that lives in the `createRulesUtil` function. A `getRMWRRequestInverse` method is also added here intending to be the inverse of the `getRMWRRequest` method which will be used by the test proxy service to change Gapic requests to handwritten layer requests.

**`test/test-proxy/readModifyWriteRow.ts`**: Some tests are added that guarantee `getRMWRRequest` and `getRMWRRequestInverse` are inverses.

**Next Steps:**

Make the `readModifyWriteRow` test proxy service use the `getRMWRRequestInverse` method and the handwritten layer of the client library so that the test proxy service is actually testing the handwritten layer as intended.

### DIFF
--- a/src/row-data-utils.ts
+++ b/src/row-data-utils.ts
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {getRMWRRequest} from './utils/readModifyWriteRow/getRMWRRequest';
+
 const dotProp = require('dot-prop');
 import {Filter, RawFilter} from './filter';
 import {
@@ -161,37 +163,12 @@ class RowDataUtils {
     const callback =
       typeof optionsOrCallback === 'function' ? optionsOrCallback : cb!;
 
-    if (!rules || (rules as Rule[]).length === 0) {
-      throw new Error('At least one rule must be provided.');
-    }
-
-    rules = arrify(rules).map(rule => {
-      const column = Mutation.parseColumnName(rule.column);
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      const ruleData: any = {
-        familyName: column.family,
-        columnQualifier: Mutation.convertToBytes(column.qualifier!),
-      };
-
-      if (rule.append) {
-        ruleData.appendValue = Mutation.convertToBytes(rule.append);
-      }
-
-      if (rule.increment) {
-        ruleData.incrementAmount = rule.increment;
-      }
-
-      return ruleData;
+    const reqOpts = getRMWRRequest({
+      reqOpts: properties.reqOpts,
+      id: properties.requestData.id,
+      rules,
+      appProfileId: properties.requestData.bigtable.appProfileId,
     });
-
-    const reqOpts = Object.assign(
-      {
-        appProfileId: properties.requestData.bigtable.appProfileId,
-        rowKey: Mutation.convertToBytes(properties.requestData.id),
-        rules,
-      },
-      properties.reqOpts
-    );
     properties.requestData.data = {};
     properties.requestData.bigtable.request<google.bigtable.v2.IReadModifyWriteRowResponse>(
       {
--- a/src/row-data-utils.ts
+++ b/src/row-data-utils.ts
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {getRMWRRequest} from './utils/readModifyWriteRow/getRMWRRequest';
-
 const dotProp = require('dot-prop');
 import {Filter, RawFilter} from './filter';
 import {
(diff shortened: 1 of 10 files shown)


src/row-data-utils.ts
```diff
@@ -161,37 +163,12 @@ class RowDataUtils {
...
-        ruleData.incrementAmount = rule.increment;
-      }
-
-      return ruleData;
+    const reqOpts = getRMWRRequest({
+      reqOpts: properties.reqOpts,
+      id: properties.requestData.id,
+      rules,
+      appProfileId: properties.requestData.bigtable.appProfileId,
     });
```
danieljbruce, line 171: All the code removed in this file now lives in getRMWRRequest and one call to getRMWRRequest will be made to build the request.

daniel-sanche (changes requested):
Do you think you can move this gapic<->veneer translation code into the test proxy codebase? I have concerns that this change adds more layers of abstraction, complexity, and room for bugs to the library, just to support tests

But my larger concern is that IMO the test proxy should only interact with the library's public API, never share internal helpers. Otherwise we run the risk that bugs slip through, because both the library and the test proxy are using the same faulty implementation. There should be a clear wall between the library and the tests

I think the test proxy's role should be:
1. receive gapic request from test runner
2. convert into hand-written representation for interacting with the library
3. execute the rpc on the library
4. convert the results back to a gapic representation
5. send results to test runner

But let me know if you think I'm missing something

danieljbruce:
> Do you think you can move this gapic<->veneer translation code into the test proxy codebase? I have concerns that this change adds more layers of abstraction, complexity, and room for bugs to the library, just to support tests
> 
> But my larger concern is that IMO the test proxy should only interact with the library's public API, never share internal helpers. Otherwise we run the risk that bugs slip through, because both the library and the test proxy are using the same faulty implementation. There should be a clear wall between the library and the tests
> 
> I think the test proxy's role should be:
> 
> 1. receive gapic request from test runner
> 2. convert into hand-written representation for interacting with the library
> 3. execute the rpc on the library
> 4. convert the results back to a gapic representation
> 5. send results to test runner
> 
> But let me know if you think I'm missing something

I think (1), (2) and (3) are correct if by 3 you mean call the handwritten method in the client library (createRules in this case). 4 and 5 are irrelevant for this PR since this PR only touches request code.

I definitely have no objections to moving `getRMWRRequestInverse` into the test proxy folder since it's only being used by the test proxy anyway. I can undo all the changes to `src/row-data-utils.ts` if you like, I realize it's less risky to make no changes to the `src` folder, but I also think it's more modular if the request construction is a separate function. I think there's a broader discussion about code style on that topic.

daniel-sanche (approved):

daniel-sanche:
>  I can undo all the changes to `src/row-data-utils.ts` if you like, I realize it's less risky to make no changes to the `src` folder, but I also think it's more modular if the request construction is a separate function. I think there's a broader discussion about code style on that topic.

I don't have context on the other changes in `src/row-data-utils.ts`, but in general I don't think we want to be sharing internal code between the test proxy and the library if at all possible

PR #1531, opened 2024-12-03
chore(testproxy): Change the readModifyWriteRow test proxy so that it uses the handwritten layer

danieljbruce:
**Summary:**

Currently the test proxy calls the readModifyWriteRow method in the Gapic layer, but [it should use the veneer](https://docs.google.com/document/d/1AAaH08Nb33GnYBOfYHbh9TsCNsZuCZqv8rrg2J6cVuU/edit?tab=t.0#heading=h.xzptrog8pyxf). This PR includes the changes required to make the readModifyWriteRow test proxy service use the veneer with a properly formed request.

**Background:**

A [previous PR](https://github.com/googleapis/nodejs-bigtable/pull/1529) adds a helper method for changing the request passed into the test proxy readModifyWriteRow service into a request suitable for the handwritten layer. In this PR, that helper method is used by the readModifyWriteRow service to form a request for the handwritten layer and the handwritten layer is called is called by this test proxy service as required.

**Changes:**

**`test/test-proxy/readModifyWriteRow.ts:`** This file ensures that the helper method for changing the Gapic request to a format for the handwritten layer prepares the request in the proper format.

**`test/test-proxy/readModifyWriteRowService.ts:`** This test file ensures that the Gapic shaped request passed into the test proxy service is exactly the same as the request that reaches the Gapic layer after being transformed and passed through the handwritten layer.

**`testproxy/services/read-modify-write-row.js:`** The test proxy service is changed to use the handwritten layer.

**`testproxy/services/request/readModifyWriteRow.ts:`** This file contains a `getRMWRRequest` function which is equivalent to how the client builds a readModifyWriteRequest and a `getRMWRRequestInverse` function that prepares a Gapic layer shaped request for the handwritten layer. `getRMWRRequestInverse` does the inverse of `getRMWRRequest` so that when requests pass through `getRMWRRequestInverse` and then `getRMWRRequest` they remain unchanged.

**Next Steps:**

Now we can address some of the test proxy tests that are failing for this service.


testproxy/services/read-modify-write-row.js
```diff
@@ -16,24 +16,24 @@
...
-    const [result] = await client.readModifyWriteRow({
-      appProfileId,
-      rowKey,
-      rules,
-      tableName,
-    });
-
+    const {clientId, request: readModifyWriteRow} = request;
+    const {appProfileId, tableName} = readModifyWriteRow;
+    const handWrittenRequest = getRMWRRequestInverse(readModifyWriteRow);
```
danieljbruce, line 31: This is where the inverse method is used.

danieljbruce:
Opening this on https://github.com/googleapis/nodejs-bigtable/pull/1535 instead of forked repo.

PR #1533, opened 2024-12-04
chore(testproxy): Address the TestReadRows_Generic_DeadlineExceeded test

danieljbruce:
**Summary:**

This change allows the TestReadRows_Generic_DeadlineExceeded conformance test to pass by not retrying on DEADLINE_EXCEEDED errors when the client timeout has been exceeded.

**Changes:**

**`src/tabular-api-service.ts`:** The changes in this file make it so that readRows calls get the timeout value that was passed into the client, record the time the call was made and then only retry on DEADLINE_EXCEEDED errors when the amount of time that has passed is less than the timeout. Longer term this change should be moved downstream into google-gax.

**`testproxy/services/create-client.js`:** A change is made to use the perOperationTimeout passed into the create-client test proxy service to set the timeout parameter on the client so that it will be used by readRows calls.

**`testproxy/services/read-rows.js`:** The readRows test proxy service is modified so that the status is actually passed to the test runner instead of Nil.

**`test/table.ts`:** A unit test is added that ensures timeouts can be set per call.

**Next Steps:**

- In a separate PR we should allow the timeout to apply to all methods
- We should move the timeout logic upstream to google-gax and emit an error there when the client times out instead of relying on custom logic in the handwritten layer to track a timeout when handling DEADLINE_EXCEEDED errors from the server.

### DIFF
--- a/src/tabular-api-surface.ts
+++ b/src/tabular-api-surface.ts
@@ -538,9 +538,17 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
           if (
             numConsecutiveErrors <= maxRetries &&
             (RETRYABLE_STATUS_CODES.has(error.code) || isRstStreamError(error))
+            !(error.code === grpc.status.DEADLINE_EXCEEDED && timeout < new Date().getTime() - callTimeMillis)
           ) {
             retryTimer = setTimeout(makeNewRequest, nextRetryDelay);
           } else {
+            const backOffSettings =
+              options.gaxOptions?.retry?.backoffSettings ||
+              DEFAULT_BACKOFF_SETTINGS;
+            const nextRetryDelay = getNextDelay(
+              numConsecutiveErrors,
+              backOffSettings
+            );
             if (
               !error.code &&
               error.message === 'The client has already been closed.'
--- a/src/tabular-api-surface.ts
+++ b/src/tabular-api-surface.ts
@@ -537,15 +537,20 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
           numRequestsMade++;
           if (
             numConsecutiveErrors <= maxRetries &&
-            (RETRYABLE_STATUS_CODES.has(error.code) || isRstStreamError(error)) &&
-            !(timeout && error.code === grpc.status.DEADLINE_EXCEEDED && timeout < new Date().getTime() - callTimeMillis)
+            (RETRYABLE_STATUS_CODES.has(error.code) ||
+              isRstStreamError(error)) &&
+            !(
+              timeout &&
+              error.code === grpc.status.DEADLINE_EXCEEDED &&
+              timeout < new Date().getTime() - callTimeMillis
+            )
           ) {
             const backOffSettings =
-                options.gaxOptions?.retry?.backoffSettings ||
-                DEFAULT_BACKOFF_SETTINGS;
+              options.gaxOptions?.retry?.backoffSettings ||
+              DEFAULT_BACKOFF_SETTINGS;
             const nextRetryDelay = getNextDelay(
-                numConsecutiveErrors,
-                backOffSettings
+              numConsecutiveErrors,
+              backOffSettings
             );
             retryTimer = setTimeout(makeNewRequest, nextRetryDelay);
           } else {
--- a/src/tabular-api-surface.ts
+++ b/src/tabular-api-surface.ts
@@ -231,27 +231,10 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
     downstream in google-gax.
      */
     const timeout =
-      this.bigtable &&
-      this.bigtable.options &&
-      this.bigtable.options.BigtableClient &&
-      this.bigtable.options.BigtableClient.clientConfig &&
-      this.bigtable.options.BigtableClient.clientConfig.interfaces &&
-      this.bigtable.options.BigtableClient.clientConfig.interfaces[
+      this?.bigtable?.options?.BigtableClient?.clientConfig?.interfaces &&
+      this?.bigtable?.options?.BigtableClient?.clientConfig?.interfaces[
         'google.bigtable.v2.Bigtable'
-      ] &&
-      this.bigtable.options.BigtableClient.clientConfig.interfaces[
-        'google.bigtable.v2.Bigtable'
-      ].methods &&
-      this.bigtable.options.BigtableClient.clientConfig.interfaces[
-        'google.bigtable.v2.Bigtable'
-      ].methods['ReadRows'] &&
-      this.bigtable.options.BigtableClient.clientConfig.interfaces[
-        'google.bigtable.v2.Bigtable'
-      ].methods['ReadRows'].timeout_millis
-        ? this.bigtable.options.BigtableClient.clientConfig.interfaces[
-            'google.bigtable.v2.Bigtable'
-          ].methods['ReadRows'].timeout_millis
-        : undefined;
+      ]?.methods['ReadRows']?.timeout_millis;
     const callTimeMillis = new Date().getTime();
 
     const ranges = TableUtils.getRanges(options);
--- a/src/tabular-api-surface.ts
+++ b/src/tabular-api-surface.ts
@@ -537,18 +537,18 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
           numRequestsMade++;
           if (
             numConsecutiveErrors <= maxRetries &&
-            (RETRYABLE_STATUS_CODES.has(error.code) || isRstStreamError(error))
+            (RETRYABLE_STATUS_CODES.has(error.code) || isRstStreamError(error)) &&
             !(error.code === grpc.status.DEADLINE_EXCEEDED && timeout < new Date().getTime() - callTimeMillis)
           ) {
+            const backOffSettings =
+                options.gaxOptions?.retry?.backoffSettings ||
+                DEFAULT_BACKOFF_SETTINGS;
+            const nextRetryDelay = getNextDelay(
+                numConsecutiveErrors,
+                backOffSettings
+            );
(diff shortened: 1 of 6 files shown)


testproxy/services/read-rows.js
```diff
@@ -72,7 +72,13 @@ const readRows = ({clientMap}) =>
...
       };
     } catch (e) {
-      return {status: e};
+      return {
+        status: {
+          code: e.code,
+          details: [],
+          message: e.message,
+        },
+      };
```
danieljbruce, line 81: The test proxy service was not sending back the status because the details weren't in an array format. This made it impossible for the test to check the code and message because status was nil.

src/tabular-api-surface.ts
```diff
@@ -224,6 +224,36 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+      this.bigtable.options.BigtableClient.clientConfig.interfaces[
+        'google.bigtable.v2.Bigtable'
+      ].methods['ReadRows'] &&
+      this.bigtable.options.BigtableClient.clientConfig.interfaces[
+        'google.bigtable.v2.Bigtable'
+      ].methods['ReadRows'].timeout_millis
+        ? this.bigtable.options.BigtableClient.clientConfig.interfaces[
+            'google.bigtable.v2.Bigtable'
+          ].methods['ReadRows'].timeout_millis
+        : undefined;
```
leahecole, line 254: This is very hard to read - can you use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to simplify it? 

something like `this.bigtable?.options?.BigtableClient?.clientConfig?.interfaces[
            'google.bigtable.v2.Bigtable'
          ]?.methods['ReadRows']?.timeout_millis?`

(not sure how it works with the square bracket syntax but I think it can get rid of at least a few lines

testproxy/services/read-rows.js
leahecole: does `e` have details that could be surfaced? Or is details truly empty?

leahecole (changes requested):
also left a comment in buganizer

src/tabular-api-surface.ts
danieljbruce: I just added optional chaining to shorten this.

src/tabular-api-surface.ts
```diff
@@ -224,6 +224,19 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
 
     rowKeys = options.keys || [];
 
+    /*
+    The following line of code sets the timeout if it was provided while
+    creating the client. This will be used to determine if the client should
+    retry on DEADLINE_EXCEEDED errors. Eventually, this will be handled
+    downstream in google-gax.
+     */
+    const timeout =
```
mutianf, line 233: I think timeout should come from [getRowsOptions](https://github.com/googleapis/nodejs-bigtable/blob/main/src/tabular-api-surface.ts#L90) gaxOptions -> timeout https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html#timeout

src/tabular-api-surface.ts
```diff
@@ -507,7 +520,13 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
           numRequestsMade++;
           if (
             numConsecutiveErrors <= maxRetries &&
-            (RETRYABLE_STATUS_CODES.has(error.code) || isRstStreamError(error))
+            (RETRYABLE_STATUS_CODES.has(error.code) ||
+              isRstStreamError(error)) &&
+            !(
+              timeout &&
+              error.code === grpc.status.DEADLINE_EXCEEDED &&
```
mutianf, line 527: The error code doesn't matter here. Client should fail with deadline exceeded error after the total timeout expires.

mutianf (changes requested):

src/tabular-api-surface.ts
danieljbruce: In this conformance test the timeout is provided in the create-client test proxy service, not the readRows test proxy service. So for the conformance test to pass we have to read the timeout from the client. See the screenshots below for proof of this.

image
image

That said, I understand that if a timeout is provided in `GetRowOptions` then it should be used instead of the timeout set in the client so I can change that.

mutianf: We're not just trying to make the test pass, we're trying to verify that the client behaves as expected. Will a customer be able to configure the overall timeout on the client directly? If not, we should always use the timeout from `GetRowsOptions` as the source of truth, and figure out how we can fix the test proxy so the client opts from conformance test is passed in correctly.

danieljbruce: > Will a customer be able to configure the overall timeout on the client directly?

In the way that I had it before, yes. They do it by setting the clientConfig when creating the client [like so](https://github.com/googleapis/nodejs-bigtable/pull/1533/files#diff-167461f4f7152422b73f5542871adc43acb88b564b420fde31736a3442168ef0L62). That said, I have made the change so that the timeout can be overridden for each call as you suggested.

Note right now, that the timeout is not being passed into the readRows test proxy service at all. So the test runner needs to change to pass the timeout along in the readRows call if you want to see that this timeout is being is used from `GetRowOptions`.

src/tabular-api-surface.ts
danieljbruce: Sounds good. Now that it is understood that it is a requirement that the timeout should be respected for all error codes so I've removed this line accordingly.

src/tabular-api-surface.ts
danieljbruce: Note that the test will still pass because it looks for `timeout` passed into the method and if it is `undefined` it will use the timeout provided when the client was created which happens to be 2 seconds for this test. So the test runner doesn't need to change now for this test, but maybe a test for that `timeout` parameter should be added in the future to the test runner.

```
      opts?.gaxOptions?.timeout ||
      (this?.bigtable?.options?.BigtableClient?.clientConfig?.interfaces &&
        this?.bigtable?.options?.BigtableClient?.clientConfig?.interfaces[
          'google.bigtable.v2.Bigtable'
        ]?.methods['ReadRows']?.timeout_millis);
```

mutianf: I see that the gaxOption is hard to test with the current conformance test set up. Can we create a unit test to test this timeout is used?

danieljbruce: Sure. Mocking errors and responses from a stream is somewhat cumbersome so I'll just need a bit of time to get that set up.

mutianf (approved):

test/table.ts
```diff
@@ -1120,7 +1120,28 @@ describe('Bigtable/Table', () => {
...
+      // The timeout is 2 seconds, but the error is received after 3 seconds
+      // so the client doesn't retry because more than 2 seconds have elapsed.
+      const requestSpy = (table.bigtable.request = sinon.spy(() => {
+        const stream = new PassThrough({
+          objectMode: true,
+        });
+        (stream as any).abort = () => {};
+        setTimeout(() => {
+          const error = new Error('retry me!') as ServiceError;
+          error.code = 4;
```
leahecole, line 1133:
```suggestion
          error.code = 4; 
```
Suggest using a different error code than the one you're expecting to make the test more robust

test/table.ts
```diff
@@ -1120,7 +1120,28 @@ describe('Bigtable/Table', () => {
...
+          stream.emit('error', error);
+        }, 3000);
+        return stream;
+      }));
+      const stream = table.createReadStream({gaxOptions: {timeout: 2000}});
+      stream.on('error', (error: ServiceError) => {
+        assert.strictEqual(error.code, 4);
```
leahecole, line 1140:
```suggestion
        assert.strictEqual(error.code, 4);
        assert.notStrictEqual(error.message, "retry me!");
```

just another check for posterity

testproxy/services/create-client.js
```diff
@@ -54,12 +61,33 @@ const createClient = ({clientMap}) =>
...
+      /**
+       * This block of code ensures the server times out for every method call
+       * after the amount of time specified in request.perOperationTimeout.
+       */
+      Object.entries(
+        clientConfig.interfaces['google.bigtable.v2.Bigtable'].methods
+      ).forEach(([k, v]) => {
+        if (k === 'ReadRows') {
+          /*
+          TODO: In the future we should apply this for all methods, but right
```
leahecole, line 74: If a tracking bug isn't open for this in buganizer please open one!

testproxy/services/read-rows.js
```diff
@@ -72,7 +72,13 @@ const readRows = ({clientMap}) =>
         rows: rows.map(getRowResponse),
       };
     } catch (e) {
-      return {status: e};
+      return {
+        status: {
+          code: e.code,
+          details: [],
```
leahecole, line 78:
```suggestion
          details: [], // e.details must be in an empty array for the test runner to return the status, work to fix the test runner to properly return details is tracked in b/xxxxxxx
```

testproxy/services/read-rows.js
leahecole: It sounds like an underlying issue with the test runner that's not critical to this test - I made a suggestion to add a comment and open a buganizer bug that you then reference in the comment

leahecole (approved):
Approval once changes are addressed!

testproxy/services/create-client.js
danieljbruce: https://b.corp.google.com/issues/383094924

test/table.ts
danieljbruce: I added another test so 4 and 14 retryable error codes are tested.

testproxy/services/read-rows.js
danieljbruce: Committed and https://b.corp.google.com/issues/383096533 added.

testproxy/services/read-rows.js
danieljbruce: Sounds good

PR #1535, opened 2024-12-05
chore(testproxy): Change the readModifyWriteRow test proxy so that it uses the handwritten layer

danieljbruce:
**Summary:**

Currently the test proxy calls the readModifyWriteRow method in the Gapic layer, but [it should use the veneer](https://docs.google.com/document/d/1AAaH08Nb33GnYBOfYHbh9TsCNsZuCZqv8rrg2J6cVuU/edit?tab=t.0#heading=h.xzptrog8pyxf). This PR includes the changes required to make the readModifyWriteRow test proxy service use the veneer with a properly formed request.

**Background:**

A [previous PR](https://github.com/googleapis/nodejs-bigtable/pull/1529) adds a helper method for changing the request passed into the test proxy readModifyWriteRow service into a request suitable for the handwritten layer. In this PR, that helper method is used by the readModifyWriteRow service to form a request for the handwritten layer and the handwritten layer is called is called by this test proxy service as required.

**Changes:**

**`test/test-proxy/readModifyWriteRowService.ts:`** This test file ensures that the Gapic shaped request passed into the test proxy service is exactly the same as the request that reaches the Gapic layer after being transformed and passed through the handwritten layer.

**`testproxy/services/read-modify-write-row.js:`** The test proxy service is changed to use the handwritten layer.

**Next Steps:**

Now we can address some of the test proxy tests that are failing for this service.


test/test-proxy/readModifyWriteRowService.ts
```diff
@@ -0,0 +1,116 @@
...
+
+import * as assert from 'assert';
+
+import {describe} from 'mocha';
+import {protos} from '../../src';
+import {BigtableClient} from '../../src/v2';
+const readModifyWriteRowService = require('../../../testproxy/services/read-modify-write-row.js');
+const createClient = require('../../../testproxy/services/create-client.js');
+
+describe('TestProxy/ReadModifyWriteRow', () => {
```
daniel-sanche, line 23: Is this just testing the test proxy to make sure it uses the veneer layer instead of gapic? I'm not sure if unit tests for this are really necessary. I would focus on getting the conformance tests passing, and fixing bugs in the proxy that arise as you go

daniel-sanche (approved):
LGTM, but I'm not sure if those test proxy tests are worth keeping, unless I misunderstand the issue

PR #1540, opened 2024-12-09
Fix: TestMutateRow_Generic_Headers

kevkim-codes:
Mutate rows is grabbing the `appProfileId` from the client. 

https://github.com/googleapis/nodejs-bigtable/blob/eb2cd9fb5134e454709b724d99b4da2db9ccadc0/src/tabular-api-surface.ts#L765

We should modify the test proxy test here to do the same.


danieljbruce (approved):
Eventually we'll have to get mutateRow service to use the handwritten layer.

testproxy/services/mutate-row.js
```diff
@@ -23,10 +23,11 @@ const mutateRow = ({clientMap}) =>
   normalizeCallback(async rawRequest => {
     const {request} = rawRequest;
     const {request: mutateRequest} = request;
-    const {appProfileId, mutations, tableName, rowKey} = mutateRequest;
```
danieljbruce, line 26: Eventually we will want the appProfileId to be able to be set per mutateRow call, but the change to use appProfileId to address the test is good for now.

PR #1541, opened 2024-12-10
chore(testproxy): Change the checkAndMutateRow test proxy service so that it uses the handwritten layer

danieljbruce:
**Summary:**

Currently the test proxy calls the checkAndMutateRow method in the Gapic layer, but [it should use the veneer](https://docs.google.com/document/d/1AAaH08Nb33GnYBOfYHbh9TsCNsZuCZqv8rrg2J6cVuU/edit?tab=t.0#heading=h.xzptrog8pyxf). This PR includes the changes required to make the checkAndMutateRow test proxy service use the veneer with a properly formed request.

**Background**

To accompany changes to the checkAndMutateRow test proxy service to make it call the handwritten layer of the client library instead of the Gapic layer, the mutations passed into the service need to be transformed into a format suitable for the handwritten layer. The transformations that change mutations passed into the handwritten layer into the format the Gapic layer needs are very complex and they use an entire [Mutation](https://github.com/googleapis/nodejs-bigtable/blob/main/src/mutation.ts) object to make these transformations. The `checkAndMutateRow` test proxy service must do the inverse of these complex transformations to ensure that data passed into the `checkAndMutateRow` service is in the intended format when it reaches the Gapic layer.

**Changes:**

**Test Changes:**

**`test/mutation.ts:`** This test is Modifying the Mutation object which is affecting all other tests including the new tests. The test should use the sinon sandbox so that this mock is used just for this one test.

**`test/test-proxy/checkAndMutateRowService.ts:`** This adds a test that ensures the data passed into the test proxy is exactly the same as the data that reaches the Gapic layer after the data has been transformed and passed through the handwritten layer.

**`test/test-proxy/mutationParseInverse.ts:`** This file tests to ensure that the mutationParseInverse function is indeed the inverse of the existing `Mutation.parse` function which is a necessary requirement for the checkAndMutateRow test proxy service to transform data it receives so that it is suitable for the handwritten layer.

**`test/test-proxy/readModifyWriteRowService`**: This just cleans up the test a bit so that the promise completes and an async call isn't left hanging.

**Test proxy changes:**

**`testproxy/known_failures.txt`**: Just by making changes to use the handwritten layer a high priority test now passes :)

**`testproxy/services/check-and-mutate-row.js:`** The checkAndMutateRow test proxy service is modified to use the handwritten layer of the client library.

**`testproxy/services/utils/request/createFlatMutationsList.ts:`** This file defines a createFlatMutationsListWithFnInverse function which is used to transform mutations suitable for the Gapic layer into mutations suitable for the handwritten layer.

**`testproxy/services/utils/request/mutateInverse.ts`** This file defines a `mutationParseInverse` function which is the inverse of `Mutation.parse` and which helps `createFlatMutationsListWithFnInverse` transform Gapic layer shaped mutations into handwritten layer shaped mutations.

**Next Steps:**

Address all the tests for the checkAndMutateRow service.

Also polish up the remaining issues with the checkAndMutateRow test proxy service:
https://b.corp.google.com/issues/383526415
https://b.corp.google.com/issues/383526172
https://b.corp.google.com/issues/383525386
https://b.corp.google.com/issues/383562313


kevkim-codes (approved):

test/test-proxy/createFlatMutationsInverse.ts
```diff
@@ -0,0 +1,58 @@
...
+import {
+  createFlatMutationsListWithFn,
+  createFlatMutationsListWithFnInverse as inverse,
+} from '../../testproxy/services/utils/request/createFlatMutationsList';
+import {mutationParseInverse} from '../../testproxy/services/utils/request/mutateInverse';
+import * as assert from 'assert';
+import {google} from '../../protos/protos';
+import IMutation = google.bigtable.v2.IMutation;
+
+describe('Check createFlatMutationsList and createFlatMutationsListInverse are inverses', () => {
```
daniel-sanche, line 25: All of these tests against the test infrastructure seems a little overkill to me. What is the failure case you're trying to protect agaisnt?

test/test-proxy/mutationParseInverse.ts
```diff
@@ -0,0 +1,73 @@
...
+        {
+          setCell: {
+            familyName: 'cf2',
+            timestampMicros: 1000007,
+            columnQualifier: Buffer.from('cq2'),
+            value: Buffer.from('value2'),
+          },
+        },
+        /*
+        TODO: Fix the inverse function later so that these mutations can be used in the test proxy.
```
daniel-sanche, line 40: Is this TODO still relevant?

testproxy/services/check-and-mutate-row.js
```diff
@@ -16,30 +16,89 @@
...
-    const {clientId} = request;
-    const client = clientMap.get(clientId)[v2];
-    const [result] = await client.checkAndMutateRow({
+    const {clientId, request: checkAndMutateRowRequest} = request;
+    const {
+      // authorizedViewName, // TODO: Pass the authorizedViewName along in the test proxy.
       appProfileId,
       falseMutations,
       rowKey,
+      // predicateFilter, // TODO: Pass the predicateFilter along in the test proxy.
```
daniel-sanche, line 74: is the TODO needed?

testproxy/services/check-and-mutate-row.js
```diff
@@ -16,30 +16,89 @@
...
-    return {
-      status: {code: grpc.status.OK, details: []},
-      result,
-    };
+    } = checkAndMutateRowRequest;
+    const onMatch = handwrittenLayerMutations(trueMutations);
+    const onNoMatch = handwrittenLayerMutations(falseMutations);
+    const id = convertFromBytes(rowKey);
+    const bigtable = clientMap.get(clientId);
+    bigtable.appProfileId = appProfileId; // TODO: Remove this line and pass appProfileId into row.filter.
```
daniel-sanche, line 82: is todo needed?

daniel-sanche (changes requested):
I left a small question asking about the tests verifying test infrastructure, but my main concern is all the commented out TODOs

test/test-proxy/mutationParseInverse.ts
danieljbruce: Yes. The inverse function only properly inverts setCell mutations and I decided to punt the work to address other mutations because of the amount of time it will take. So this TODO is still outstanding.

testproxy/services/check-and-mutate-row.js
danieljbruce: Yes. Changing the predicateFilter to be ready for the handwritten layer is a complex problem. The entire [Filter](https://github.com/googleapis/nodejs-bigtable/blob/f6176c110f832dbacc14ac60fa1870e69aa139ce/src/filter.ts#L273) object is dedicated to preparing the `predicateFilter` so inverting this operation is hard. At least a week of work so I made a TODO.

testproxy/services/check-and-mutate-row.js
danieljbruce: Yes. The `filter` method currently does not accept an `appProfileId`, but we should allow the app profile id to be overridden per method call. The evidence of this is that the `appProfileId` is passed into the `checkAndMutateService` and not the `createClient` service so the intent is to set the appProfileId per method call.

test/test-proxy/createFlatMutationsInverse.ts
danieljbruce: I removed `test/test-proxy/createFlatMutationsInverse.ts`. Let me know if you need anything else.

test/test-proxy/mutationParseInverse.ts
danieljbruce: Removed TODO

testproxy/services/check-and-mutate-row.js
danieljbruce: Remove TODO

testproxy/services/check-and-mutate-row.js
danieljbruce: Removed TODO

daniel-sanche (approved):

PR #1542, opened 2024-12-11
chore(testproxy): Address the TestReadModifyWriteRow_Generic_Headers test

danieljbruce:
Without these changes the `TestReadModifyWriteRow_Generic_Headers` test fails. In this test the call to the `createClient` test proxy service supplies the appProfileId and stores it on the client. However, when the `readModifyWriteRow` service is called, an empty `appProfileId` is provided and overwrites the appProfileId that was stored on the client in the `createClient` service call. The `appProfileId` is an empty string because the `appProfileId` is a `string` in the proto and the test runner does not provide an appProfileId for the call to the `readModifyWriteRow` service.

The change involves assuming the test runner does not intend to provide an appProfileId for the `readModifyWriteRow` service when the service receives an empty string for `appProfileId` and therefore doesn't override the `appProfileId` on the client in that situation.

### DIFF
--- a/testproxy/known_failures.txt
+++ b/testproxy/known_failures.txt
@@ -3,7 +3,6 @@ TestMutateRow_Generic_DeadlineExceeded\|
 TestMutateRows_Generic_CloseClient\|
 TestMutateRows_Retry_WithRoutingCookie\|
 TestMutateRows_Generic_DeadlineExceeded\|
-TestReadModifyWriteRow_Generic_Headers\|
 TestReadModifyWriteRow_Generic_DeadlineExceeded\|
 TestReadModifyWriteRow_NoRetry_MultiValues\|
 TestReadModifyWriteRow_Generic_CloseClient\|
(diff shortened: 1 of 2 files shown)


bhshkh:
Do we want to handle the scenario where user intentionally wants to pass in an empty appProfileID ?

bhshkh (approved):

bhshkh:
Is passing in an empty profile id with readmodifywrite request a valid usecase for end customers? I'm not very familiar with how end user makes the readmodifywrite call using the nodejs library.

cindy-peng:
LGTM. Somehow I don't have permission to approve the PR.

danieljbruce:
> Is passing in an empty profile id with readmodifywrite request a valid usecase for end customers? I'm not very familiar with how end user makes the readmodifywrite call using the nodejs library.

It is a valid use case, but we can't distinguish a "" from undefined coming from the grpc service and making changes so that the distinction could be made would require fundamental changes to grpc or changing the proto. Neither of these are options so this is the best we can do.

PR #1543, opened 2024-12-12
chore(testproxy): Address the TestReadModifyWriteRow_Generic_DeadlineExceeded conformance test

danieljbruce:
**Summary**

The conformance test passes a timeout setting into the `createClient` test proxy service. With this change the timeout setting will now be applied to the client and used in the readModifyWriteRow call. This allows the `TestReadModifyWriteRow_Generic_DeadlineExceeded` conformance test to pass because now the timeout is being used for these calls.

### DIFF
--- a/testproxy/services/create-client.js
+++ b/testproxy/services/create-client.js
@@ -69,7 +69,7 @@ const createClient = ({clientMap}) =>
       Object.entries(
         clientConfig.interfaces['google.bigtable.v2.Bigtable'].methods
       ).forEach(([k, v]) => {
-        if (k === 'ReadRows') {
+        if (k === 'ReadRows' || k === 'ReadModifyWriteRow') {
           /*
           TODO: In the future we should apply this for all methods, but right
           now doing so results in regressions in the TestSampleRowKeys_Generic_MultiStreams
--- a/testproxy/services/create-client.js
+++ b/testproxy/services/create-client.js
@@ -70,18 +70,8 @@ const createClient = ({clientMap}) =>
        */
       Object.entries(
         clientConfig.interfaces['google.bigtable.v2.Bigtable'].methods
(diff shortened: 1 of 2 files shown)


testproxy/services/create-client.js
```diff
@@ -31,7 +31,9 @@ const createClient = ({clientMap}) =>
   normalizeCallback(async rawRequest => {
     // TODO: Handle refresh periods
     const {request} = rawRequest;
-    const clientConfig = require('../../src/v2/bigtable_client_config.json');
+    const clientConfig = JSON.parse(
+      JSON.stringify(require('../../src/v2/bigtable_client_config.json'))
+    );
```
danieljbruce, line 36: To prevent conformance tests from modifying behaviour in other conformance tests, we must use a copy of the clientConfig so that we don't modify the original and set timeouts for tests that shouldn't be using timeouts.

testproxy/services/create-client.js
```diff
@@ -68,18 +70,8 @@ const createClient = ({clientMap}) =>
        */
       Object.entries(
         clientConfig.interfaces['google.bigtable.v2.Bigtable'].methods
-      ).forEach(([k, v]) => {
-        if (k === 'ReadRows') {
-          /*
-          TODO: In the future we should apply this for all methods, but right
-          now doing so results in regressions in the TestSampleRowKeys_Generic_MultiStreams
-          and TestSampleRowKeys_Generic_CloseClient conformance tests that need
-          to be addressed.
```
danieljbruce, line 77: The regressions have been addressed by making a copy of the clientConfig so now this timeout can apply to all methods.

kevkim-codes (approved):

cindy-peng (approved):

PR #1554, opened 2024-12-13
chore(testproxy): Address the TestMutateRows_Generic_DeadlineExceeded conformance test

danieljbruce:
**Summary**

With these new changes the TestMutateRows_Generic_DeadlineExceeded test now passes. In the mutate method in `src/tabular-api-surface.ts` we get a timeout value first if it was provided in the method call and if it's not available there then from the timeout value provided at client creation time. Then, we never retry an error if the timeout has been exceeded and instead return the error as a partial failure error without the code along with an error message for each row that hasn't been successfully sent back to the user yet. Finally, some adjustments needed to be made to the test proxy so that the test runner would receive results.

**WARNING**

This is a breaking change because errors that used to come back from the server will not have a code anymore. [The test runner is actually throwing an assertion error if there is an error code](https://github.com/googleapis/cloud-bigtable-clients-test/blob/c78ca1867bbf862ce7d742a999b7d45ae5edc478/tests/mutaterows_test.go#L226). Instead of this change consider removing this check from the test. Also consider whether there [should be an error for each row mutation that is still pending](https://github.com/googleapis/cloud-bigtable-clients-test/blob/c78ca1867bbf862ce7d742a999b7d45ae5edc478/tests/mutaterows_test.go#L227).

UPDATE: We can ignore this warning because [the test](https://github.com/googleapis/cloud-bigtable-clients-test/pull/224) has been changed and so the code is still there.

**Alternatives:**

We can remove some of the requirements in the test runner in order to avoid breaking changes.


testproxy/services/bulk-mutate-rows.js
```diff
@@ -39,10 +39,17 @@ const bulkMutateRows = ({clientMap}) =>
...
+            details: [],
+            message: error.message,
+          },
           entries: Array.from(error.errors.entries()).map(([index, entry]) => ({
             index: index + 1,
-            status: entry,
+            status: {
+              code: entry.code,
+              message: entry.message,
+            },
```
danieljbruce, line 52: These changes are necessary because if we send information that is out of scope back to the test runner then the test runner just gets a Nil response.

src/tabular-api-surface.ts
```diff
@@ -736,12 +759,30 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
-
+      if (err && timeoutExceeded) {
+        callback(
+          new PartialFailureError(
+            /* We concatenate an error onto the list for each pending entry
+              because it is required in TestMutateRows_Generic_DeadlineExceeded.
+              The test checks to see if the number of entries is greater than 0
+              indicating that it is expecting some errors even though no data
+              has been received yet.
+               */
```
danieljbruce, line 775: It is expecting at least one error reported [here](https://github.com/googleapis/cloud-bigtable-clients-test/blob/c78ca1867bbf862ce7d742a999b7d45ae5edc478/tests/mutaterows_test.go#L227).

src/tabular-api-surface.ts
```diff
@@ -736,12 +759,30 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
       }
 
+      const mutationErrors = Array.from(mutationErrorsByEntryIndex.values());
       if (mutationErrorsByEntryIndex.size !== 0) {
-        const mutationErrors = Array.from(mutationErrorsByEntryIndex.values());
         callback(new PartialFailureError(mutationErrors, err));
         return;
       }
-
+      if (err && timeoutExceeded) {
```
danieljbruce, line 767: Do we want to omit the error code is some cases where the timeout hasn't been exceeded as well?

src/tabular-api-surface.ts
```diff
@@ -736,12 +759,30 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+        callback(
+          new PartialFailureError(
```
danieljbruce, line 769: The test is expecting no error code so it seems like it is expecting a PartialFailureError because it is also expecting errors per row so that's why we use this object.

src/tabular-api-surface.ts
mutianf: There are RPC level error code and entry level error code. When the deadline is passed, rpc level error code should be deadline exceeded and entry level error code should be whichever code it gets from the server.

src/tabular-api-surface.ts
mutianf: The test expect deadline exceeded ? https://github.com/googleapis/cloud-bigtable-clients-test/blob/c78ca1867bbf862ce7d742a999b7d45ae5edc478/tests/mutaterows_test.go#L229

src/tabular-api-surface.ts
mutianf: Yes, partialFalureError should contain a list of failed entries. For example if we sent a requet [row1, row2, row3] and they all failed with unavailable errors, the failure will look something like: DeadlineExceeded(unavailable, unavailable, unavailable)

danieljbruce: Thanks for the example.

The behaviour of the test is that it makes a request and then [no errors for entries are sent back while one entry is still pending](https://github.com/googleapis/cloud-bigtable-clients-test/blob/c78ca1867bbf862ce7d742a999b7d45ae5edc478/tests/mutaterows_test.go#L197). Then the client receives a `DEADLINE_EXCEEDED` error for the test overall. So without any code changes we get DeadlineExceeded(... no responses yet).

My change takes the pending responses and turns them into DEADLINE_EXCEEDED errors at the entry level because this is what the test is expecting. I want to double check this with you though because it seems like an unnatural fix to get the test to pass.

src/tabular-api-surface.ts
danieljbruce: The line you highlighted is the entry level error (it is expecting error codes at the entry level). At the RPC level it is [expecting ok status](https://github.com/googleapis/cloud-bigtable-clients-test/blob/c78ca1867bbf862ce7d742a999b7d45ae5edc478/tests/mutaterows_test.go#L226) with [no code](https://github.com/googleapis/cloud-bigtable-clients-test/blob/c72bf49142e8208b586d51b240f46860ae5d4b28/tests/test_workflow.go#L599).

src/tabular-api-surface.ts
danieljbruce: Note that actually [the test is expecting no error code at the grpc level](https://github.com/googleapis/cloud-bigtable-clients-test/blob/c72bf49142e8208b586d51b240f46860ae5d4b28/tests/mutaterows_test.go#L226). Does the test need to change?

src/tabular-api-surface.ts
danieljbruce: The test is being changed to expect DEADLINE_EXCEEDED errors now and the Java test proxy is being changed to ensure the client error gets passed back to the test runner.

https://github.com/googleapis/cloud-bigtable-clients-test/pull/224
https://github.com/googleapis/java-bigtable/pull/2456

In the next update we will have the client apply a DEADLINE_EXCEEDED grpc error so that this new check passes instead of the old check.

src/tabular-api-surface.ts
mutianf, line 767: instead of checking timeoutExceeded here, I think we can push err to each entry if the entry doesn't have a status code.

src/tabular-api-surface.ts
```diff
@@ -736,12 +759,30 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+            /* We concatenate an error onto the list for each pending entry
+              because it is required in TestMutateRows_Generic_DeadlineExceeded.
```
mutianf, line 771: Let's update this comment to describe what the expected client behavior is instead.  Something like:
"If there's an RPC level failure and the mutation entries don't have a status code, the RPC level failure error code will be used as the entry failure code."

mutianf (changes requested):

src/tabular-api-surface.ts
danieljbruce: With the new changes to the test expecting DEADLINE_EXCEEDED I think this makes sense.

src/tabular-api-surface.ts
danieljbruce: Yup. Updated the comment.

testproxy/services/bulk-mutate-rows.js
```diff
@@ -39,16 +39,36 @@ const bulkMutateRows = ({clientMap}) =>
     } catch (error) {
       if (error.name === 'PartialFailureError') {
         return {
-          status: error,
+          status: {
```
mutianf, line 42: does this still need to be a special case? It looks the same as the other statement?

src/tabular-api-surface.ts
```diff
@@ -736,12 +759,25 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
       }
 
+      const mutationErrors = Array.from(mutationErrorsByEntryIndex.values());
       if (mutationErrorsByEntryIndex.size !== 0) {
-        const mutationErrors = Array.from(mutationErrorsByEntryIndex.values());
         callback(new PartialFailureError(mutationErrors, err));
         return;
       }
-
+      if (err) {
```
mutianf, line 767: Can we also add a unit test for this?

danieljbruce: Added.

Actually, something else I should flag is the code in the `mutationErrorsByEntryIndex.size !== 0` check. It seems odd to me that as soon as we see one entry level error that we don't send the grpc level error code anymore and we don't send an error for each pending entry.

Should we make changes so that the `mutationErrorsByEntryIndex.size !== 0` and the `mutationErrorsByEntryIndex.size === 0` cases send back more similar errors?

mutianf: I think the logic right now is fine. We want to keep the entry level error code as is.

mutianf (approved):

PR #1555, opened 2024-12-19
chore(testproxy): Address the TestReadModifyWriteRow_Generic_CloseClient conformance test

danieljbruce:
**Summary**

The `TestReadModifyWriteRow_Generic_CloseClient` conformance test is failing because results are not being passed along by the test proxy and the closed client error doesn't have an error code. These two problems are addressed in this PR so that the test passes.

Changes are discussed inline.

### DIFF
--- a/src/row-data-utils.ts
+++ b/src/row-data-utils.ts
@@ -29,7 +29,7 @@ import {google} from '../protos/protos';
 import {TabularApiSurface} from './tabular-api-surface';
 import arrify = require('arrify');
 import {Bigtable} from './index';
-import {CallOptions, ServiceError} from 'google-gax';
+import {CallOptions} from 'google-gax';
 
 interface TabularApiSurfaceRequest {
   tableName?: string;
--- a/src/row-data-utils.ts
+++ b/src/row-data-utils.ts
@@ -29,7 +29,7 @@ import {google} from '../protos/protos';
 import {TabularApiSurface} from './tabular-api-surface';
 import arrify = require('arrify');
 import {Bigtable} from './index';
-import {CallOptions} from 'google-gax';
+import {CallOptions, ServiceError} from 'google-gax';
 
 interface TabularApiSurfaceRequest {
   tableName?: string;
@@ -193,6 +193,15 @@ class RowDataUtils {
       properties.reqOpts
     );
     properties.requestData.data = {};
+    const passThroughCallback = (
+      err: ServiceError | null,
+      apiResponse?: google.bigtable.v2.IReadModifyWriteRowResponse
+    ) => {
+      if (err && err.message === 'The client has already been closed.') {
+        err.code = 1; // CANCELLED
+      }
+      callback(err, apiResponse);
+    };
     properties.requestData.bigtable.request<google.bigtable.v2.IReadModifyWriteRowResponse>(
       {
         client: 'BigtableClient',
@@ -200,7 +209,7 @@ class RowDataUtils {
         reqOpts,
         gaxOpts: gaxOptions,
       },
-      callback
+      passThroughCallback
     );
   }
 
--- a/src/row-data-utils.ts
+++ b/src/row-data-utils.ts
@@ -193,15 +193,6 @@ class RowDataUtils {
       properties.reqOpts
     );
     properties.requestData.data = {};
-    const passThroughCallback = (
-      err: ServiceError | null,
-      apiResponse?: google.bigtable.v2.IReadModifyWriteRowResponse
-    ) => {
-      if (err && err.message === 'The client has already been closed.') {
-        err.code = 1; // CANCELLED
-      }
-      callback(err, apiResponse);
-    };
     properties.requestData.bigtable.request<google.bigtable.v2.IReadModifyWriteRowResponse>(
       {
         client: 'BigtableClient',
@@ -209,7 +200,7 @@ class RowDataUtils {
         reqOpts,
         gaxOpts: gaxOptions,
       },
-      passThroughCallback
+      callback
     );
   }
 
--- a/testproxy/known_failures.txt
+++ b/testproxy/known_failures.txt
@@ -1,4 +1,5 @@
 TestMutateRow_Generic_DeadlineExceeded\|
+TestMutateRow_Generic_CloseClient\|
 TestMutateRows_Generic_CloseClient\|
 TestMutateRows_Retry_WithRoutingCookie\|
 TestMutateRows_Generic_DeadlineExceeded\|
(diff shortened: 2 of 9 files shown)


src/row-data-utils.ts
```diff
@@ -193,14 +193,23 @@ class RowDataUtils {
       properties.reqOpts
     );
     properties.requestData.data = {};
+    const passThroughCallback = (
+      err: ServiceError | null,
+      apiResponse?: google.bigtable.v2.IReadModifyWriteRowResponse
+    ) => {
+      if (err && err.message === 'The client has already been closed.') {
+        err.code = 1; // CANCELLED
```
danieljbruce, line 201: Attaching a code here is required for the test to pass, but maybe we should just modify the test runner so that it checks for a message instead of an error code and leave this method unchanged.

testproxy/services/read-modify-write-row.js
```diff
@@ -34,7 +34,7 @@ const readModifyWriteRow = ({clientMap}) =>
     const table = getTableInfo(bigtable, tableName);
     const row = table.row(handWrittenRequest.id);
     try {
-      const result = await row.createRules(handWrittenRequest.rules);
+      const [result] = await row.createRules(handWrittenRequest.rules);
```
danieljbruce, line 37: The return value here is actually an array where the first element is the results so this change is necessary.

testproxy/known_failures.txt
```diff
@@ -1,9 +1,9 @@
 TestMutateRow_Generic_DeadlineExceeded\|
+TestMutateRow_Generic_CloseClient\|
```
danieljbruce, line 2: mutateRow service isn't using the veneer so this test passing before was a false positive.

src/row-data-utils.ts
daniel-sanche: I don't like the idea of reading an error string to change the exception type, it seems hacky. Can we make it raise the proper error originally, instead of overwriting it?

danieljbruce: I don't like the magic string matching either. The problem is that the error [gets raised in the Gapic layer](https://github.com/googleapis/nodejs-bigtable/blob/86f99f59ac530f2ead89ce817f5e20b47d28728d/src/v2/bigtable_client.ts#L319) so we can't make the change where it gets raised originally without regenerating gapic.

Another alternative is to add another check to the if condition `&& !err.code` to only insert a code if the code is missing. What do you think?

daniel-sanche: Something seems a bit off here. My understanding of [part 4b](https://github.com/googleapis/cloud-bigtable-clients-test/blob/595412bc54322d9e21b7df5edf33e732561f5786/tests/readmodifywriterow_test.go#L333) is it's meant to detect rpcs cancelled in-flight, so it should be the backend raising the error here, not the gax layer. The [code you posted](https://github.com/googleapis/nodejs-bigtable/blob/86f99f59ac530f2ead89ce817f5e20b47d28728d/src/v2/bigtable_client.ts#L319) seems to be for rejecting rpcs started after the client is already closed, which would be verified by [part 4c of the test](https://github.com/googleapis/cloud-bigtable-clients-test/blob/595412bc54322d9e21b7df5edf33e732561f5786/tests/readmodifywriterow_test.go#L349)

This is a bit of a complicated situation though, so let me know if I'm missing something. I think I'll need to dive a bit deeper to really give a recommendation here

daniel-sanche: I'm actually a bit unsure why `Cancelled` is even acceptable for this test? Looking at the test, I'd expect all the [batch 1](https://github.com/googleapis/cloud-bigtable-clients-test/blob/595412bc54322d9e21b7df5edf33e732561f5786/tests/readmodifywriterow_test.go#L337) rpcs to be successful, since the client isn't closed until they are completed...

danieljbruce: all the batch 1 rpcs _are_ successful. The assertion statement that this change addresses is [here](https://github.com/googleapis/cloud-bigtable-clients-test/blob/595412bc54322d9e21b7df5edf33e732561f5786/tests/readmodifywriterow_test.go#L353). Looking into this further...

danieljbruce: [The assertion statement throwing the error](https://github.com/googleapis/cloud-bigtable-clients-test/blob/595412bc54322d9e21b7df5edf33e732561f5786/tests/readmodifywriterow_test.go#L353) throws an error if the code is 0 or if it is not provided. I am okay with using a different error code, but we must use a non-zero error code to address this assert statement. If you want to see why the code adjustment here is necessary, try running the conformance test and commenting out this code change.

What do you think we should do?

danieljbruce: The test proxy service returns a status to the runner.

It might make it easier to see with a screenshot of what happens with these changes:

image

If you set a breakpoint just before the test proxy sends back the status, you can see that the error code that will be sent back is 1. Without these changes, `e` has a `stack` and a `message`, but not a code. So it looks like the close client error from gapic is not providing an error code.

daniel-sanche: Yeah, my confusion was over why a method that returns an exception was being passed as successful, but I think I understand now. The gapic layer is using `this._gaxModule.GoogleError` withot a grpc code, since the error is client side, and not from a grpc channel. And the test proxy is interpreting no grpc code as `0`, or `OK`

So to address this, I think you should change line 45 in your screenshot to something like `code: e.code ?? grpc.status.UNKNOWN`. Currently, the test proxy is reporting that exceptions without a grpc status code are OK, but that's not what the client is actually telling the proxy. The client had an error, but the proxy is covering it up. We need to fix that

daniel-sanche (changes requested):
please fix the test proxy to add a failed `UNKNOWN` status code for client-side exceptions, instead of passing `OK`

src/row-data-utils.ts
danieljbruce: Ok. We need to make that particular change for all the services then.

src/row-data-utils.ts
```diff
@@ -29,7 +29,7 @@ import {google} from '../protos/protos';
 import {TabularApiSurface} from './tabular-api-surface';
 import arrify = require('arrify');
 import {Bigtable} from './index';
-import {CallOptions} from 'google-gax';
+import {CallOptions, ServiceError} from 'google-gax';
```
daniel-sanche, line 32: looks like this isn't needed anymore

src/row-data-utils.ts
daniel-sanche: Ok, sounds good

daniel-sanche:
> We have to be careful because maybe the assertion actually was intending to check that the error the client library sent back contained an error code. Or maybe it's just checking to see that an error was sent back by the client and is looking for an error code to validate that. It's hard to know what the intent of the test was for sure.

Yeah I guess It's a bit confusing that the client libraries can raise client-side exceptions that aren't associated with grpc errors, while the conformance test runner can only receive grpc errors back. But it's the test proxy's job to translate between those two error formats. To convert nodejs-native exceptions into language-agnostic gprc errors

Sometimes we might have to read the tests to make sure we are understanding the intentions properly, but the test proxy reporting client-side errors as successful is a pretty clear bug in the proxy

Typically, if a test is expecting to see a specific error code from the server, that code would be part of the assertion

daniel-sanche (approved):

PR #1562, opened 2024-12-23
fix: Conformance test sample rowkeys generic deadline

kevkim-codes:
Fixes generic deadline issue for sample rowkeys method. We were not handling for deadline exceeded errors in the original method as we were expecting GAX to handle it. Hopefully, down the road we will have GAX handle these at a parent level and send back the proper error code.


cindy-peng (approved):

testproxy/services/utils/request/sampleRowKeys.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const getTableInfo = (bigtable: any, tableName: any) => {
+    const [, , , instanceId, , tableId] = tableName.split('/');
+    const instance = bigtable.instance(instanceId);
+    return instance.table(tableId);
+  };
+
+  const table = getTableInfo(client, tableName);
+  const sampleRowKeysResponse = table.sampleRowKeys();
```
danieljbruce, line 45: I think this may actually need `await` before `table.sampleRowKeys()`.

danieljbruce (approved):
Looks good for meeting the needs of the tests. We may want to add retry behaviour to sampleRowKeys at some point.

PR #1563, opened 2025-01-08
fix: Use the universe domain if it is provided by the user

danieljbruce:
**Summary:**

The user should be able to set the universe domain so that outgoing calls work with a different Google Cloud Universe. Right now, if the user specifies a universe domain then it will not get used and the request will just be sent to the default Bigtable endpoint.

**Changes:**

`src/index.ts:` The change here is that when a custom url is not provided, but a universe domain is provided then the servicePath will use the universe domain provided. This is done for each Gapic client and ensures requests will be made to the universe domain instead of the default Bigtable service.

`system-test/service-path.ts`: Two tests are added to ensure the service path is always set correctly. For instance, when a custom endpoint is provided then the service path will still always be that custom endpoint. When a custom endpoint is NOT provided, but a universe domain is provided then the service path is set to use the universe domain instead of the default endpoint.

**Alternatives:**

1. Instead of requiring the user to set the universe domain for each Gapic client, we could expose a `universeDomain` option that would apply to all clients. While this would make things easier for users, it is an API change so it is not reversible. For now we should not change the API since users already have a way to specify the universe domains for the Gapic clients. We can always add this option later.
2. When a custom endpoint is not provided then in the handwritten layer we could not provide a service path. The Pros of this option is that we can rely on the code in the Gapic layer for building the service path instead of creating the getDomain function introducing duplicate code and this would reduce technical debt. The main problem with this option is that it means [providedCustomServicePath](https://github.com/googleapis/nodejs-bigtable/blob/2fdf98f9398761046f41bd58e00258f0d4963a3d/src/v2/bigtable_client.ts#L132) will now be undefined and this value is used in the grpc stub so we don't know how such a change might affect the outgoing API calls. The code to implement this alternative also gets messy if Alternative 1 isn't adopted as well.


src/index.ts
```diff
@@ -413,9 +415,6 @@ export class Bigtable {
       }
     }
 
-    const defaultBaseUrl = 'bigtable.googleapis.com';
-    const defaultAdminBaseUrl = 'bigtableadmin.googleapis.com';
```
danieljbruce, line 417: These variables aren't used anymore. These constants also exist in the Gapic layer where the service path is built downstream.

src/index.ts
```diff
@@ -464,14 +486,26 @@ export class Bigtable {
       {},
       baseOptions,
       {
-        servicePath: customEndpointBaseUrl || defaultAdminBaseUrl,
+        servicePath:
+          customEndpointBaseUrl ||
+          `bigtableadmin.${getDomain(options.BigtableTableAdminClient)}`,
```
danieljbruce, line 491: If no universe domain is provided then this will still be bigtable.googleapis.com.

src/index.ts
```diff
@@ -445,7 +465,9 @@ export class Bigtable {
       {},
       baseOptions,
       {
-        servicePath: customEndpointBaseUrl || defaultBaseUrl,
+        servicePath:
+          customEndpointBaseUrl ||
+          `bigtable.${getDomain(options.BigtableClient)}`,
```
danieljbruce, line 470: If no universe domain and custom endpoint is provided then this will still be bigtable.googleapis.com.

src/index.ts
```diff
@@ -102,6 +102,29 @@ export interface BigtableOptions extends gax.GoogleAuthOptions {
...
+/**
+ * Retrieves the domain to be used for the service path.
+ *
+ * This function retrieves the domain from gax.ClientOptions passed in or via an environment variable.
+ * It defaults to 'googleapis.com' if none has been set.
+ * @param {gax.ClientOptions} [opts] The gax client options
+ * @returns {string} The universe domain.
+ */
+function getDomain(opts?: gax.ClientOptions) {
+  // This code for universe domain was taken from the Gapic Layer.
```
daniel-sanche, line 114: can you add a permalink to where this came from?

src/index.ts
daniel-sanche, line 470: What would you think about adding `pathPrefix` as an argument to getDomain (and maybe renaming the method if needed), instead of manually building the string here? I think that would be a bit easier to read, and seems more natural to me since the domain string isn't useful on its own

daniel-sanche (approved):
Mostly LGTM, with a few small comments. 

I'm not too familiar with the universe domains yet though, so you might also want to assign a reviewer who understands potential complications there

src/index.ts
```diff
@@ -102,6 +102,29 @@ export interface BigtableOptions extends gax.GoogleAuthOptions {
...
+  // It is reused here to build the service path.
+  const universeDomainEnvVar =
+    typeof process === 'object' && typeof process.env === 'object'
+      ? process.env['GOOGLE_CLOUD_UNIVERSE_DOMAIN']
+      : undefined;
+  return (
+    opts?.universeDomain ??
+    opts?.universe_domain ??
```
bshaffer, line 122: Is this a nodeJS convention to support both casings? AFAIK this is not the case in any other language implementation

testproxy/known_failures.txt
```diff
@@ -12,4 +12,7 @@ TestReadRows_Retry_WithRoutingCookie_MultipleErrorResponses\|
 TestReadRows_Retry_WithRetryInfo\|
 TestReadRows_Retry_WithRetryInfo_MultipleErrorResponse\|
 TestSampleRowKeys_Retry_WithRoutingCookie\|
-TestSampleRowKeys_Generic_CloseClient
+TestSampleRowKeys_Generic_CloseClient\|
+TestSampleRowKeys_Generic_Headers\|
+TestSampleRowKeys_NoRetry_NoEmptyKey\|
+TestSampleRowKeys_Retry_WithRetryInfo
```
bshaffer, line 18: Why are these part of this PR? These tests shouldn't be affected as a result of these changes

system-test/service-path.ts
```diff
@@ -0,0 +1,128 @@
...
+// limitations under the License.
+
+import {describe, it} from 'mocha';
+import * as assert from 'assert';
+import {ServiceError} from 'google-gax';
+import {Bigtable} from '../src';
+import {BigtableClient, BigtableInstanceAdminClient} from '../src/v2';
+
+describe('Service Path', () => {
+  it('Setting universe domain should set the service path', async () => {
```
bshaffer, line 22: I would suggest also

 1. adding a test for when the universe domain ENV VAR is set.
 2. setting the ENV VAR in the other two tests as well, which will ensure that it's not used when the `universeDomain` and `apiEndpoint` options are set.

bshaffer (commented):
This looks good - a few test suggestions and questions, but otherwise LGTM

testproxy/known_failures.txt
danieljbruce: These needed to be added because suddenly conformance tests started failing for sampleRowKeys. This is most likely because new tests were recently added to the test runner.

src/index.ts
danieljbruce: This is code is taken from [here](https://github.com/googleapis/nodejs-bigtable/blob/589540475b0b2a055018a1cb6e475800fdd46a37/src/v2/bigtable_client.ts#L120-L129). I suggest we leave it as is because this is just Gapic code, but if you have an exact suggestion then maybe I could try that as well?

system-test/service-path.ts
danieljbruce: ok. I'll do that next.

src/index.ts
bshaffer: In the JSON file, it's defined as `universe_domain`. I don't have a exact suggestion other than to say since the language agnostic spec doesn't require both casings, it must be specific to NodeJS.

According to https://github.com/googleapis/nodejs-bigtable/pull/1386, that's how support was added for all GAPICs, so LGTM

testproxy/known_failures.txt
kevkim-codes: I'd pull these out of the PR.

src/index.ts
danieljbruce: Done

src/index.ts
danieljbruce: Yup. That would be fine.

system-test/service-path.ts
danieljbruce: Done

src/index.ts
danieljbruce: Ok. Thanks for looking into that.

testproxy/known_failures.txt
danieljbruce: These have been pulled out.

bshaffer (approved):

PR #1564, opened 2025-01-09
chore: Common pitfalls unit test

kevkim-codes:
A problem occurs when all of the `ReadRowsResponses` have been received, but the deadline is exceeded before receiving an OK status in the trailers. The problem manifests as 2 symptoms:

- A naive implementation will schedule a retry attempt (since DEADLINE_EXCEEDED is considered to be a transient error). However, when truncating the resumption request, it will remove all of the elements from the RowSet (because all of the keys and/or ranges have been received). A ReadRowsRequest that is an empty RowSet implies a full table scan.
- If the original request had a row limit set in the request, that count will be decremented to 0. Which implies no limit. For example if the end user wants to find the first row of the table, they would craft a request with limit: 1 and an empty rowset. If a deadline is exceeded after receiving the row but before an OK, then rowLimit will be 0, which means unlimited, causing a full table scan.

We have now added a unit test that confirms this behavior.


test/readrows.ts
```diff
@@ -461,6 +462,47 @@ describe('Bigtable/ReadRows', () => {
     })();
   });
 
+  it('requests a full table scan during a retry on a transient error', () => {
```
danieljbruce, line 465: I'm not sure I understand the whole situation here as this test is passing. Is the intent of adding this test to confirm that the client library doesn't contain the common pitfall that the PR describes? Or is this a unit test that confirms the common pitfall is present in the client library right now?

danieljbruce (commented):
A couple of questions. Just trying to understand the whole situation at the moment.

test/readrows.ts
kevkim-codes: It could be either, hypothetically, but the common pitfalls are _likely_ to be present. The point of the unit test is to confirm by code that the common pitfall **is** present in the library.

test/readrows.ts
```diff
@@ -461,6 +462,47 @@ describe('Bigtable/ReadRows', () => {
...
+
+      const readStream = table.createReadStream();
+
+      // Stream push simulate receiving rows.
+      readStream.push({
+        id: '0',
+      });
+
+      // Simulate deadline exceeded.
+      readStream.destroy(new GoogleError('DEADLINE_EXCEEDED'));
```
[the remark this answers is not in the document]
kevkim-codes: Doesn't this simulate the same behavior?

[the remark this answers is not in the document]
danieljbruce: If you do this then the retry layer won't receive the DEADLINE_EXCEEDED error and the full table scan won't be attempted. I think you'll have to set up some kind of mock service that sends an error back like [this](https://github.com/googleapis/nodejs-bigtable/blob/2fdf98f9398761046f41bd58e00258f0d4963a3d/test/utils/readRowsImpl.ts#L429) for example.

test/readrows.ts
danieljbruce: Wouldn't that mean the test passes when the library is not working correctly and fails when the library is working correctly?

If so, I would recommend writing the test so that it fails when the pitfall is present and then skip it for now with .skip.

kevkim-codes: Got it, updated the code. Lmk what you think.

test/readrows.ts
```diff
@@ -461,6 +462,45 @@ describe('Bigtable/ReadRows', () => {
...
+    try {
+      readRowsWithDeadline();
+      assert.fail('Should have thrown error');
+    } catch (err) {
+      if (err instanceof GoogleError) {
+        assert.equal(err.code, 'DEADLINE_EXCEEDED');
+      }
+
+      // Assert that the retry was attempted with an empty RowSet and limit 0
+      assert.strictEqual(retrySpy.callCount, 0);
```
danieljbruce, line 497: The assert fails here although `retrySpy` is counting the number of calls to `createReadStream` so we would expect this to be 1 because of `table.createReadStream();` above so this assertion statement doesn't actually check for something we want. After changing `retrySpy` to `requestSpy`, you could check `assert.strictEqual(requestSpy.callCount, 1);` to make sure exactly one request is made and therefore no retries are attempted.

test/readrows.ts
```diff
@@ -461,6 +462,45 @@ describe('Bigtable/ReadRows', () => {
...
+      service.setService({
+        ReadRows: ReadRowsImpl.createService(
+          TRANSIENT_ERROR_SERVICE
+        ) as ServerImplementationInterface,
+      });
+
+      table.createReadStream();
+    }
+    // Spy on the retry mechanism.
+    const retrySpy = sinon.spy(table, 'createReadStream');
```
danieljbruce, line 486: I would replace this with `const requestSpy = sinon.spy(table.bigtable, 'request_')` since this spy is currently useless. We would always expect `createReadStream` to be called exactly once.

test/utils/readRowsImpl.ts
```diff
@@ -431,7 +437,17 @@ export class ReadRowsImpl {
...
         }
+
+        if (deadlineExceededError) {
+          debugLog('sending deadline exceeded error');
+          const error = new GoogleError('Deadline exceeded');
+          error.code = Status.DEADLINE_EXCEEDED;
+          stream.emit('error', error);
+          readRowsRequestHandler.cancelled = true;
+          break;
+        }
```
danieljbruce, line 448: This looks like it will simulate the right behaviour. You could put this in the `chunkIdx === this.errorAfterChunkNo` block above and specify `errorAfterChunkNo` when defining the service and that might be less code, but this will work.

danieljbruce (commented):
It's definitely on the right track. We should make sure the test is measuring the right thing and then I think it is good to go.

danieljbruce (approved):
Looks good!

PR #1566, opened 2025-01-20
feat: Add the MetricsCollector for client side metrics

danieljbruce:
### **Summary:**

This PR adds the `MetricsCollector` which the rest of the client library will use to collect client side metrics to send to GCP backend for analysis. None of the methods will use `MetricsCollector` yet so this PR doesn't change client library behaviour in any way. It is just in the stage of building infrastructure to support client side metrics.

### **Background:**

Below is a diagram of the new data structures that will be added.

CSM diagram 6 drawio

### **Changes:**

**Files that will contain utilities for unit tests and system tests:**

`common/client-side-metrics-attributes.ts:` Contains interfaces for various client side metrics attributes.

`common/logger.ts:` Provides abstract classes for other test classes so that those test classes can easily log test results.

`common/test-date-provider:`: Exports a DateProvider class used in metrics collector tests that provides a predictable series of dates so that the metrics collector class used in the tests behaves consistently.

`common/test-metrics-handler.ts:` Exports a metrics handler for testing that simply collects results the metrics collector records instead of sending the recorded results to the GCP backend.

**Files directly pertaining to the metrics collector:**

`src/client-side-metrics/metrics-collector.ts:` Contains the new MetricsCollector class which will be created to collect metrics each time a method is called.

`src/client-side-metrics/metrics-handler.ts:` The metrics handler interface that the metrics collector can use. Users can also supply their own metrics handler for the metrics collector to use.

`test/metrics-collector/metrics-collector.ts`: A test file for the metrics collector that introduces a fake table with a fake method that records metrics using the metrics collector and ensures the right metrics from the method are getting recorded.

`test/metrics-collector/typical-method-call.ts`: The test in `test/metrics-collector/metrics-collector.ts` compares this file with the logger output of the metrics collector and ensures that these two series of logs are exactly the same.

`test/metrics-collector/metrics-handler-fixture.ts`: A file containing a constant that is compared against the data the metrics handler collects in the test.

**Other changes:**

`package.json:` Add new open telemetry packages

### **Suggestions for how to review this PR:**

It may be a good idea to review `test/metrics-tracer/typical-method-call.txt` first because it contains a plain-text log of what the metrics tracer is actually recording for a method call. Looking at this file may reveal details regarding when the client is collecting certain metrics without having to understand the client metrics collector architecture itself.

### **Next Steps:**

0. The GCPMetricHandler was removed from this PR. Introduce it in another PR, pass the `metrics-handler-fixture.ts` constant into it and ensure it writes to the metrics dashboard with a system test. 
1. Write some system tests using this new test infrastructure that logs metrics collected and ensure the metrics collected are what we would expect.
2. Add client side metrics to readRows calls
3. Add client side metrics to other calls
4. Add client_blocking_latencies, application_blocking_latencies, the only metrics not covered yet


src/client-side-metrics/metrics-tracer-factory.ts
```diff
@@ -0,0 +1,600 @@
...
+   * Called when the operation starts. Records the start time.
+   */
+  onOperationStart() {
+    this.operationStartTime = this.dateProvider.getDate();
+  }
+
+  /**
+   * Called after the client reads a row. Records application blocking latencies.
+   */
+  onRead() {
```
daniel-sanche, line 288: Did you do any benchmarking on this one? ReadRows is very performance sensitive. At least in Python, we have to be very thoughtful about literally every line

src/client-side-metrics/metrics-tracer-factory.ts
```diff
@@ -0,0 +1,600 @@
...
+    const currentTime = this.dateProvider.getDate();
+    const projectId = this.projectId;
+    if (this.lastReadTime) {
```
daniel-sanche, line 291: I assume this is doing different logic for the first row vs subsequent rows

It might be worth doing something a bit more complex here, to make sure assumptions hold (i.e. a state machine or something similar)

The state can get complicated, and we want to be careful that we don't accidentally send wrong data

src/client-side-metrics/metrics-tracer-factory.ts
```diff
@@ -0,0 +1,600 @@
...
+  };
+}
+
+const packageJSON = fs.readFileSync('package.json');
+const version = JSON.parse(packageJSON.toString()).version;
+
+/**
+ * A class for tracing and recording client-side metrics related to Bigtable operations.
+ */
+class MetricsTracer {
```
daniel-sanche, line 133: It looks like this class is tightly coupled to OpenTelemetry metrics. I really think you should add a layer of abstraction, so we can change things out easier in the future

If you do keep it coupled, maybe rename this to OpenTelemetryMetricsTracer, so it's more clear what this class is doing

src/client-side-metrics/metrics-tracer-factory.ts
```diff
@@ -0,0 +1,600 @@
...
+
+  /**
+   * Called when an operation completes (successfully or unsuccessfully).
+   * Records operation latencies, retry counts, and connectivity error counts.
+   * @param {OperationInfo} info Information about the completed operation.
+   */
+  onOperationComplete(info: OperationInfo) {
+    const endTime = this.dateProvider.getDate();
+    const projectId = this.projectId;
+    this.onAttemptComplete(info);
```
daniel-sanche, line 359: another comment about adding some kind of state machine. It looks like there are no guards against calling onAttemptComplete or onOperationComplete multiple times, which could pollute the data

src/client-side-metrics/metrics-tracer-factory.ts
```diff
@@ -0,0 +1,600 @@
...
+      this.cluster = instanceInformation[1];
+    }
+  }
+}
+
+/**
+ * A factory class for creating MetricsTracer instances. Initializes
+ * OpenTelemetry metrics instruments.
+ */
+export class MetricsTracerFactory {
```
daniel-sanche, line 462: Same feedback here: please consider adding a layer of abstraction. Or rename this to OtelMetricsTracerFactory, to make the coupling explicit

daniel-sanche (changes requested):

src/client-side-metrics/metrics-handler.ts
```diff
@@ -0,0 +1,68 @@
...
+  firstResponseLatency: number | null;
+  connectivityErrorCount: number | null;
+}
+
+// TODO: Trim attributes so only necessary attributes are required.
+/**
+ * An interface for handling client-side metrics related to Bigtable operations.
+ * Implementations of this interface can define how metrics are recorded and processed.
+ */
+export interface IMetricsHandler {
```
danieljbruce, line 49: I thought about the design further and perhaps it is worthwhile to abstract the interface the user provides even further. ie. Instead of onAttemptComplete, the MetricsHandler has an optional method for each metric like on onRecordAttemptLatency. The proposed design is ok, but the proposed solution tightly couples us to the idea that all metrics provided for onAttemptComplete will have the same dimensions. Having a method per metric is more flexible.

src/client-side-metrics/metrics-tracer-factory.ts
danieljbruce: I have now decided to leave this one out for now and include it in a separate PR. The logistics of making this work for the application blocking latencies are rather complicated so I would prefer to prioritize it last.

src/client-side-metrics/metrics-tracer-factory.ts
danieljbruce: Upon further consideration I have decided to eliminate this onAttemptComplete call. This will require one more line of code from the method collecting the metrics, but I think requiring the method to be more explicit when collecting metrics is a better move.

src/client-side-metrics/metrics-tracer-factory.ts
danieljbruce: This class doesn't exist anymore. There is now just a metrics collector class which can use a user defined metrics handler or a metrics handler that sends metrics to the GCP dashboard.

common/test-date-provider.ts
```diff
@@ -0,0 +1,56 @@
...
+// limitations under the License.
+
+import {WithLogger} from './logger';
+
+/**
+ * A test implementation of a Date-like object.  Used for testing purposes.  It provides a
+ * getTime method that returns a pre-determined fake date value, allowing for
+ * deterministic testing of time-dependent functionality.
+ */
+class TestDateLike {
```
daniel-sanche, line 22: Why do we have these test files in `common/`?

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,189 @@
...
+      this.initialized = true;
+      // Use MeterProvider provided by user
+      // If MeterProvider was not provided then use the default meter provider.
+      const meterProvider = new MeterProvider({
+        // This is the default meter provider
+        // Create a resource. Fill the `service.*` attributes in with real values for your service.
+        // GcpDetectorSync will add in resource information about the current environment if you are
+        // running on GCP. These resource attributes will be translated to a specific GCP monitored
+        // resource if running on GCP. Otherwise, metrics will be sent with monitored resource
+        // `generic_task`.
```
daniel-sanche, line 75: are these comments intentional?

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,189 @@
...
+          new PeriodicExportingMetricReader({
+            // Export metrics every 10 seconds. 5 seconds is the smallest sample period allowed by
+            // Cloud Monitoring.
+            exportIntervalMillis: 100_000,
+            exporter: new MetricExporter({
+              projectId,
+            }),
+          }),
+        ],
+      });
```
daniel-sanche, line 90: I think you'll also need to include aggregation views: https://github.com/googleapis/python-bigtable/blob/987e3df4ce64a6079e03295f653363f1cc391e0b/google/cloud/bigtable/data/_metrics/handlers/gcp_exporter.py#L73

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,189 @@
...
+   * @param {OnOperationCompleteAttributes} attributes Attributes associated with the completed operation.
+   */
+  onOperationComplete(
+    metrics: OnOperationCompleteMetrics,
+    attributes: OnOperationCompleteAttributes
+  ) {
+    this.initialize();
+    this.otelMetrics?.operationLatencies.record(
+      metrics.operationLatency,
+      attributes
```
daniel-sanche, line 158: are you sure you attach all attributes?

What about potentially empty fields like zone?

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,189 @@
...
+   * This method records the operation latency and retry count, associating them with provided attributes.
+   * @param {OnOperationCompleteMetrics} metrics Metrics related to the completed operation.
+   * @param {OnOperationCompleteAttributes} attributes Attributes associated with the completed operation.
+   */
+  onOperationComplete(
+    metrics: OnOperationCompleteMetrics,
+    attributes: OnOperationCompleteAttributes
+  ) {
+    this.initialize();
+    this.otelMetrics?.operationLatencies.record(
```
daniel-sanche, line 156: why is otelMetrics optional?

src/client-side-metrics/metrics-collector.ts
```diff
@@ -0,0 +1,329 @@
...
+ */
+export interface ITabularApiSurface {
+  instance: {
+    id: string;
+  };
+  id: string;
+  bigtable: {
+    appProfileId?: string;
+  };
+}
```
daniel-sanche, line 72: Are these interfaces necessary?

If they're mostly for mocking, can they live under `/tests`?

src/client-side-metrics/metrics-collector.ts
```diff
@@ -0,0 +1,329 @@
...
+   */
+  private getOperationLatencyAttributes(
+    projectId: string,
+    operationOnlyAttributes: OperationOnlyAttributes
+  ): OnOperationCompleteAttributes {
+    return Object.assign(
+      operationOnlyAttributes,
+      this.getBasicAttributes(projectId)
+    );
+  }
```
daniel-sanche, line 164: Can you explain this a bit more? I'm having a hard time following all the different layers (StandardAttributes/OperationOnlyAttributes/onOperationCompleteAttributes/getOperationLatencyAttributes/getBasicAttributes)

Especially since they are spread across multiple files. It seems a bit much

src/client-side-metrics/metrics-collector.ts
```diff
@@ -0,0 +1,329 @@
...
+
+  /**
+   * Assembles the attributes for attempt metrics. These attributes provide context
+   * about the Bigtable environment, the operation being performed, and the status of the attempt.
+   * Includes whether the operation was a streaming operation or not.
+   * @param {string} projectId The Google Cloud project ID.
+   * @param {AttemptOnlyAttributes} attemptOnlyAttributes The attributes of the attempt.
+   * @returns {OnAttemptCompleteAttributes} An object containing the attributes for attempt metrics.
+   */
```
daniel-sanche, line 173:
```
   * @param {AttemptOnlyAttributes} attemptOnlyAttributes The attributes of the attempt.
   * @returns {OnAttemptCompleteAttributes} An object containing the attributes for attempt metrics.
```

This is not clear to me

src/client-side-metrics/metrics-collector.ts
```diff
@@ -0,0 +1,329 @@
...
+ */
+export class MetricsCollector {
+  private operationStartTime: DateLike | null;
+  private attemptStartTime: DateLike | null;
+  private zone: string | undefined;
+  private cluster: string | undefined;
+  private tabularApiSurface: ITabularApiSurface;
+  private methodName: string;
+  private projectId?: string;
+  private attemptCount = 0;
```
daniel-sanche, line 88: is a MetricsCollector created for each operation? 

If it's not global, I'd probably suggest a different name, like `OperationMetricsCollector`. But this also makes me wonder why we need the operation attributes classes, since it seems like the same data is here?

src/client-side-metrics/metrics-collector.ts
```diff
@@ -0,0 +1,329 @@
...
+const packageJSON = fs.readFileSync('package.json');
+const version = JSON.parse(packageJSON.toString()).version;
+
+/**
+ * A class for tracing and recording client-side metrics related to Bigtable operations.
+ */
+export class MetricsCollector {
+  private operationStartTime: DateLike | null;
+  private attemptStartTime: DateLike | null;
+  private zone: string | undefined;
```
daniel-sanche, line 83: I mentioned in a different comment that state is duplicated across the different attribute classes. It looks like there's also duplication here?

src/client-side-metrics/metrics-collector.ts
```diff
@@ -0,0 +1,329 @@
...
+  /**
+   * Called when an operation completes (successfully or unsuccessfully).
+   * Records operation latencies, retry counts, and connectivity error counts.
+   * @param {OperationOnlyAttributes} info Information about the completed operation.
+   */
+  onOperationComplete(info: OperationOnlyAttributes) {
+    const endTime = this.dateProvider.getDate();
+    const projectId = this.projectId;
+    if (projectId && this.operationStartTime) {
+      const totalTime = endTime.getTime() - this.operationStartTime.getTime();
```
daniel-sanche, line 254: consider some kind of protection to make sure this doesn't get called multiple times

In Python, we used a state machine to track the lifecycle

src/client-side-metrics/metrics-handler.ts
daniel-sanche: I'm not sure I really see it that way. In practice, all current metrics essentially **do** have the same dimensions, and they are pretty much all finalized when a status is passed down, either on the completion of an attempt or an operation. So knowing that, doing a single callback for each attempt should have better performance, and performance really matters for this (especially for slow languages like Python, and I assume node)

But it's hard to say what the performance impact will be in practice. The bigtable team will probably want to see benchmarks before and after this change, so if you think this is worth pursuing, it may be worth playing with as part of the benchmarks

daniel-sanche (changes requested):

common/client-side-metrics-attributes.ts
```diff
@@ -0,0 +1,80 @@
+// Copyright 2025 Google LLC
```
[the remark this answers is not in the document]
danieljbruce: It is located here because it is going to be used by both tests and system tests.

common/client-side-metrics-attributes.ts
```diff
@@ -0,0 +1,80 @@
...
+// limitations under the License.
+
+interface StandardAttributes {
+  projectId: string;
+  instanceId: string;
+  table: string;
+  cluster?: string;
+  zone?: string;
+  appProfileId?: string;
+  methodName: string;
```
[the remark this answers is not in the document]
danieljbruce: An enum has now been added here.

common/client-side-metrics-attributes.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export enum AttemptStatus {
+  OK = 'OK',
+  ERROR = 'ERROR',
+}
+
+export enum StreamingOperation {
+  YES = 'YES',
+  NO = 'NO',
+}
```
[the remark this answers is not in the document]
danieljbruce: Just because I'm sure of how a string appears in the exported metrics, but not sure about how a boolean value appears. That said, I'm sure that we could try a boolean value out and then if it doesn't show up in the metrics we can easily change it back.

common/client-side-metrics-attributes.ts
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * The attempt status of the operation.
+   */
+  attemptStatus: AttemptStatus;
+}
+
+export enum FinalOperationStatus {
+  OK = 'OK',
+  ERROR = 'ERROR',
+}
```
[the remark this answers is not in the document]
danieljbruce: Ok. I wasn't clear about the values. Should `FinalOperationStatus` be replaced with a number? Same with `AttemptStatus`?

common/logger.ts
```diff
@@ -0,0 +1,35 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * A simple logger interface for logging messages.  Implementations of this interface
+ * can provide various logging mechanisms (e.g., console logging, file logging, etc.).
+ */
```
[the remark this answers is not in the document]
danieljbruce: This is used by the other test utilities in the `common` folder so that they can log messages to be compared in the test.

common/test-date-provider.ts
danieljbruce: Mainly because they are going to be used by both system-tests and tests. At another time I tried importing a file in the system tests from the test folder and it ran both test suites when I ran `npm run system-test`. So these shared files have to live somewhere other than `test` or `system-test`.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: This comment was a copy/paste from the code copied from the npm registry used for exporting metrics to GCP. I just removed it.

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,189 @@
...
+        resource: new Resources.Resource({
+          'service.name': 'bigtable-metrics',
+        }).merge(new ResourceUtil.GcpDetectorSync().detect()),
+        readers: [
+          // Register the exporter
+          new PeriodicExportingMetricReader({
+            // Export metrics every 10 seconds. 5 seconds is the smallest sample period allowed by
+            // Cloud Monitoring.
+            exportIntervalMillis: 100_000,
+            exporter: new MetricExporter({
```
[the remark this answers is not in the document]
danieljbruce: I tried this out and saw the metrics appear in the dashboard. We'll be able to see this in action for sure in the next PR containing a system test.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: `otelMetrics` is optional because `otelMetrics` are not defined in the constructor which is what would be necessary if it wasn't optional. Instead, `otelMetrics` are defined in the initialize function. It has to be this way because the metrics handlers get stored at client creation time, but the `projectId` isn't available until a method is called and the project Id is used when creating the MetricsProvider which creates the `otelMetrics`.

If we wanted to make `otelMetrics` not optional we could create this handler in the method instead of at client initialization time. That is a possible alternative.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: Yes. This is attached in the metrics collector. You can also see that it appears in the output in the `typical-method-call.txt` file.

src/client-side-metrics/metrics-collector.ts
danieljbruce: They are used to accomodate tests yes, but the interfaces are used directly in the `MetricsCollector` to make the class more flexible so that the test can use them. For instance, a `ITabularApiSurface` object is passed into the `MetricsCollector` constructor instead of a `TabularApiSurface` object so that the test can use the object without having to define all the methods that `TabularApiSurface` has.

This follows the principle of program to an interface, not an implementation.

Do you still want me to move these to tests folder?

src/client-side-metrics/metrics-collector.ts
danieljbruce: I'm assuming by the other comment you mean the discussion about the `OnOperationCompleteAttributes`, `StandardAttributes`, `OperationOnlyAttributes` and `OnAttemptCompleteAttributes`, `StandardAttributes` and `AttemptOnlyAttributes` objects? These are interfaces and do not store a zone.

src/client-side-metrics/metrics-collector.ts
danieljbruce: There are no operation attributes classes. Those are interfaces. They are for assigning the parameters types. They don't store data.

danieljbruce: I just renamed it to `OperationMetricsCollector`.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: Okay. This has been added now.

src/client-side-metrics/metrics-handler.ts
danieljbruce: I did some measurements:
```
Before onOperationComplete
recording metrics
callback: 0.233ms
onOperationComplete: 0.394ms
After onOperationComplete
```

It takes 0.233ms to get from the fake method to the metrics handler so with this new design it will have to do that times 2. Is adding an additional 0.233ms unacceptable?

My main point about the dimensions is that while we do expect them to be basically the same for all metrics right now, that could easily not be the case with new metrics in the future. It doesn't seem like a good idea to commit to this pattern being always true if we have the opportunity right now to be more flexible and let the user decide what happens with several callbacks that each just provide one metric at a time. But maybe its not worth the performance tradeoff.

What do you think?

src/client-side-metrics/metrics-collector.ts
danieljbruce: This has been revised

src/client-side-metrics/metrics-collector.ts
danieljbruce: A state machine is now in the PR

common/client-side-metrics-attributes.ts
```diff
@@ -0,0 +1,80 @@
...
+  NO = 'NO',
+}
+
+export interface OnOperationCompleteAttributes
+  extends StandardAttributes,
+    OperationOnlyAttributes {}
+
+export interface OnAttemptCompleteAttributes
+  extends StandardAttributes,
+    AttemptOnlyAttributes {}
```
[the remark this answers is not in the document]
danieljbruce: I thought about this again and I think I am going to create an interface of attributes for every metric and then assign this interface to a union of the interfaces for the metrics it supports.

common/client-side-metrics-attributes.ts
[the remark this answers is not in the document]
daniel-sanche: Maybe call it `test-common` then?

But I'm confused, is this entire attributes file all just for test mocks then?

common/logger.ts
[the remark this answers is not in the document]
daniel-sanche: Is there a reason we can't use console.log directly? What other implementations of this interface do you plan to add?

src/client-side-metrics/gcp-metrics-handler.ts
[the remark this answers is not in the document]
daniel-sanche: FWIW, you need to point this to the private bigtable end-point, not your own GCP project. You shouldn't be seeing these metrics in your own dashboard. Unless you're making separate handlers for this?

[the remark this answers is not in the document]
daniel-sanche: Also, do you have any plans of exposing the raw otel instruments?

src/client-side-metrics/gcp-metrics-handler.ts
daniel-sanche: > It has to be this way because the metrics handlers get stored at client creation time, but the projectId isn't available until a method is called and the project Id is used when creating the MetricsProvider which creates the otelMetrics

Can you expand on this a bit? What method needs to be called? When is this handler created currently?

We don't want to end up in the state where onOperationComplete is called with an empty otelMetrics, right? Assuming that's a bad state, we should use the type checker to help us avoid it

src/client-side-metrics/metrics-handler.ts
daniel-sanche: > It takes 0.233ms to get from the fake method to the metrics handler so with this new design it will have to do that times 2. 

I don't think I understand what those numbers mean. Why is it times 2? Shouldn't it be times (num_metrics)?

> Is adding an additional 0.233ms unacceptable?

That depends on how often the calls are being made. If callbacks were being triggered on each row, it would definitely add up (napkin math: bigtable targets ~10k rows per second, so each row should take 0.1ms). But I think we agreed they'd only be triggered at the end of each rpc, so I'm not sure how relevant it would be there

I was suggesting doing a full throughput benchmark, where you do as many read_row calls as you can over ~10 mins, and see what the impact is. We've done that in other languages, but I'm not sure if you've done that kind of benchmark for node before. But the bigtable team might ask about it before merging

> seem like a good idea to commit to this pattern being always true if we have the opportunity right now to be more flexible and let the user decide what happens with several callbacks that each just provide one metric at a time

My reasoning: we want to make sure the system can scale well. I can imagine both the number of metrics and the number of handlers scaling over time, and at some point, managing hundreds of callbacks on each rpc would become unmanageable. Since we expect most fields to be the same across all metrics, we can generalize down the API into something smaller and more manageable, where the handler just receives the raw data associated with each attempt, and uses that to build all metrics at once.

But really, I think we should be treating this as an internal implementation detail for now, so we shouldn't have to worry about lock-in. We can make the MetricHandler interface public at some point, if users request exporting metrics to other locations. But for now, we can just use it internally to build the handlers we need, and change it as needed

I'd just say do whatever feels cleanest/most performant for the requirements we have now, and we can re-visit later if requirements change

common/logger.ts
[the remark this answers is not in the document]
danieljbruce: This interface is not here to replace console.log. One implementation of it is shown below and used in the test utility classes:

```
/**
 * A basic logger class that stores log messages in an array. Useful for testing.
 */
class Logger {
  private messages: string[] = [];

  /**
   * Logs a message by adding it to the internal message array.
   * @param {string} message The message to be logged.
   */
  log(message: string) {
    this.messages.push(message);
  }

  /**
   * Retrieves all logged messages.
   * @returns {string[]} An array of logged messages.
   */
  getMessages() {
    return this.messages;
  }
}
```

This log collects strings that the test utilities log and then when the test is almost finished it compares the strings collected with a text file that you can see in this PR.

common/client-side-metrics-attributes.ts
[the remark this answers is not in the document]
danieljbruce: I moved a couple of interfaces that don't need to be in this file. This client-side-metrics-attributes.ts file is used in a bunch of places as shown above. The other files `logger.ts`, `test-date-provider.ts` and `test-metrics-handler.ts` will specifically be used for tests.

What if I do this. I rename this folder `test-common` and then I move the `client-side-metrics-attributes.ts` file out of here into `src/client-side-metrics` folder. How does that sound?

src/client-side-metrics/gcp-metrics-handler.ts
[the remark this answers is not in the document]
danieljbruce: I don't plan to expose the otel instruments no. Maybe I'll expose the `GCPMetricsHandler`. The exact script where I saw the metrics in the dashboard is the following:

```
const { MeterProvider, PeriodicExportingMetricReader } = require("@opentelemetry/sdk-metrics");
const { Resource } = require("@opentelemetry/resources");
const { MetricExporter } = require("@google-cloud/opentelemetry-cloud-monitoring-exporter");
const { GcpDetectorSync } = require("@google-cloud/opentelemetry-resource-util");

// Create MeterProvider
const meterProvider = new MeterProvider({
  // Create a resource. Fill the `service.*` attributes in with real values for your service.
  // GcpDetectorSync will add in resource information about the current environment if you are
  // running on GCP. These resource attributes will be translated to a specific GCP monitored
  // resource if running on GCP. Otherwise, metrics will be sent with monitored resource
  // `generic_task`.
  resource: new Resource({
    "service.name": "example-metric-service",
    "service.namespace": "samples",
    "service.instance.id": "12345",
    "cloud.resource_manager.project_id": "cloud-native-db-dpes-shared"
  }).merge(new GcpDetectorSync().detect()),
  readers: [ // Register the exporter
    new PeriodicExportingMetricReader({
      // Export metrics every 10 seconds. 5 seconds is the smallest sample period allowed by
      // Cloud Monitoring.
      exportIntervalMillis: 10_000,
      exporter: new MetricExporter({
        projectId: 'cloud-native-db-dpes-shared'
      }),
    })
  ]
});

// Create a meter
const meter = meterProvider.getMeter("metrics-sample");

// Create a counter instrument
const counter = meter.createCounter("metric_name79");
// Record a measurement
counter.add(10, { key: "value" });
counter.add(5, { key: "value" });

// Wait for the metric to be exported
new Promise((resolve) => {
  setTimeout(async () => {
    await meterProvider.shutdown();
    resolve();
  }, 30_000);
});
```

We don't want to use the Bigtable client's project? Well that would certainly allow me to simplify some of the architecture.

common/client-side-metrics-attributes.ts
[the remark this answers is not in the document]
danieljbruce: I confirmed with Mattie that these statuses should be replaced with grpc status codes.

src/client-side-metrics/gcp-metrics-handler.ts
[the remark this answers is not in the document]
danieljbruce: Should I remove the projectId here? That would certainly make things easier. What is the private Bigtable endpoint and how can I see if metrics are being written there?

src/client-side-metrics/metrics-handler.ts
danieljbruce: > I don't think I understand what those numbers mean. Why is it times 2? Shouldn't it be times (num_metrics)?

My mistake. It _would be_ times num_metrics with my previous analysis. But now that I think about it the time it takes to get through the metrics collector isn't really relevant at all since the same number of calls will be made to the metrics collector. The change would actually be that in `onOperationComplete` of the metrics collector the metrics collector will actually make `num_metrics` calls to the metrics handler instead of 1 call. I don't think this will add any more time, but we can still do performance benchmarking I guess.

> But I think we agreed they'd only be triggered at the end of each rpc

Yes. Extra processing time only per rpc.

> managing hundreds of callbacks on each rpc would become unmanageable

Keep in mind that this would not add more calls to the metricsCollector from createReadStream, mutateRows etc.

> we can generalize down the API into something smaller and more manageable

With the design I proposed certainly if the user wanted to custom handle hundreds of metrics then they would have to provide hundreds of methods.

> I think we should be treating this as an internal implementation detail for now,

This design isn't internal because we are deciding on the API of the metrics handlers that users provide. It certainly doesn't need to be finalized in this PR because this PR doesn't support metrics handlers yet, but it will be important to get this right.

> I'd just say do whatever feels cleanest/most performant for the requirements we have now

Okay. Let me get back to you on this.

danieljbruce: https://github.com/googleapis/nodejs-bigtable/pull/1566/commits/2781561f3791868855e0daf76d90d4f80736beca is what the API change would look like. Let's leave this out for now since we can easily include/not include this in future PRs since this PR doesn't affect customers yet. It'll be easier to see this change in isolation.

src/client-side-metrics/gcp-metrics-handler.ts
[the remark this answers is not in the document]
danieljbruce: I looked into this and it looks like it is as easy as passing an apiEndpoint into the MetricsExporter constructor.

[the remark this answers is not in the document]
danieljbruce: I looked at [your exporter](https://github.com/googleapis/python-bigtable/pull/924/files#diff-142d623e0503afa707461f3bd80d8808a2e10232038c24d7d69ce1df6bd4dbceR138) and it looks like I'll have to make something like this in order to see the metrics in the dashboard because I can't see them with the code I am using.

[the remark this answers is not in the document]
danieljbruce: I can now see the metrics with this code, but it's looking like to get metrics to appear under the "Cloud Bigtable Table" resource we may need a separate exporter. Can we punt the introduction of this file to a future PR? It is not being used right now anyway.

src/client-side-metrics/metrics-handler.ts
daniel-sanche: Ok, yeah I'm not too concerned either way as long as the benchmarks are fine. But:

> This design isn't internal because we are deciding on the API of the metrics handlers that users provide. It certainly doesn't need to be finalized in this PR because this PR doesn't support metrics handlers yet, but it will be important to get this right.

Are you sure you want users creating custom handlers?

We discussed designing the handlers API to be easily extensible, but the main reason for that was for future internal changes, so we can add more handlers in the future if requirements change. It would be easy enough to let users plug into it too, but we can probably wait for demand before marking it as public. And that gives us a bit of time to iterate on it internally if we run into limitations in the design

common/client-side-metrics-attributes.ts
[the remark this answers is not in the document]
danieljbruce: Alright. I made this change.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: This file isn't part of the PR anymore. Let's revisit this conversation when we start to add the open telemetry stuff.

src/client-side-metrics/metrics-handler.ts
danieljbruce: I thought we wanted to allow users to provide their own handlers, but if we don't have to do that then that would be great because it doesn't commit us to an API. I didn't realize there was no demand for them. If we don't allow the user to provide handlers then this decision just becomes an internal decision and doesn't matter as much because we can always change it later.

common/client-side-metrics-attributes.ts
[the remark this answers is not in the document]
danieljbruce: Even though FinalOperationStatus and AttemptStatus are just aliases for the grpc.status type, I still think it is useful throughout the codebase for the maintainer to know if a variable intends to track a FinalOperationStatus or an AttemptStatus. That said, I appreciate the simplicity of your suggestion so I decided to make the change you suggested.

src/client-side-metrics/metrics-collector.ts
danieljbruce: With the new metrics handler API the call stack can be simplified so I removed getBasicAttributes, getOperationAttributes, getAttemptAttributes.

common/logger.ts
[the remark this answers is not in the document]
danieljbruce: Everything this logger does can be accomplished by an object wrapping a string so I removed the logger class. Now the object wrapping the string gets passed around, we append values to the string and then the string gets compared at the end of the test.

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,356 @@
...
+      options: {};
+    }
+  ) {
+    const mappedEntries = new Map(
+      Array.from(metadata.internalRepr.entries(), ([key, value]) => [
+        key,
+        value.toString(),
+      ])
+    );
+    const durationValues = mappedEntries.get('server-timing')?.split('dur=');
```
daniel-sanche, line 320: I don't think parsing the header is quite this simple IIRC. See go/gfe-server-timing. It can contain multiple durations, and we need to check for `gfet4t7`

I used [this regex in python](https://github.com/googleapis/python-bigtable/blob/b1d08c9662c92bd97bcb3e9b2fd0282b7faf38b1/google/cloud/bigtable/data/_metrics/data_model.py#L46). I can't remember if I got it from somewhere

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,356 @@
...
+    if (durationValues && durationValues[1]) {
+      if (!this.serverTimeRead) {
+        this.serverTimeRead = true;
+        const serverTime = parseInt(durationValues[1]);
+        if (projectId) {
```
daniel-sanche, line 325: why check the projectID way down here?

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,356 @@
...
+    metadata: {internalRepr: Map<string, Buffer>; options: {}};
+  }) {
+    const mappedEntries = new Map(
+      Array.from(status.metadata.internalRepr.entries(), ([key, value]) => [
+        key,
+        value.toString(),
+      ])
+    );
+    const instanceInformation = mappedEntries
+      .get('x-goog-ext-425905942-bin')
```
daniel-sanche, line 346: replacing these magic strings with constants would be helpful

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,356 @@
...
+      ?.replace(new RegExp('\\n', 'g'), '')
+      .split(' \r'); // The data returned actually has a space after the zone.
+    if (instanceInformation && instanceInformation[0]) {
+      this.zone = instanceInformation[0];
+    }
+    if (instanceInformation && instanceInformation[1]) {
+      this.cluster = instanceInformation[1];
+    }
```
daniel-sanche, line 354: I'm not super confident in this parsing logic. Are you sure these structure assumptions are safe? What if the cluster is present, but not zone? It seems like [the proto](https://github.com/googleapis/python-bigtable/blob/89b8da8a445aeb08854d9fa77cbc0e4fc042c87f/google/cloud/bigtable_v2/types/response_params.py#L31C7-L31C21) marks them both as optional

daniel-sanche (changes requested):

src/client-side-metrics/client-side-metrics-attributes.ts
```diff
@@ -0,0 +1,138 @@
...
+ * operation.  These attributes provide context about the Bigtable environment,
+ * the specific attempt, its status, and whether the operation was streaming. They
+ * are used for recording metrics such as attempt latency, server latency, and
+ * connectivity errors.
+ */
+export type OnAttemptCompleteAttributes =
+  | AttemptLatencyAttributes
+  | ConnectivityErrorCountAttributes
+  | ServerLatenciesAttributes
+  | ClientBlockingLatenciesAttributes;
```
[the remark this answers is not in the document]
danieljbruce: They have the same set of attributes for the most part, but they can be slightly different. Okay, I've left AttemptLatencyAttributes, ConnectivityErrorCountAttributes, ServerLatenciesAttributes, ClientBlockingLatenciesAttributes out for now and defined the OnAttemptCompleteAttributes interface without any aliases, but I might introduce those interfaces again in GCPMetricsHandler PR.

src/client-side-metrics/metrics-handler.ts
danieljbruce: Ok. I asked about this, looks like we only need to support collecting GCM metrics and allowing users to opt out of metrics collection for now.

src/client-side-metrics/metrics-handler.ts
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Called when an operation completes (successfully or unsuccessfully).
+   * @param {OnOperationCompleteMetrics} metrics Metrics related to the completed operation.
+   * @param {OnOperationCompleteAttributes} attributes Attributes associated with the completed operation.
+   */
+  onOperationComplete?(
+    metrics: OnOperationCompleteMetrics,
+    attributes: OnOperationCompleteAttributes
```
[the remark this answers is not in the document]
danieljbruce: Yes. OperationLatencyAttributes | FirstResponseLatencyAttributes | RetryCountAttributes interfaces are gone now so we can expect all attributes to be provided here that will be used in any one of these metrics.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: Yes. Thanks for sharing that.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: I removed projectId from here. This must be a leftover artifact.

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,356 @@
...
+            streamingOperation: info.streamingOperation,
+            projectId,
+            instanceId: this.tabularApiSurface.instance.id,
+            table: this.tabularApiSurface.id,
+            cluster: this.cluster,
+            zone: this.zone,
+            appProfileId: this.tabularApiSurface.bigtable.appProfileId,
+            methodName: this.methodName,
+            clientName: `nodejs-bigtable/${version}`,
+          };
```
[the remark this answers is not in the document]
danieljbruce: Aw, but even that doesn't work because all these data members are private. So if I pass `this` along then its private properties that the handler needs won't be accessible. But I could add a helper method called `getMetricsCollectorData` that will build the object I need. This is kind of like what I had before.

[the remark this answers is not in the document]
danieljbruce: Ok. I created a `IMetricsCollectorData` object to pass the appropriate data along.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: The only way to be super confident in this parsing logic is to know what the server data looks like when there is a missing zone or missing cluster and I'm not really sure if there is any way to simulate that. Do you know if there is?

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: Done

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,356 @@
...
+      Array.from(metadata.internalRepr.entries(), ([key, value]) => [
+        key,
+        value.toString(),
+      ])
+    );
+    const durationValues = mappedEntries.get('server-timing')?.split('dur=');
+    if (durationValues && durationValues[1]) {
+      if (!this.serverTimeRead) {
+        this.serverTimeRead = true;
+        const serverTime = parseInt(durationValues[1]);
```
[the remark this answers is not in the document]
danieljbruce: This is a good catch. We probably want falsy values passed into the metrics handler not NaN.

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,356 @@
...
+  /**
+   * Returns the time value in milliseconds.
+   * @returns The time value in milliseconds.
+   */
+  getTime(): number;
+}
+
+/**
+ * Interface for a provider that returns DateLike objects. Used for mocking dates in tests.
+ */
```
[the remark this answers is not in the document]
danieljbruce: The date providers have all been removed now. I found a way to mock out date by overriding the Date global singleton.

src/client-side-metrics/metrics-handler.ts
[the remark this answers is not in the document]
daniel-sanche: Thanks, this is much cleaner! Although I'm still a little unsure why we need to divide the data up between OnOperationCompleteAttributes and OnOperationCompleteMetrics though, since it seems like they're all used in the same places. Could it be simplified a bit more if you combine them? Or am I missing something?

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,337 @@
...
+    }
+  }
+
+  /**
+   * Called when an operation completes (successfully or unsuccessfully).
+   * Records operation latencies, retry counts, and connectivity error counts.
+   * @param {string} projectId The id of the project.
+   * @param {OperationOnlyAttributes} info Information about the completed operation.
+   */
+  onOperationComplete(projectId: string, info: OperationOnlyAttributes) {
```
daniel-sanche, line 247: Same question as onAttemptComplete: don't we just need the status code here?

src/client-side-metrics/operation-metrics-collector.ts
daniel-sanche: It's not safe to assume that the server responses you get in testing are the only ones that will be returned.

You'll have to look at the spec, but my understanding is that the server returns a binary-encoded version of the [ResponseParams proto](https://github.com/googleapis/python-bigtable/blob/89b8da8a445aeb08854d9fa77cbc0e4fc042c87f/google/cloud/bigtable_v2/types/response_params.py#L31C7-L31C21). In Python, [we use ResponseParams.FromString](https://github.com/googleapis/python-bigtable/blob/b1d08c9662c92bd97bcb3e9b2fd0282b7faf38b1/google/cloud/bigtable/data/_metrics/data_model.py#L224) to convert it back into the proto object, and then read from the proto fields

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,337 @@
...
+    } else {
+      this.connectivityErrorCount++;
+    }
+  }
+
+  /**
+   * Called when status information is received. Extracts zone and cluster information.
+   * @param {object} status The received status information.
+   */
+  onStatusReceived(status: {
```
daniel-sanche, line 316: I'm a little confused why onStatusReceived is different from onMetadataReceived. Aren't they both parsing metadata fields? (I noticed the argument is called metadata)

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,337 @@
...
+      console.warn('Invalid state transition');
+    }
+  }
+
+  /**
+   * Called when an attempt (e.g., an RPC attempt) completes. Records attempt latencies.
+   * @param {string} projectId The id of the project.
+   * @param {OnAttemptCompleteInfo} info Information about the completed attempt.
+   */
+  onAttemptComplete(projectId: string, info: OnAttemptCompleteInfo) {
```
daniel-sanche, line 166: why does a StreamingState need to be passed here? and project id? shouldn't they be known in advance?

It seems to me that the only new information here would be the status code

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,337 @@
...
+interface DateLike {
+  /**
+   * Returns the time value in milliseconds.
+   * @returns The time value in milliseconds.
+   */
+  getTime(): number;
+}
+
+/**
+ * An interface representing a tabular API surface, such as a Bigtable table.
```
daniel-sanche, line 34: why does this exist here? Isn't there [already a class for the TabularAPISurface](https://github.com/googleapis/nodejs-bigtable/blob/5da73e1eec10a4729d158ebcc181df88497ea5c2/src/tabular-api-surface.ts#L155)?

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,337 @@
...
+
+import * as fs from 'fs';
+import {IMetricsHandler} from './metrics-handler';
+import {MethodName, StreamingState} from './client-side-metrics-attributes';
+import {grpc} from 'google-gax';
+
+/**
+ * An interface representing a Date-like object.  Provides a `getTime` method
+ * for retrieving the time value in milliseconds.  Used for abstracting time
+ * in tests.
```
daniel-sanche, line 23: is this really necessary? There's not a way to mock a stdlib Date without this extra interface?

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,337 @@
...
+    clientUid: string;
+  };
+}
+
+/**
+ * Information about the completion of a single attempt of a Bigtable operation.
+ * This information is used for recording metrics.
+ */
+interface OnAttemptCompleteInfo {
+  connectivityErrorCount: number;
```
daniel-sanche, line 52: is this connectivityErrorCount supposed to be here? I see it's also in OnAttemptCompleteMetrics, which seems more natural

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -0,0 +1,337 @@
...
+  /**
+   * Whether the operation is a streaming operation or not.
+   */
+  streamingOperation: StreamingState;
```
daniel-sanche, line 56: Why is this part of OnAttemptComplete? Isn't it known in advance?

My understanding is this is basically just to check if the user called the public `read_row` vs `read_rows` method, which should be known at call time? I'm not sure if that even applies to node though, since I'm not seeing a point read method.

How were you planning on populating this?

daniel-sanche (changes requested):

danieljbruce:
In response to https://github.com/googleapis/nodejs-bigtable/pull/1566#discussion_r1956772106. I have made this method require just one argument.

src/client-side-metrics/metrics-handler.ts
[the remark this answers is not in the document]
danieljbruce: I changed this to match your suggestion so now there is just one argument. I'm indifferent between the two choices, dividing them up makes the variable purposes more clear, but combining them simplifies some internal logic.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: We could definitely pass in the _streamingOperation_ parameter when we create the metrics collector instead so we don't need to pass it in here. _projectId_ is a different story, because the projectId property on the client is only populated by calling auth just before the grpc call is made so it's not available when the metrics collector gets created.

danieljbruce: Ok _streamingOperation_ has been removed.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: We could definitely pass in the streamingOperation parameter when we create the metrics collector instead so we don't need to pass it in here. projectId is a different story, because the projectId property on the client is only populated by calling auth just before the grpc call is made so it's not available when the metrics collector gets created.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: When you create a stream in Node and make a grpc call, the stream receives `status` and `metadata` events. The shape of the status event received matches the first argument of onStatusReceived method and the shape of the metadata event received matches the first argument of onMetadataReceived.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: An `ITabularApiSurface` parameter is provided when the metrics collector is created. We don't want to require this parameter to be a whole `TabularApiSurface` object because only a small subset of the properties are actually used and we don't want to restrict the `OperationMetricsCollector` further than we have to. In particular, this simplifies the test for the metrics collector where we create a simple `FakeTable` object which would otherwise have to be a whole `TabularApiSurface` object.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: This is no longer needed since we removed the date provider.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: Yeah. This is a typo and OnAttemptCompleteInfo is gone now because after other corrections it only contains one property so that property is just passed directly.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: >My understanding is the streaming flag is basically just to check if the user called the public read_row vs read_rows method, which should be known at call time?

That's my understanding too. This is now populated when the metrics collector gets created.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: The custom parsing logic is gone now and we use ResponseParams.

daniel-sanche (approved):
I left a comment to consider about de-duplicating the attempt/operation data. But overall looks good!

src/client-side-metrics/metrics-handler.ts
```diff
@@ -0,0 +1,72 @@
...
+export interface OnAttemptCompleteData {
+  attemptLatency: number;
+  serverLatency?: number;
+  connectivityErrorCount: number;
+  projectId: string;
+  metricsCollectorData: IMetricsCollectorData;
+  clientName: string;
+  attemptStatus: grpc.status;
+  streamingOperation: StreamingState;
+}
```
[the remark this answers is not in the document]
danieljbruce: I had something similar before with `StandardAttributes` when the metrics and attributes were separate so I decided to de-duplicate the data with a new StandardData interface and then extend it.

PR #1567, opened 2025-01-21
chore(testproxy): Update known failures

danieljbruce:
New tests have been added to the test proxy so we have to update the known failures file


daniel-sanche (approved):

cindy-peng (approved):

bhshkh:
I was facing the same issue and have updated the conformance test flow to use v0.0.3 instead of main.
There might be many more such failed tests in the future in the main branch and that will require updates in known_failures file

bhshkh (approved):

PR #1569, opened 2025-02-18
feat: Add plumbing PR for client side metrics to support the open telemetry instruments

danieljbruce:
**Summary:**

In this PR we introduce all the plumbing that runs the operations from the point where the metrics handler receives data to the point where the data is sent to the Google Cloud Monitoring dashboard.

The diagram below shows all the pieces added in this PR in green:

2nd plumbing PR

A test stub exists that contains the requests that the metrics collector passes into its metrics handler when a fake method is used. This test stub is used to pass data into the metrics handler in this PR and test the new pipeline that this PR adds.

This is a follow up PR to https://github.com/googleapis/nodejs-bigtable/pull/1566.

**Changes:**

`package.json`: Add the package that contains the MetricServiceClient used for sending data to Google Cloud Monitoring.

`src/client-side-metrics/exporter.ts`: Add the CloudMonitoringExporter which accepts metrics and then converts them to the right format and sends them to the Google Cloud Monitoring dashboard.

`src/client-side-metrics/gcp-metrics-handler.ts`: The metrics handler that stores requests in open telemetry instruments and then periodically sends them to its exporter.

`system-test/cloud-monitoring-exporter.ts`: System tests that write to the google cloud monitoring dashboard with a fixed value passed into the exporter.

`system-test/gcp-metrics-handler.ts`: A system test that passes fixed values into the `gcp-metrics-handler.ts` and ensures it properly writes information to the Google Cloud dashboard.

`test-common/expected-otel-export-input.ts`: Some fixed constants related to values open telemetry provides to the exporter.

`test/metrics-collector/gcp-metrics-handler.ts`: A test that ensures the GCP Metrics handler passes the right information to the exporter.

`test/metrics-collector/metrics-to-request.ts`: A test that ensures the method that converts the value the exporter receives to the value that GCM needs is correct.

**Next Steps:**

- Add client side metrics collection to read-rows operations
- Add application blocking latencies and throttling latencies


src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,460 @@
...
+              client_uid: allAttributes.clientUid,
+            },
+            streaming ? {streaming} : null
+          );
+          const timeSeries = {
+            metric: {
+              type: metricName,
+              labels: metricLabels,
+            },
+            resource: {
```
danieljbruce, line 349: Note: This is where the monitored resource is provided.

danieljbruce: image

src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,460 @@
...
+                (
+                  allAttributes as OnOperationAttribute
+                ).finalOperationStatus?.toString(),
+              client_uid: allAttributes.clientUid,
+            },
+            streaming ? {streaming} : null
+          );
+          const timeSeries = {
+            metric: {
+              type: metricName,
```
danieljbruce, line 346: This really ends up being the same thing as in [Python](https://github.com/googleapis/python-bigtable/pull/924/files#diff-142d623e0503afa707461f3bd80d8808a2e10232038c24d7d69ce1df6bd4dbceR180) even though a prefix isn't here. The prefix is just included when we define the open telemetry instruments.

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+    });
+    this.otelMetrics?.operationLatencies.record(data.operationLatency, {
+      appProfileId: data.metricsCollectorData.appProfileId,
+      methodName: data.metricsCollectorData.methodName,
+      clientUid: data.metricsCollectorData.clientUid,
+      finalOperationStatus: data.finalOperationStatus,
+      streamingOperation: data.streamingOperation,
+      clientName: data.clientName,
+    });
+    this.otelMetrics?.retryCount.add(data.retryCount, {
```
danieljbruce, line 259: Do we only want to log completed attempts if there were retries like in here?

https://github.com/googleapis/python-bigtable/pull/924/files#diff-291971b83bd5a1e863f09438da490a750a39a632869adff64cd1018dac7532c6R181

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+      appProfileId: data.metricsCollectorData.appProfileId,
+      methodName: data.metricsCollectorData.methodName,
+      clientUid: data.metricsCollectorData.clientUid,
+      finalOperationStatus: data.finalOperationStatus,
+      clientName: data.clientName,
+    });
+    this.otelMetrics?.firstResponseLatencies.record(data.firstResponseLatency, {
```
danieljbruce, line 266: Should we only do this for ReadRows like in Python?

https://github.com/googleapis/python-bigtable/pull/924/files#diff-291971b83bd5a1e863f09438da490a750a39a632869adff64cd1018dac7532c6R220

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+    });
+    this.otelMetrics?.attemptLatencies.record(data.attemptLatency, {
+      appProfileId: data.metricsCollectorData.appProfileId,
+      methodName: data.metricsCollectorData.methodName,
+      clientUid: data.metricsCollectorData.clientUid,
+      attemptStatus: data.attemptStatus,
+      streamingOperation: data.streamingOperation,
+      clientName: data.clientName,
+    });
+    this.otelMetrics?.connectivityErrorCount.add(data.connectivityErrorCount, {
```
danieljbruce, line 297: In Node we actually calculate the connectivityErrorCount in the Metrics collector, not in the metrics handler.

This is not done the same way in Python FYI:

https://github.com/googleapis/python-bigtable/pull/924/files#diff-291971b83bd5a1e863f09438da490a750a39a632869adff64cd1018dac7532c6R231

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+          'monitored_resource.table': data.table,
+          'monitored_resource.cluster': data.cluster,
+          'monitored_resource.zone': data.zone,
+        }).merge(new ResourceUtil.GcpDetectorSync().detect()),
+        readers: [
+          // Register the exporter
+          new PeriodicExportingMetricReader({
+            // Export metrics every 10 seconds. 5 seconds is the smallest sample period allowed by
+            // Cloud Monitoring.
+            exportIntervalMillis: 1_000,
```
leaf94, line 146: I assume this is the default (also across all languages). Just double checking : 0

test/metrics-collector/metricsToRequest.ts
```diff
@@ -0,0 +1,38 @@
...
+
+describe('Bigtable/metricsToRequest', () => {
+  it('Converts an otel request to a request ready for the metric service client', () => {
+    const convertedValue = metricsToRequest(expectedOtelExportInput);
+    assert.deepStrictEqual(
+      convertedValue.timeSeries.length,
+      expectedOtelExportConvertedValue.timeSeries.length
+    );
+    for (let index = 0; index < convertedValue.timeSeries.length; index++) {
+      // We need to compare pointwise because mocha truncates to an 8192 character limit.
```
leaf94, line 31: very helpful comment : )

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: Thanks for catching this. Based on what I see it actually looks like we should export every 60 seconds. That is what the Python client does:

https://github.com/googleapis/python-bigtable/pull/924/files#diff-142d623e0503afa707461f3bd80d8808a2e10232038c24d7d69ce1df6bd4dbceR105

The comment should be corrected as well.

src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,462 @@
...
+    metrics: ResourceMetrics,
+    resultCallback: (result: ExportResult) => void
+  ): void {
+    (async () => {
+      try {
+        const request = metricsToRequest(metrics as unknown as ExportInput);
+        await this.monitoringClient.createTimeSeries(
+          request as ICreateTimeSeriesRequest
+        );
+        // {code: 0} is typically the format the callback expects in the super class.
```
daniel-sanche, line 454: This comment kind of makes me nervous. Are there ever times the callback expects something other than {code: 0}?

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+              ? Aggregation.Sum()
+              : new ExplicitBucketHistogramAggregation(latencyBuckets),
+          })
+      );
+      const meterProvider = new MeterProvider({
+        views: viewList,
+        resource: new Resources.Resource({
+          'service.name': 'Cloud Bigtable Table',
+          'cloud.provider': 'gcp',
+          'cloud.platform': 'gce_instance',
```
daniel-sanche, line 132: why are you hard-coding `gce_instance`?

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+          'cloud.resource_manager.project_id': data.projectId,
```
daniel-sanche, line 133: why do you have both `cloud.resource_manager.project_id` and `monitored_resource.project_id`?

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+          'monitored_resource.type': 'bigtable_client_raw',
+          'monitored_resource.project_id': data.projectId,
+          'monitored_resource.instance_id': data.instanceId,
+          'monitored_resource.table': data.table,
+          'monitored_resource.cluster': data.cluster,
+          'monitored_resource.zone': data.zone,
```
daniel-sanche, line 139: I think you should also be recording `client_name` and `client_uid` here

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+        ),
+        retryCount: meter.createCounter(
+          'bigtable.googleapis.com/internal/client/retry_count',
+          {
+            description:
+              'A counter that records the number of attempts that an operation required to complete. Under normal circumstances, this value is empty.',
+          }
+        ),
+        applicationBlockingLatencies: meter.createHistogram(
+          'bigtable.googleapis.com/internal/client/application_blocking_latencies',
```
daniel-sanche, line 183: IIRC this should be `application_latencies` for the internal metric

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+        ),
+        connectivityErrorCount: meter.createCounter(
+          'bigtable.googleapis.com/internal/client/connectivity_error_count',
+          {
+            description:
+              "The number of requests that failed to reach Google's network. In normal cases, this number is 0. When the number is not 0, it can indicate connectivity issues between the application and the Google network.",
+          }
+        ),
+        clientBlockingLatencies: meter.createHistogram(
+          'bigtable.googleapis.com/internal/client/client_blocking_latencies',
```
daniel-sanche, line 224: I think this should be throttling_latencies

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+   * This method records the operation latency and retry count, associating them with provided attributes.
+   * @param {OnOperationCompleteData} data Data related to the completed operation.
+   */
+  onOperationComplete(data: OnOperationCompleteData) {
+    this.initialize({
+      projectId: data.projectId,
+      instanceId: data.metricsCollectorData.instanceId,
+      table: data.metricsCollectorData.table,
+      cluster: data.metricsCollectorData.cluster,
+      zone: data.metricsCollectorData.zone,
```
daniel-sanche, line 249: Just to be sure: is the system designed so that each Table gets its own instrumentation stack? (That is, is it safe to hard-code the instance and table, since the library can have multiple?)

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+      streamingOperation: data.streamingOperation,
+      clientName: data.clientName,
+    });
+    this.otelMetrics?.retryCount.add(data.retryCount, {
+      appProfileId: data.metricsCollectorData.appProfileId,
+      methodName: data.metricsCollectorData.methodName,
+      clientUid: data.metricsCollectorData.clientUid,
+      finalOperationStatus: data.finalOperationStatus,
+      clientName: data.clientName,
+    });
```
daniel-sanche, line 265: most of these labels are repeated. Can you use spread syntax or something similar to define them in one place?

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+      cluster: data.metricsCollectorData.cluster,
+      zone: data.metricsCollectorData.zone,
+    });
+    this.otelMetrics?.operationLatencies.record(data.operationLatency, {
+      appProfileId: data.metricsCollectorData.appProfileId,
+      methodName: data.metricsCollectorData.methodName,
+      clientUid: data.metricsCollectorData.clientUid,
+      finalOperationStatus: data.finalOperationStatus,
+      streamingOperation: data.streamingOperation,
+      clientName: data.clientName,
```
daniel-sanche, line 257: Shouldn't appProfileId, clientUid, and clientName be the same for all operations? If so, can't they be handle in `initialize`?

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+   */
+  onAttemptComplete(data: OnAttemptCompleteData) {
+    this.initialize({
+      projectId: data.projectId,
+      instanceId: data.metricsCollectorData.instanceId,
+      table: data.metricsCollectorData.table,
+      cluster: data.metricsCollectorData.cluster,
+      zone: data.metricsCollectorData.zone,
+    });
+    this.otelMetrics?.attemptLatencies.record(data.attemptLatency, {
```
daniel-sanche, line 289: is there a way to get rid of these optionals? It would be good if we could have some more guarantees here.

Maybe turn `initialize` into `getMetrics`, and have it return the metrics object, creating it if needed?

daniel-sanche: Also, nit: I think Metrics is the wrong terminology for these. I think these are [Instruments](https://opentelemetry.io/docs/specs/otel/metrics/#programming-model)

src/client-side-metrics/gcp-metrics-handler.ts
daniel-sanche: That seems fine to me. My only concern is this leaves open the possibility of a bug causing it to record connectivity_error_count values > 1, which I don't think should be possible. But I don't think that detail matters too much

I think this field is only supposed to be recorded if it's not 0, but you can double check on that

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,313 @@
...
+      clientName: data.clientName,
+    });
+    this.otelMetrics?.connectivityErrorCount.add(data.connectivityErrorCount, {
+      appProfileId: data.metricsCollectorData.appProfileId,
+      methodName: data.metricsCollectorData.methodName,
+      clientUid: data.metricsCollectorData.clientUid,
+      attemptStatus: data.attemptStatus,
+      clientName: data.clientName,
+    });
+    this.otelMetrics?.serverLatencies.record(data.serverLatency, {
```
daniel-sanche, line 304: what happens if serverLatency is null?

daniel-sanche (changes requested):

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: Ok. I'll check with Andre.

@andre-sampaio Can you comment on whether connectivityErrorCount should be recorded if it is not 0?

src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,462 @@
...
+      exportArgs.resource._syncAttributes['monitored_resource.instance_id'],
+    project_id:
+      exportArgs.resource._syncAttributes['monitored_resource.project_id'],
+    table: exportArgs.resource._syncAttributes['monitored_resource.table'],
+    zone: exportArgs.resource._syncAttributes['monitored_resource.zone'],
+  };
+  for (const scopeMetrics of exportArgs.scopeMetrics) {
+    for (const metric of scopeMetrics.metrics) {
+      const metricName = metric.descriptor.name;
+      if (isCounterMetric(metric)) {
```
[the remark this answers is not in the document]
danieljbruce: > It looks like there's a lot of duplicate code here

Yes. This could be much better. I pulled the `for` outside of the `if` which allows me to eliminate a lot of the duplicate code. This involved using a new type guard for the metric value so a lot of the code here is much more simplified now.

src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,462 @@
...
+              type: metricName,
+              labels: metricLabels,
+            },
+            resource: {
+              type: exportArgs.resource._syncAttributes[
+                'monitored_resource.type'
+              ],
+              labels: resourceLabels,
+            },
+            valueType: 'INT64',
```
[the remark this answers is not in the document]
danieljbruce: Note that `INT64` is only used for `retry_count` and `connectivity_error_count`. For all other metrics the 'DISTRIBUTION' value type is used.

src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,462 @@
...
+          );
+          const timeSeries = {
+            metric: {
+              type: metricName,
+              labels: metricLabels,
+            },
+            resource: {
+              type: exportArgs.resource._syncAttributes[
+                'monitored_resource.type'
+              ],
```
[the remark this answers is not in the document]
danieljbruce: > Shouldn't it always be bigtable_client_raw?

You're right. In practice it is always `bigtable_client_raw` hardcoded [here](https://github.com/googleapis/nodejs-bigtable/pull/1569/files#diff-0613c4be6e08c6da67824291f9be7e4d391e723f88c1417673066fb91077890aR134) and right now the gcp metrics handler always feeds into this exporter. Technically this is more extensible in case we had a different exporter one day, but it is simpler if we just hardcode it here too.

What do you think?

src/client-side-metrics/exporter.ts
danieljbruce: I tested this a while ago and when OpenTelemetry's MetricExporter is successful it provides the callback with this result:

image

And with this code if the exporter fails then we feed the callback with an error that looks like this:

image

And when the OpenTelemetry's MetricExporter fails for some reason then the value provided to the callback looks like this:

image

I think the point I was trying to make was that the result passed to the callback has a code which in this PR is mainly used to verify that the exporter worked properly.

Now, for this discussion I think it's really important to talk about how the callback gets used. It looks as though the meter provider with the periodic exporter is printing errors to the console when the callback receives errors. This won't crash their program, but it will flood their console with messages (That's how the OTEL metrics provider appears to use the exporter). However, the system tests in the next PR will fail if the exporter is failing to write the metrics to Google Cloud Monitoring (ie. error code not equal to 0).

Do you think it is good enough that we catch this behaviour with a good system test?

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: I think this was copy pasted from a Node sample for working with open telemetry.

The main thing here is that this map passed into `Resource` is a ResourceAttributes types which is defined to be just a map of strings to values. It is just used as a mechanism to send persistent data to the exporter which happens to include `projectId`, `instanceId`, `table`, `cluster` and `zone` so for that matter, we can eliminate anything we don't need.

image

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: We can eliminate `'cloud.resource_manager.project_id': data.projectId,` in this case.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: Done. Note that the metrics fixture doesn't have `throttling_latencies` yet. We intend to add this at a later stage and when we do so any tests are going to flush out these naming errors.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: > Just to be sure: is the system designed so that each Table gets its own instrumentation stack?

Yes. That's the idea. And admittedly, to make it work this way we have to do something slightly different from the design doc in that each table or perhaps even each [method call](https://github.com/googleapis/nodejs-bigtable/pull/1571/files#diff-b4e03e7f207567c6df97e62d26cb9a8ccb4f410f3901683d7e399ee944c8ba32R349) will get its own metrics handler because there is a one to one relationship between the metric handler and the instruments.

These parameters define the resource which remains the same across the lifecycle of the instrument. Since the resource contains table information then for each table we need a unique resource and therefore a unique instrument set.

In theory, we could allow one metrics handler per client if the metrics handler stored a map of tables to instrument sets instead, but that might make the metrics handler more complicated then it needs to be right now. There is currently no requirement to allow the user to provide a custom metrics handler so I think a metrics handler per table will be fine.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: > Shouldn't appProfileId, clientUid, and clientName be the same for all operations?

Yes, but in this case I chose to group them with attributes to be consistent with [requests in Golang](https://paste.googleplex.com/5938076009103360). Either way will work and either way has its tradeoffs. What do you think?

danieljbruce: > If so, can't they be handle in initialize

Yes, we could pass them into initialize and store them on the resource in this case.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: Yup. Used spread syntax. Done.

src/client-side-metrics/exporter.ts
danieljbruce: > I can't remember if you planned to support that or not

Not right now. [there isn't a need to support that right now](https://docs.google.com/document/d/1n_D1RBCYa44QTgtThEYgIQ3vLO0YuGeK8sAsilCwhIE/edit?disco=AAABc_QMooA) so I think we should leave that out to optimize for feature delivery time.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: > Maybe turn initialize into getMetrics, and have it return the metrics object, creating it if needed?

Yes. I like this. Done.

src/client-side-metrics/exporter.ts
danieljbruce: We don't need to support this at this time as per the discussion [here](https://docs.google.com/document/d/1n_D1RBCYa44QTgtThEYgIQ3vLO0YuGeK8sAsilCwhIE/edit?disco=AAABc_QMooA).

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: I looked into this and if server latency is null then the metric simply doesn't make it to the export function.

danieljbruce:
@andre-sampaio For the comment at https://github.com/googleapis/nodejs-bigtable/pull/1569#discussion_r1982385150 could you please provide us with information formally on when we should record the connectivity error count?

danieljbruce:
@andre-sampaio @daniel-sanche Tests are failing in CI now due to the Node upgrade, but they are passing locally so this is ready for review again.

danieljbruce:
For https://github.com/googleapis/nodejs-bigtable/pull/1569#discussion_r1982351173: I asked the Bigtable team and we don't need to support this yet. Let's prioritize getting the metrics working for ReadRows and we can go from there.

src/client-side-metrics/gcp-metrics-handler.ts
daniel-sanche: We talked about this offline, but my main worry is that multiple tables could cause duplicate exporters/meters, and we could end up duplicating hte data set to the backend. We'll have to make sure there is some kind of isolation between tables here

danieljbruce: Just adding a note here might help. One thing we discussed we wanted to be sure of is that the design doesn't allow for duplicated data. In the current plan, each table will have a gcp metrics handler and each gcp metrics handler will have an exporter.

src/client-side-metrics/exporter.ts
danieljbruce: As discussed offline we expect the code to be 0 when the operation sending the metrics to GCP monitoring is successful and non-zero otherwise.

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,298 @@
...
+   * Creates and registers metric instruments (histograms and counters) for various Bigtable client metrics.
+   * Sets up a MeterProvider and configures a PeriodicExportingMetricReader for exporting metrics to Cloud Monitoring.
+   *
+   * @param {MonitoredResourceData} data The data that will be used to set up the monitored resource
+   * which will be provided to the exporter in every export call.
+   *
+   */
+  private getMetrics(data: MonitoredResourceData): Metrics {
+    if (!this.otelInstruments) {
+      const latencyBuckets = [
```
andre-sampaio, line 101: have we defined the latency buckets somewhere? I'm trying to find a reference. These are different than the ones we use in the java client: https://github.com/googleapis/java-bigtable/blob/966f46d54d8d7ccd7fd8a5dcb6bb6fcf9ace14e7/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsConstants.java#L67-L73

src/instance.ts
```diff
@@ -1310,7 +1310,7 @@ Please use the format 'my-instance' or '${bigtable.projectName}/instances/my-ins
         : callback!;
 
     if (policy.etag !== null && policy.etag !== undefined) {
-      (policy.etag as {} as Buffer) = Buffer.from(policy.etag);
+      (policy.etag as {} as Buffer) = Buffer.from(policy.etag as string);
```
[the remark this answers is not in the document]
danieljbruce: Yeah. With the Node 18 upgrade the compiler began complaining so a cast is actually required here.

src/table.ts
```diff
@@ -1042,7 +1042,7 @@ export class Table extends TabularApiSurface {
         : callback!;
 
     if (policy.etag !== null && policy.etag !== undefined) {
-      (policy.etag as {} as Buffer) = Buffer.from(policy.etag);
+      (policy.etag as {} as Buffer) = Buffer.from(policy.etag as string);
```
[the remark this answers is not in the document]
danieljbruce: Yeah. With the Node 18 upgrade the compiler began complaining so a cast is actually required here.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: Good callout. I modelled this after the Python client [here](https://github.com/googleapis/python-bigtable/pull/924/files#diff-142d623e0503afa707461f3bd80d8808a2e10232038c24d7d69ce1df6bd4dbceR56), but must have missed Mattie's comment below. I'll make this update now.

src/client-side-metrics/exporter.ts
danieljbruce: Comment added

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: Done

src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,409 @@
...
+              (
+                allAttributes as OnOperationAttribute
+              ).finalOperationStatus?.toString(),
+            client_uid: allAttributes.clientUid,
+          },
+          streaming ? {streaming} : null
+        );
+        const metric = {
+          type: metricName,
+          labels: metricLabels,
```
daniel-sanche, line 276: Maybe I'm misunderstanding node, but is converting this to a new metricLabels object necessary? Shouldn't it be equivalent to datapoint.attributes? (assuming the status/streaming attributes were written in the proper format to begin with)

src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,462 @@
...
+ *
+ */
+type CounterMetric = Metric<OnAttemptAttribute | OnOperationAttribute, number>;
+
+/**
+ * Represents the input data structure for exporting OpenTelemetry metrics.
+ *
+ * This interface defines the structure of the object that is passed to the
+ * `metricsToRequest` function to convert OpenTelemetry metrics into a format
+ * suitable for the Google Cloud Monitoring API.
```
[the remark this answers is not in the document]
daniel-sanche: Can you explain the kinds of errors you're seeing? I see the red in the screenshots, but I'm not sure what the cause is

This is feeling like a bit of a workaround to me. We should try to use the types supplied by OTel if we can. If _syncAttributes is not advertised, maybe it's meant to be internal

src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,409 @@
...
+export function metricsToRequest(exportArgs: ExportInput) {
+  const timeSeriesArray = [];
+  const resourceLabels = {
+    cluster: exportArgs.resource._syncAttributes['monitored_resource.cluster'],
+    instance:
+      exportArgs.resource._syncAttributes['monitored_resource.instance_id'],
+    project_id:
+      exportArgs.resource._syncAttributes['monitored_resource.project_id'],
+    table: exportArgs.resource._syncAttributes['monitored_resource.table'],
+    zone: exportArgs.resource._syncAttributes['monitored_resource.zone'],
```
daniel-sanche, line 245: I mentioned above, but seeing `._syncAttributes` seems a little sketchy to me. That doesn't seem to be part of the advertised API surface

src/client-side-metrics/gcp-metrics-handler.ts
daniel-sanche: I see. I think I just stored all monitored_resource fields as data_point attributes, and ignored the OTel.Resource layer. I'm fine with this as-is if it works too though (although I do see _syncAttributes is giving you some trouble)

src/client-side-metrics/exporter.ts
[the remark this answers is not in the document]
daniel-sanche: I think I understand what you're going for a bit better now. Personally, since the field is only relevant for the exporter, I think it makes sense to keep it here. Keep all the GCP Cloud Monitoring data in one place, all the OTel configuration in another, etc, instead of mixing them all together

But keeping all the Monitored Resource data together as an OTel Resource[ like you're doing](https://github.com/googleapis/nodejs-bigtable/pull/1569/files#diff-0613c4be6e08c6da67824291f9be7e4d391e723f88c1417673066fb91077890aR134) also makes sense, so I'm fine with it as-is

src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,409 @@
...
+        await this.monitoringClient.createTimeSeries(
+          request as ICreateTimeSeriesRequest
+        );
+        // The resultCallback typically accepts a value equal to {code: x}
+        // for some value x along with other info. When the code is equal to 0
+        // then the operation completed successfully. When the code is not equal
+        // to 0 then the operation failed. Open telemetry logs errors to the
+        // console when the resultCallback passes in non-zero code values and
+        // logs nothing when the code is 0.
+        const exportResult = {code: 0};
```
daniel-sanche, line 402: can you move this constant to the top of the file?

src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,409 @@
...
+                          bounds: value.buckets.boundaries,
+                        },
+                      },
+                      bucketCounts: value.buckets.counts.map(String),
+                    },
+                  },
+                },
+              ],
+              unit: scopeMetric.descriptor.unit || 'ms', // Default to 'ms' if no unit is specified
+            };
```
daniel-sanche, line 327: Can this be turned into an if? I think long ternary statements like this can be hard to follow

src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,409 @@
...
+        const metricLabels = Object.assign(
+          {
+            app_profile: allAttributes.appProfileId,
+            client_name: allAttributes.clientName,
+            method: allAttributes.methodName,
+            status:
+              (allAttributes as OnAttemptAttribute).attemptStatus?.toString() ??
+              (
+                allAttributes as OnOperationAttribute
+              ).finalOperationStatus?.toString(),
```
daniel-sanche, line 269: instead of doing toString here, can't you write it as a string to start with?

daniel-sanche (changes requested):

src/client-side-metrics/exporter.ts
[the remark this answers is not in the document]
danieljbruce: The image below shows the error message I get if I change the type for the argument to `ResourceMetrics` from `ExportInput`. As you can see the compiler says exportArgs.resource doesn't have the property `_syncAttributes`.

image

But when we pass the argument from the `export` function in here then `exportArgs.resource` actually has a `_syncAttributes` property even though the interface doesn't include this property!

image

Here is another screenshot so you can see exactly what will get passed into export:

image

I could try to make this work an eliminate `ExportInput` by not using the `_syncAttributes` at all, but I don't know what the side effects of that will be. I'll try it and get back to you.

I took another look at the `buckets` error and it looks like we can deal with it by using a cast so I'm not worried about that anymore:
ie. (value as Histogram)

danieljbruce:
For threads like [this](https://github.com/googleapis/nodejs-bigtable/pull/1569#discussion_r1994371184) it is not letting me comment in the conversation tab

src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,462 @@
...
+ *   (e.g., 'nodejs-bigtable', 'go-bigtable/1.35.0').
+ */
+interface OnOperationAttribute {
+  methodName: string;
+  clientUid: string;
+  appProfileId?: string;
+  finalOperationStatus: number;
+  streamingOperation?: string;
+  clientName: string;
+}
```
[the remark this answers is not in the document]
danieljbruce: By eliminating `ExportInput` I was able to eliminate the need for `OnOperationAttribute` and `OnAttemptAttribute` along with a number of interfaces the `ExportInput` was counting on.

src/client-side-metrics/exporter.ts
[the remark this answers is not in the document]
danieljbruce: So I eliminated `ExportInput` and used some careful casting to deal with the `_syncAttributes` error. This eliminates a whole dependency chain of interfaces we don't need anymore.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: With some careful casting I got rid of the `_syncAttributes` error and still managed to get rid of that `ExportInput` interface so that problem is gone.

src/client-side-metrics/exporter.ts
danieljbruce: It's only used in one place. I just eliminated it and replaced it inline on the next line of code.

src/client-side-metrics/exporter.ts
danieljbruce: I used ternary here because it's more _functional_, but I see your point on readability. It's back to an if statement now.

src/client-side-metrics/exporter.ts
danieljbruce: `_syncAttributes` aren't part of the API surface, but they actually exist when the `export` method is called as shown in the screenshot in one of my other comments.

We could be grouping these under the data point attributes instead, but I think in other comments we discussed either way will work?

src/client-side-metrics/exporter.ts
danieljbruce: Yes. I pushed the string conversion back to the metrics collector [here](https://github.com/googleapis/nodejs-bigtable/pull/1569/commits/a587d0c97b83b061b20016cf5edb062049284d10) and use status now instead of attemptStatus and finalOperationStatus.

src/client-side-metrics/exporter.ts
danieljbruce: In this case it was necessary because the property names upstream were slightly different from the property names here like `client_uid` versus `clientUid`. So I changed the property names upstream so we can just use `datapoint.attributes` because to your point, I think it's nicer if we don't have to do any conversions here.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: The [test run](https://btx.cloud.google.com/invocations/32fcab0c-ab93-4baa-b678-c88052a68fc2/targets/cloud-devrel%2Fclient-libraries%2Fnodejs%2Fpresubmit%2Fgoogleapis%2Fnodejs-bigtable%2Fnode14%2Fsystem-test/log) succeeds, but I want to revisit your idea which involves storing all the dimensions on the OTEL instrument instead of the resource object because I think it's a more loosely coupled design.

src/client-side-metrics/exporter.ts
danieljbruce: I actually ended up deciding to group these under data point attributes mainly because then it doesn't tightly couple the metrics handler to a table.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: I changed the design so that we can store all dimensions except project on the OTEL attributes. The project still needs to provided per Exporter because it is used in the metric name.

src/client-side-metrics/exporter.ts
danieljbruce: After including more attributes in datapoint.attributes like table, cluster and zone this is now a subset of `datapoint.attributes` so I had to add more code here.

src/client-side-metrics/exporter.ts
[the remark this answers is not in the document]
danieljbruce: I actually ended up going your way on this because I didn't like the design pattern of using the Resource to store export data in retrospect. I think the extra complication is unnecessary.

src/client-side-metrics/gcp-metrics-handler.ts
andre-sampaio: I do think it should be recorded even being 0. I'm trying to find some authoritative source for this but I couldn't find any, but this would be consistent with the [java implementation](https://github.com/googleapis/java-bigtable/blob/f9684290bc97ccebf297711aa85e662b5560c4f9/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsTracer.java#L426-L431). 

An eventual bug that leads the value to be higher than 1 does slightly concern me though, since it could throw us of during an investigation.

danieljbruce: Ok. I just pushed an update to set connectivityErrorCount to 1 instead of incrementing it.

src/client-side-metrics/gcp-metrics-handler.ts
daniel-sanche: Awesome, thanks!

daniel-sanche (commented):
This is looking very solid! My only remaining request would be to rename the Metrics interface to `Instruments` or `MetricInstruments`, since [Metrics has a different meaning in OTel](https://opentelemetry.io/docs/concepts/signals/metrics/)

After that, this should be good to merge

src/client-side-metrics/exporter.ts
```diff
@@ -0,0 +1,250 @@
...
+            valueType: 'INT64',
+            points: [
+              {
+                interval,
+                value: {
+                  int64Value: dataPoint.value,
+                },
+              },
+            ],
+          });
```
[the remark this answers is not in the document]
danieljbruce: The tests that export results to Google Cloud Monitoring succeed at exporting the metrics so I don't think so. If we did have a unit for `connectivity_error_count` and `retry_count` then what would it be? These are just counters.

src/client-side-metrics/exporter.ts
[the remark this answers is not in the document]
danieljbruce: Yeah. I like breaking this stuff down because to the code reader it makes the variable interelationship clear. In particular, helper functions let us know which variables aren't related which makes it simpler to look at.

I added four helpers:
getMetric
getResource
getIntegerPoints
getDistributionPoints

So this function is much shorter now.

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,272 @@
...
+  MeterProvider,
+  Histogram,
+  PeriodicExportingMetricReader,
+} = require('@opentelemetry/sdk-metrics');
+
+/**
+ * A collection of OpenTelemetry metric instruments used to record
+ * Bigtable client-side metrics.
+ */
+interface Metrics {
```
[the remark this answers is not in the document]
danieljbruce: Done

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -0,0 +1,272 @@
...
+
+  /**
+   * Initializes the OpenTelemetry metrics instruments if they haven't been already.
+   * Creates and registers metric instruments (histograms and counters) for various Bigtable client metrics.
+   * Sets up a MeterProvider and configures a PeriodicExportingMetricReader for exporting metrics to Cloud Monitoring.
+   *
+   * which will be provided to the exporter in every export call.
+   *
+   */
+  private getMetrics(projectId: string): Metrics {
```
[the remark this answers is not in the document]
danieljbruce: Yup. Renamed to getInstruments.

danieljbruce:
Re: https://github.com/googleapis/nodejs-bigtable/pull/1569#discussion_r2006465172. I think this is the MetricsExporter file? We can change `{code: 0}` elsewhere in `src` if you see it in another place, but I can't find one. `{code: 0}` appears in a lot of tests though.

daniel-sanche (approved):
LGTM!

PR #1571, opened 2025-03-06
feat: Add client side metrics collection to readRows calls

danieljbruce:
**Summary:**

This PR will collect client side metrics for readRows calls.

**Background:**

The following PRs have been added to support the plumbing for client-side metrics.
https://github.com/googleapis/nodejs-bigtable/pull/1569
https://github.com/googleapis/nodejs-bigtable/pull/1566

They didn't change the client behaviour at all. This PR starts using this plumbing for readRows calls.

The diagram below shows the CSM architecture with new pieces in green in the diagram:

CSM With fixture - 3rd PR - ReadRows calls (1)

**Changes:**

`src/client-side-metrics/operation-metrics-collector.ts:` Added optimizations and added guards to make sure that if the data from the server is of some unexpected form that it won't crash our clients.
`src/index.ts:` Add code to generate a uuid for each client and store the GCP metrics handler on the client. Also add piping to pass status events along and add a setting that lets users turn off client side metrics.
`src/tabular-api-surface.ts`: Use the metrics collector with the gcp metrics handler to collect client side metrics in readRows calls
`system-test/client-side-metrics-setup-table.ts:` A test utility function for setting up an instance with a few tables
`system-test/client-side-metrics.ts:` A file with some tests for capturing various scenarios where a user might be collecting client side metrics that ensures the right metrics are exported without error even when multiple tables, instances or clients are involved.
`system-test/read-rows.ts:` This test has a mock that is client side metrics incompatible so don't use client side metrics in this test.

**Next Steps:**

- Add plumbing for application blocking latencies
- Collect client side metrics for other grpc endpoints


danieljbruce:
We still need to do the QPS check so I'll come back here and comment when that is done.

src/index.ts
```diff
@@ -417,9 +424,19 @@ export class Bigtable {
   appProfileId?: string;
   projectName: string;
   shouldReplaceProjectIdToken: boolean;
+  clientUid = crypto.randomUUID();
```
daniel-sanche, line 427: I'm not seeing where this is used, but IIRC the client_uid is supposed to follow a specific structure: https://github.com/googleapis/python-bigtable/blob/987e3df4ce64a6079e03295f653363f1cc391e0b/google/cloud/bigtable/data/_metrics/handlers/opentelemetry.py#L144

src/index.ts
```diff
@@ -517,6 +534,16 @@ export class Bigtable {
...
     this.projectName = `projects/${this.projectId}`;
     this.shouldReplaceProjectIdToken = this.projectId === '{{projectId}}';
+
+    if (options.collectMetrics === false) {
+      this.collectMetrics = false;
+      this.metricsHandlers = [];
+    } else {
+      this.collectMetrics = true;
+      this.metricsHandlers = [
+        new GCPMetricsHandler(new CloudMonitoringExporter()),
```
daniel-sanche, line 544: is there a reason to have `CloudMonitoringExporter` as an argument? Will a different one ever be passed in?

src/index.ts
```diff
@@ -417,9 +424,19 @@ export class Bigtable {
...
   static Cluster: Cluster;
+  // Each time a metrics handler is created it introduces significant latency.
+  // Therefore, metrics handlers should be created at the client level and
+  // reused throughout the library to reduce latency:
+  metricsHandlers: IMetricsHandler[];
+  // collectMetrics is a member variable that is used to ensure that if the
+  // user provides a `false` value and opts out of metrics collection that
+  // the metrics collector is ignored altogether to reduce latency in the
+  // client.
+  collectMetrics: boolean;
```
daniel-sanche, line 439: nit: Using a verb for a flag can be a bit confusing, it sounds more like a function name. I'd suggest something like `metricsEnabled`

src/tabular-api-surface.ts
```diff
@@ -507,6 +524,24 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
         return false;
       };
 
+      if (this.bigtable.collectMetrics) {
```
daniel-sanche, line 527: Could we have it do the callbacks every time, regardless of this variable?

We should assume almost all users have CSM enabled, since that's the default. And we'd be introducing a lot of extra branching if we check the metrics collection state every time

src/tabular-api-surface.ts
```diff
@@ -532,6 +567,10 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
               numConsecutiveErrors,
               backOffSettings
             );
+            metricsCollector?.onAttemptComplete(
```
daniel-sanche, line 570: I see a `return` on line 553. Don't you need to finalize things there?

src/tabular-api-surface.ts
```diff
@@ -547,16 +586,33 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
               //
               error.code = grpc.status.CANCELLED;
             }
+            metricsCollector?.onAttemptComplete(
```
daniel-sanche, line 589: I REALLY don't like the comment above this: 

> According to Gemini, the appropriate code for a closed client is CANCELLED since the user actually cancelled the call by closing the client.

You're responsible for the code, not Gemini. What do you think is the appropriate code for a closed client?

src/tabular-api-surface.ts
```diff
@@ -547,16 +586,33 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
           // time won't keep increasing when as stream had multiple errors
           numConsecutiveErrors = 0;
+          metricsCollector?.onResponse(this.bigtable.projectId);
         })
         .on('end', () => {
           activeRequestStream = null;
+          metricsCollector?.onAttemptComplete(
+            this.bigtable.projectId,
+            grpc.status.OK
+          );
```
daniel-sanche, line 611: FWIW, in Python I created a separate `onOperationSucceeded` helper to call onAttemptComplete and onOpeationComplete with grpc.status.OK. This makes it a bit easier to see the happy-path, and can simplify things a bit. Up to you if thats helpful here

src/tabular-api-surface.ts
daniel-sanche: Also, I don't like the idea of referencing a specific test in the code here. We don't just want the test to pass, we want to make sure the underlying requirements are being met

daniel-sanche (changes requested):

src/index.ts
andre-sampaio: Reference in java as well: https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java#L113-L134

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -17,10 +17,23 @@ import {IMetricsHandler} from './metrics-handler';
 import {MethodName, StreamingState} from './client-side-metrics-attributes';
 import {grpc} from 'google-gax';
 import * as gax from 'google-gax';
-const root = gax.protobuf.loadSync(
-  './protos/google/bigtable/v2/response_params.proto'
-);
-const ResponseParams = root.lookupType('ResponseParams');
+
+let ResponseParams: gax.protobuf.Type | null;
+try {
```
andre-sampaio, line 22: I'm a bit concerned with failing silently here due to a test failing. If the tests that are failing are only the tests for the samples I'd rather skip those tests in the CI until we figure out what to do

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -139,75 +156,88 @@ export class OperationMetricsCollector {
...
+        this.state ===
+          MetricsCollectorState.OPERATION_STARTED_ATTEMPT_IN_PROGRESS_NO_ROWS_YET ||
+        this.state ===
+          MetricsCollectorState.OPERATION_STARTED_ATTEMPT_IN_PROGRESS_SOME_ROWS_RECEIVED
+      ) {
+        this.state =
+          MetricsCollectorState.OPERATION_STARTED_ATTEMPT_NOT_IN_PROGRESS;
+        this.attemptCount++;
+        const endTime = new Date();
+        if (projectId && this.attemptStartTime) {
```
andre-sampaio, line 170: We expect this to always be true right? If so can we log an error if this happens?

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -139,75 +156,88 @@ export class OperationMetricsCollector {
...
+    try {
+      if (!this.firstResponseLatency) {
+        if (
+          this.state ===
+          MetricsCollectorState.OPERATION_STARTED_ATTEMPT_IN_PROGRESS_NO_ROWS_YET
+        ) {
+          this.state =
+            MetricsCollectorState.OPERATION_STARTED_ATTEMPT_IN_PROGRESS_SOME_ROWS_RECEIVED;
+          const endTime = new Date();
+          if (projectId && this.operationStartTime) {
```
andre-sampaio, line 233: We expect this to always be true right? If so can we log an error if this happens? Same note for other similar checks where we expect the condition to be always be true

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -218,33 +248,40 @@ export class OperationMetricsCollector {
...
-                client_name: `nodejs-bigtable/${version}`,
-                projectId,
-                operationLatency: totalTime,
-                retryCount: this.attemptCount - 1,
-                firstResponseLatency: this.firstResponseLatency ?? undefined,
-              });
-            }
-          });
+    try {
+      if (
```
andre-sampaio, line 252: Can we log an error if this is false?

src/tabular-api-surface.ts
andre-sampaio, line 589: Instead of calling on attempt complete in both the `if` and `else` conditions we can just move it out of the block

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -17,10 +17,23 @@ import {IMetricsHandler} from './metrics-handler';
...
+
+let ResponseParams: gax.protobuf.Type | null;
+try {
+  /*
+   * Likely due to the Node 18 upgrade, the samples tests are failing with the
+   * error UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or
+   * directory, open 'protos/google/bigtable/v2/response_params.proto'. Since
+   * these tests don't use this module we can suppress the error for now to
+   * unblock the CI pipeline.
+   */
```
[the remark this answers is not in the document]
danieljbruce: [here is a log](https://btx.cloud.google.com/invocations/d22777c4-0b03-4939-a47d-a9a3762c7691/targets/cloud-devrel%2Fclient-libraries%2Fnodejs%2Fpresubmit%2Fgoogleapis%2Fnodejs-bigtable%2Fnode14%2Fsamples-test;config=default/log) that shows what happens without this change. Not a problem with a specific test, just an error in the 'fs' module all of a sudden. It might go away when we start running these tests with a higher version of Node.

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -123,13 +136,17 @@ export class OperationMetricsCollector {
...
   onOperationStart() {
-    if (this.state === MetricsCollectorState.OPERATION_NOT_STARTED) {
-      this.operationStartTime = new Date();
-      this.firstResponseLatency = null;
-      this.state =
-        MetricsCollectorState.OPERATION_STARTED_ATTEMPT_NOT_IN_PROGRESS;
-    } else {
-      console.warn('Invalid state transition');
+    try {
+      if (this.state === MetricsCollectorState.OPERATION_NOT_STARTED) {
```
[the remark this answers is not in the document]
danieljbruce: Okay. This is now handled with a wrapper `checkState` function.

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -139,75 +156,88 @@ export class OperationMetricsCollector {
...
+        this.state =
+          MetricsCollectorState.OPERATION_STARTED_ATTEMPT_NOT_IN_PROGRESS;
+        this.attemptCount++;
+        const endTime = new Date();
+        if (projectId && this.attemptStartTime) {
+          const totalTime = endTime.getTime() - this.attemptStartTime.getTime();
+          this.metricsHandlers.forEach(metricsHandler => {
+            if (metricsHandler.onAttemptComplete) {
+              // attemptStatus?.toString() is optional because in a test proxy
+              // test the server does not send back the status.
```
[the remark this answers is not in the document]
danieljbruce: Ok. I eliminated the optional and the tests seem to work so I'm guess this observation was just a mistake. I changed it back now.

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -218,33 +248,40 @@ export class OperationMetricsCollector {
...
+        this.state = MetricsCollectorState.OPERATION_COMPLETE;
+        const endTime = new Date();
+        if (projectId && this.operationStartTime) {
+          const totalTime =
+            endTime.getTime() - this.operationStartTime.getTime();
+          {
+            this.metricsHandlers.forEach(metricsHandler => {
+              if (metricsHandler.onOperationComplete) {
+                // finalOperationStatus?.toString() is optional because in a test
+                // proxy test the server does not send back the status.
```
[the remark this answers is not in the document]
danieljbruce: Ok. I eliminated the optional and the tests seem to work so I'm guess this observation was just a mistake. I changed it back now.

src/index.ts
```diff
@@ -417,9 +424,19 @@ export class Bigtable {
   appProfileId?: string;
   projectName: string;
   shouldReplaceProjectIdToken: boolean;
+  clientUid = crypto.randomUUID();
   static AppProfile: AppProfile;
   static Instance: Instance;
   static Cluster: Cluster;
+  // Each time a metrics handler is created it introduces significant latency.
```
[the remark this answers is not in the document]
danieljbruce: In our other PR we are considering changing the code to initialize the stack 1 time so let's resolve that discussion and then decide what to do here next.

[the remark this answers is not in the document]
danieljbruce: I did some measurements [here](https://paste.googleplex.com/6658498605350912) on `main-readrows-latency-experiment`. It honestly seems like the latency is negligible compared to the client initialization time, its just that it would add up a lot if we initialized a stack for every table or every call.

src/index.ts
danieljbruce: It gets passed in [here](https://github.com/googleapis/nodejs-bigtable/blob/c37a451c52ff5639fc9f5e76c88e93ba549fe6c9/src/client-side-metrics/operation-metrics-collector.ts#L116).

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: The trouble is that the error is for the entire suite so we would have to skip the entire CI check in order for this to be unblocked.

src/index.ts
danieljbruce: Within the client library code `CloudMonitoringExporter` is always used, but tests sometime pass in a mock exporter so that we can unit test GCPMetricsHandler's behaviour. If you want we can define a DefaultGCPMetricsHandler class that just always uses the CloudMonitoringExporter to eliminate the confusion.

src/tabular-api-surface.ts
danieljbruce: Good catch. Added that fix!

src/tabular-api-surface.ts
danieljbruce: I added a onLastAttemptCompleted method which simplifies things in a few places. It will reduce the amount of code necessary for methods that use the metrics collector.

src/tabular-api-surface.ts
danieljbruce: onAttemptComplete is no longer called in both places after the introduction of the `onLastAttemptCompleted` method.

src/tabular-api-surface.ts
danieljbruce: Sure. I added the check out of concern for latency, but it is likely pretty minor.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: Done. I added an else with a warning

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: Now logging an error if this happens

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: Yup. Done

src/index.ts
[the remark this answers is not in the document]
daniel-sanche: Maybe worth changing this comment then if latency isn't the reason

src/index.ts
daniel-sanche: It looks like [we discussed this in the other PR](https://github.com/googleapis/nodejs-bigtable/pull/1583#discussion_r2034178097). But this one probably makes more sense

Isn't there a way to inject the mock without making the internal variable part of the API surface?

src/client-side-metrics/operation-metrics-collector.ts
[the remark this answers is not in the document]
daniel-sanche: is this still relevant? It looks like that CL was rolled back

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -58,6 +75,33 @@ enum MetricsCollectorState {
...
+// Checks that the state transition is valid and if not it throws a warning.
+function checkState<T>(
+  currentState: MetricsCollectorState,
+  allowedStates: MetricsCollectorState[],
+  fn: () => T
+): T | undefined {
+  if (allowedStates.includes(currentState)) {
+    return fn();
+  } else {
+    console.warn('Invalid state transition');
```
daniel-sanche, line 100: Did you intend to have this always log, or should this also be behind METRICS_DEBUG?

src/tabular-api-surface.ts
```diff
@@ -507,6 +524,22 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
+          }
+        )
+        .on(
+          'status',
+          (status: {
+            metadata: {internalRepr: Map<string, Uint8Array[]>; options: {}};
+          }) => {
+            metricsCollector?.onStatusMetadataReceived(status);
+          }
+        );
```
daniel-sanche, line 541: I imagine you'll need to do something like this for every rpc? Should there be a helper for this?

src/index.ts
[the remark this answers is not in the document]
danieljbruce: Latency would be a reason if we created a stack for every readrows call. Too many calls to the exporter is another reason. I've updated the comment.

src/client-side-metrics/operation-metrics-collector.ts
andre-sampaio: Is it only for the [samples](https://github.com/googleapis/nodejs-bigtable/tree/main/samples/test) test suite though? If it is I would still rather just skip those to be honest

src/tabular-api-surface.ts
```diff
@@ -605,11 +602,7 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
         })
         .on('end', () => {
```
andre-sampaio, line 603: Won't this be called on errors as well?

src/client-side-metrics/operation-metrics-collector.ts
andre-sampaio: If we really don't want to skip can we gate this behind an env variable that is set from the test?

src/index.ts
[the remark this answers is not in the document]
danieljbruce: Actually, we don't need this on the client anymore because the OTEL stacks now belong to a singleton. So I moved the metrics handlers to live as a singleton on the OperationMetricsCollector instead of on the client here.

We shouldn't be exposing a `metricsHandlers` variable on the client if we don't have to. This is a public interface.

src/client-side-metrics/operation-metrics-collector.ts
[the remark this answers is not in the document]
danieljbruce: [the most recent log](https://btx.cloud.google.com/invocations/c3d9117f-67da-4c7c-b522-bb5707a1ce38/targets/cloud-devrel%2Fclient-libraries%2Fnodejs%2Fpresubmit%2Fgoogleapis%2Fnodejs-bigtable%2Fnode14%2Fsamples-test/log) shows this is still an issue so it must be due to a fs upgrade or something. I don't know why this is suddenly cropping up.

src/tabular-api-surface.ts
danieljbruce: Yes. I added `handleStatusAndMetadata` which should help us here.

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -218,34 +280,53 @@ export class OperationMetricsCollector {
...
+  /**
+   * This is a useful helper method for all the times we want to record that an
+   * attempt is complete when an operation is also complete.
+   *
+   * @param {string} projectId The id of the project.
+   * @param {grpc.status} finalOperationStatus Information about the completed operation.
+   */
+  onLastAttemptCompleted(projectId: string, attemptStatus: grpc.status) {
+    this.onAttemptComplete(projectId, attemptStatus);
+    this.onOperationComplete(projectId, attemptStatus);
```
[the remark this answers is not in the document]
danieljbruce: Okay. I eliminated onLastAttemptCompleted.

src/tabular-api-surface.ts
danieljbruce: No. I did an experiment on https://github.com/googleapis/nodejs-bigtable/commit/e9349762dced971590fe07f1752c6fd70ebbf085 and I get the output:

```
  Bigtable/ReadRows
error row stream happened
error reached user
    ✔ should create read stream and read synchronously (220ms)
```

src/index.ts
danieljbruce: In multiple tests a mock exporter is passed in here:

image

We could inject the mock with proxyquire, doing something like https://github.com/googleapis/nodejs-bigtable/pull/1571/commits/1ddaa7df1d142abff4fe399a2da6fa8b1cf00aaf for each test, but when proxyquire is involved it can be hard to debug later. What do you want to do here?

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: Yes. Only for the samples suite.

We can change the npm script to do `cd samples/ && npm link ../ && npm install && npx cross-env METRICS_DEBUG=true npm test && cd ../` instead of `cd samples/ && npm link ../ && npm install && npm test && cd ../` and that might work. Do you want to do this?

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: I ended up gating this with an environment variable and actually throwing an error when the variable is true. We have to change the npm script in package json to use the variable in tests:

ie. npx cross-env METRICS_DEBUG=true mocha build/system-test --timeout 600000

Should we use the variable in tests too?

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -58,6 +72,35 @@ enum MetricsCollectorState {
...
+  currentState: MetricsCollectorState,
+  allowedStates: MetricsCollectorState[],
+  fn: () => T
+): T | undefined {
+  if (allowedStates.includes(currentState)) {
+    return fn();
+  } else {
+    if (METRICS_DEBUG) {
+      throw Error('Invalid state transition');
+    }
```
daniel-sanche, line 99: Since you're throwing errors now, this can be simplified quite a bit:
1. remove the `if (METRICS_DEBUG)` check here, and always throw an error if it's in the wrong state. `withMetricsDebug` will catch it and decide whether to raise. That keeps all the `METRICS_DEBUG` logic in one place
2. You now have the option to remove the whole `fn: () => T` callback part, and just return if it's in a valid state. Execution will just continue if the state is good, or end here if it's in a bad state. (But I know JS is more callback-centric than other languages, so maybe it's more idiomatic to keep it as-is. I find lots of nested callbacks hard to reason about, but maybe that's just me)

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -13,10 +13,24 @@
...
+// with failures in the metrics collector.
+const METRICS_DEBUG = process.env.METRICS_DEBUG;
+
+/*
+ * The samples tests are failing with the
+ * error UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or
+ * directory, open 'protos/google/bigtable/v2/response_params.proto'. Since
+ * these tests don't use this module we can suppress the error for now to
+ * unblock the CI pipeline.
+ */
```
daniel-sanche, line 33: Is this comment still relevant?

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -139,76 +209,92 @@ export class OperationMetricsCollector {
...
+                    status: attemptStatus.toString(),
+                    client_name: `nodejs-bigtable/${version}`,
+                    metricsCollectorData: this.getMetricsCollectorData(),
+                    projectId,
+                  });
+                }
+              }
+            );
+          } else {
+            console.warn('ProjectId and start time should always be provided');
```
daniel-sanche, line 244: should this be a throw?

It seems to me like all metrics errors should be handled with the METRICS_DEBUG logic. But maybe you wanted different levels for exceptions vs warnings?

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -139,76 +209,92 @@ export class OperationMetricsCollector {
...
-      console.warn('Invalid state transition attempted');
-    }
+    withMetricsDebug(() => {
+      checkState(
+        this.state,
+        [MetricsCollectorState.OPERATION_STARTED_ATTEMPT_NOT_IN_PROGRESS],
+        () => {
+          this.state =
+            MetricsCollectorState.OPERATION_STARTED_ATTEMPT_IN_PROGRESS_NO_ROWS_YET;
+          this.attemptStartTime = new Date();
```
daniel-sanche, line 262: I thought I saw a change to `hrtime.bigint` for timestamps. Should that be done here too?

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -139,76 +209,92 @@ export class OperationMetricsCollector {
...
+      if (!this.firstResponseLatency) {
+        checkState(
+          this.state,
+          [
+            MetricsCollectorState.OPERATION_STARTED_ATTEMPT_IN_PROGRESS_NO_ROWS_YET,
+          ],
+          () => {
+            this.state =
+              MetricsCollectorState.OPERATION_STARTED_ATTEMPT_IN_PROGRESS_SOME_ROWS_RECEIVED;
+            const endTime = new Date();
```
daniel-sanche, line 285: same comment about `hrtime.bigint` timestamps

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: > I find lots of nested callbacks hard to reason about, but maybe that's just me

I do too. I think this is a good suggestion.

> withMetricsDebug will catch it and decide whether to raise

This is true, but it does mean we should only use checkState when it is wrapped in withMetricsDebug.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: Yes. I think throwing here keeps the logic simpler so I made this change.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: This change has been cherry picked.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: Unfortunately, this still comes up even after the Node 18 upgrade: https://btx.cloud.google.com/invocations/b88b3b6c-0ace-413b-9b8c-c9bef9eceb03/targets/cloud-devrel%2Fclient-libraries%2Fnodejs%2Fpresubmit%2Fgoogleapis%2Fnodejs-bigtable%2Fnode18%2Fsamples-test/log

danieljbruce: We'll see if using the path module for relative path resolution fixes the problem....

src/client-side-metrics/operation-metrics-collector.ts
daniel-sanche: > This is true, but it does mean we should only use checkState when it is wrapped in withMetricsDebug.

Wasn't that the intention? That you wrap everything that could call in `withMetricsDebug`, and can then surpress exceptions?

src/client-side-metrics/exporter.ts
```diff
@@ -328,7 +342,8 @@ export class CloudMonitoringExporter extends MetricExporter {
             maxRetryDelayMillis: 50000,
           },
         );
-        await this.monitoringClient.createTimeSeries(
+        const monitoringClient = new MetricServiceClient({projectId});
```
daniel-sanche, line 345: we may need to authenticate for this project Id

Since we have a bug tracking this, we can probably address this later

daniel-sanche (approved):

daniel-sanche (changes requested):
It looks like we have to make some architecture changes to properly pass the auth token down to the MetricsServiceClient

src/tabular-api-surface.ts
```diff
@@ -540,23 +567,31 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
-              // to Gemini, the appropriate code for a closed client is
-              // CANCELLED since the user actually cancelled the call by closing
-              // the client.
+              // a grpc code to be present when the client is closed. The
+              // appropriate code for a closed client is CANCELLED since the
+              // user actually cancelled the call by closing the client.
               //
               error.code = grpc.status.CANCELLED;
             }
+            metricsCollector?.onOperationComplete(
```
mutianf, line 576: where are the other metric labels recorded? e.g. instance id, table id, method, etc

mutianf (changes requested):

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -59,6 +69,31 @@ enum MetricsCollectorState {
...
 }
 
+// This method swallows errors when metrics debugging is not enabled so
+// that errors don't bubble up to the user.
+function withMetricsDebug<T>(fn: () => T): T | undefined {
+  try {
+    return fn();
+  } catch (e) {
+    if (METRICS_DEBUG) {
+      throw e;
```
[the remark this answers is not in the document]
danieljbruce: Yes it will. But only when the METRICS_DEBUG variable is enabled.

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -59,6 +69,31 @@ enum MetricsCollectorState {
...
+    }
```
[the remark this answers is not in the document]
danieljbruce: Sure. Logged a warning here instead.

src/client-side-metrics/operation-metrics-collector.ts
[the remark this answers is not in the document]
danieljbruce: This logs a warning now anyways

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -334,29 +393,33 @@ export class OperationMetricsCollector {
...
-      if (
-        decodedValue &&
-        (decodedValue as unknown as {clusterId: string}).clusterId
-      ) {
-        this.cluster = (
-          decodedValue as unknown as {clusterId: string}
-        ).clusterId;
+    withMetricsDebug(() => {
+      if (!this.zone || !this.cluster) {
+        const INSTANCE_INFORMATION_KEY = 'x-goog-ext-425905942-bin';
```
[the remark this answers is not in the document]
danieljbruce: Sure. Moved to global variable

src/client-side-metrics/exporter.ts
```diff
@@ -326,9 +323,11 @@ export class CloudMonitoringExporter extends MetricExporter {
             initialRetryDelayMillis: 5000,
             retryDelayMultiplier: 2,
             maxRetryDelayMillis: 50000,
+            totalTimeoutMillis: 50000,
           },
         );
-        await this.monitoringClient.createTimeSeries(
+
+        await this.client.createTimeSeries(
```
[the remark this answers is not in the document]
danieljbruce: I don't think so. Keep in mind that this is just the MetricServiceClient used for writing the metrics. What is your concern?

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -13,13 +13,22 @@
...
 import * as gax from 'google-gax';
-const root = gax.protobuf.loadSync(
-  './protos/google/bigtable/v2/response_params.proto',
+import {AbortableDuplex, BigtableOptions} from '../index';
+import * as path from 'path';
+import {IMetricsHandler} from './metrics-handler';
+
+// When this environment variable is set then print any errors associated
+// with failures in the metrics collector.
+const METRICS_DEBUG = process.env.METRICS_DEBUG;
```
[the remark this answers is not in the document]
danieljbruce: No. Default to false. We don't want to flood the customer's terminal with warnings if there is some kind of mistake?

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -59,10 +70,37 @@ enum MetricsCollectorState {
   OPERATION_COMPLETE,
 }
 
+// This method swallows errors when metrics debugging is not enabled so
```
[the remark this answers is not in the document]
danieljbruce: Right. Changed

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -59,10 +70,37 @@ enum MetricsCollectorState {
...
+// that errors don't bubble up to the user.
+function withMetricsDebug<T>(fn: () => T): T | undefined {
+  try {
+    return fn();
+  } catch (e) {
+    if (METRICS_DEBUG) {
+      console.warn('METRICS_DEBUG warning');
+      console.warn((e as ServiceError).message);
```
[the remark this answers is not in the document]
danieljbruce: I don't think it will include the stack trace. Just something like https://github.com/googleapis/nodejs-bigtable/blob/main/system-test/service-path.ts#L191.

src/index.ts
```diff
@@ -128,6 +140,13 @@ function getDomain(prefix: string, opts?: gax.ClientOptions) {
   }`;
 }
 
+function generateClientUuid() {
```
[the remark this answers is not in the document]
danieljbruce: Yes. This is not used at all.

src/tabular-api-surface.ts
```diff
@@ -332,14 +336,24 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
       }
       return originalEnd(chunk, encoding, cb);
     };
-
+    const metricsCollector =
+      this.bigtable._metricsConfigManager.createOperation(
+        MethodName.READ_ROWS,
+        StreamingState.STREAMING,
```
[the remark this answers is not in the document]
danieljbruce: You're right. `row.get` calls `createReadStream` here, but remember that createReadStream always hits the readRows endpoint. In fact, I don't even see a gapic method for readRow.

If a user calls `row.get` then do we want to log the metrics as `READ_ROW` because the user is just getting a single row or do we want to log it as `READ_ROWS` because really it is hitting the READ_ROWS grpc endpoint?

src/tabular-api-surface.ts
```diff
@@ -540,23 +557,25 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
-              // to Gemini, the appropriate code for a closed client is
-              // CANCELLED since the user actually cancelled the call by closing
-              // the client.
+              // a grpc code to be present when the client is closed. The
+              // appropriate code for a closed client is CANCELLED since the
+              // user actually cancelled the call by closing the client.
               //
               error.code = grpc.status.CANCELLED;
             }
+            metricsCollector.onOperationComplete(error.code);
```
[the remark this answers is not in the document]
danieljbruce: Attempt complete is already part of operation complete https://github.com/googleapis/nodejs-bigtable/pull/1571/files#diff-10f88e720bed4e5a35d5d0e992011fd658dc800b0ff6510fd4ef6137f036ae29R288.

src/tabular-api-surface.ts
```diff
@@ -540,23 +557,25 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
         })
         .on('data', _ => {
           // Reset error count after a successful read so the backoff
           // time won't keep increasing when as stream had multiple errors
           numConsecutiveErrors = 0;
+          metricsCollector.onResponse();
         })
         .on('end', () => {
           activeRequestStream = null;
+          metricsCollector.onOperationComplete(grpc.status.OK);
```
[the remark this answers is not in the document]
danieljbruce: Attempt complete is already part of operation complete https://github.com/googleapis/nodejs-bigtable/pull/1571/files#diff-10f88e720bed4e5a35d5d0e992011fd658dc800b0ff6510fd4ef6137f036ae29R288.

src/tabular-api-surface.ts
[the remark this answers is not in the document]
mutianf: If user calls row.get, we want to log the method as ReadRow and Streaming = false.

src/client-side-metrics/operation-metrics-collector.ts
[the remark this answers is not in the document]
danieljbruce: > We should have integration test to make sure the instrumentation to publish metrics are correct

There are tons of tests that do this like https://github.com/googleapis/nodejs-bigtable/blob/f8fd4be1d5f415b90739090bbd4ee6d1a2cfa021/system-test/gcp-metrics-handler.ts#L238 for example. It would have been impossible to develop the instrumentation without them. See:
https://github.com/googleapis/nodejs-bigtable/pull/1583
https://github.com/googleapis/nodejs-bigtable/pull/1569
https://github.com/googleapis/nodejs-bigtable/pull/1566

So you want me to set this variable to true?

src/tabular-api-surface.ts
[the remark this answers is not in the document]
danieljbruce: https://github.com/googleapis/nodejs-bigtable/pull/1612 is the change required to complete this work.

system-test/metric-service-client-credentials.ts
```diff
@@ -0,0 +1,91 @@
...
+      } catch (e) {
+        done(e);
+      }
+    })().catch(err => {
+      throw err;
+    });
+  });
+  it('should use second project for the metric service client', async () => {
+    const SECOND_PROJECT_ID = 'second-project-id';
+    const clientOptions = {metricsEnabled: true, projectId: SECOND_PROJECT_ID};
```
mutianf, line 66: is the SECOND_PROJECT_ID here the project id of the bigtable client?

mutianf (commented):
Do we have a test that tests the actual value we're recording to make sure the measurement is correct?

system-test/metric-service-client-credentials.ts
danieljbruce: Yes. I added a comment explaining what this test intends to do. The idea is that if the second project is provided to the Bigtable client then when the exporter uses the options it gets to build the metric service client the metric service client should use the right project.

system-test/client-side-metrics.ts
```diff
@@ -0,0 +1,514 @@
...
+        }
+      })().catch(err => {
+        throw err;
+      });
+    });
+  });
+  describe('Bigtable/ClientSideMetricsToGCMTimeout', () => {
+    // This test suite simulates a situation where the user creates multiple
+    // clients and ensures that the exporter doesn't produce any errors even
+    // when multiple clients are attempting an export.
```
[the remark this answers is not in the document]
danieljbruce: > This could happen if multiple clients share the same UUID and publishes metric data points too frequently.

Yes. There has been lots of discussion about the maximum sampling error and when it occurs. We even thought we needed design changes due to this issue, but then reverted such changes because the client uuids are different for every client and this is enough to avoid the maximum sampling error. `should send the metrics to Google Cloud Monitoring for a ReadRows call with thirty clients` test captures this multiple client case.

system-test/client-side-metrics.ts
```diff
@@ -0,0 +1,514 @@
...
+          for (const instanceId of [instanceId1, instanceId2]) {
+            await setupBigtable(bigtable, columnFamilyId, instanceId, [
+              tableId1,
+              tableId2,
+            ]);
+            const instance = bigtable.instance(instanceId);
+            const table = instance.table(tableId1);
+            await table.getRows();
+            const table2 = instance.table(tableId2);
+            await table2.getRows();
```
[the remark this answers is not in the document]
danieljbruce: I'm open to checking metrics publishing programmatically though. I'll need to research more on how to do that and it could be a time investment.

src/client-side-metrics/exporter.ts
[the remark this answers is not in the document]
daniel-sanche: Yeah, it [looks like it does exist](https://github.com/googleapis/google-cloud-node/blob/b41f79df5e4f9472761ce1383d4d85785c647a15/packages/google-cloud-monitoring/src/v3/metric_service_client.ts#L1207)

IIRC, createTimeSeries is more generic, but createServiceTimeSeries is meant for GCP metrics

src/tabular-api-surface.ts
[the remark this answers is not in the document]
daniel-sanche: @mutianf "ReadRow and Streaming = False"? I thought it was "ReadRow**s** and Streaming = False". And looking at go's implementation,[ I don't see ReadRow being used anywhere](https://github.com/search?q=repo%3Agoogleapis/google-cloud-go%20gaxInvokeWithRecorder&type=code). But let me know if I'm misunderstanding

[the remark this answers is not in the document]
daniel-sanche: Oh wait, [I dug up this conversation](https://github.com/googleapis/python-bigtable/pull/923#discussion_r1473408291) in the Python PR. Igor said:

>I think java ReadRow tag was my mistake :(
> Lets keep this inline with server metrics. I think the correct path forward will be stanley's request profiles

So let me know whether what you think

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -59,10 +70,36 @@ enum MetricsCollectorState {
...
+  } else {
+    throw Error('Invalid state transition');
+  }
+}
+
 /**
  * A class for tracing and recording client-side metrics related to Bigtable operations.
  */
 export class OperationMetricsCollector {
+  private readonly INSTANCE_INFORMATION_KEY = 'x-goog-ext-425905942-bin';
```
daniel-sanche, line 102: a comment here might be helpful, since it looks pretty random without context

src/index.ts
```diff
@@ -16,9 +16,15 @@ import {replaceProjectIdToken} from '@google-cloud/projectify';
...
-import {GoogleAuth, CallOptions, grpc as gaxVendoredGrpc} from 'google-gax';
+import {
+  GoogleAuth,
+  CallOptions,
+  grpc as gaxVendoredGrpc,
+  ClientOptions,
+} from 'google-gax';
 import * as gax from 'google-gax';
 import * as protos from '../protos/protos';
+import * as os from 'os';
```
daniel-sanche, line 27: looks like this isn't needed?

src/utils/createReadStreamInternal.ts
```diff
@@ -0,0 +1,445 @@
...
+ *     match.
+ * @param {object[]} [opts.ranges] A list of key ranges.
+ * @param {string} [opts.start] Start value for key range.
+ * @param {string} [viewName] The name of the authorized view, if applicable.
+ * @returns {stream} A readable stream of {@link Row} objects.
+ *
+ */
+export function createReadStreamInternal(
+  table: TabularApiSurface,
+  singleRow: boolean,
```
daniel-sanche, line 77: I don't know if changing this would be a lot of work, but I think it would be cleaner to pass a `metricsCollector` object here directly, instead of passing a bool and then using that to build different `metricsCollector` instances

src/utils/getRowsInternal.ts
```diff
@@ -0,0 +1,66 @@
...
+ * @param {function} cb The callback function.
+ * @param {?error} cb.err An error returned while making this request.
+ * @param {Row[]} cb.rows List of Row objects.
+ *
+ * @returns {Promise<GetRowsResponse>|void} Returns a promise that resolves with the rows if no callback is provided, otherwise calls the callback with the rows.
+ *
+ * @example <caption>include:samples/api-reference-doc-snippets/table.js</caption>
+ * region_tag:bigtable_api_get_rows
+ */
+export function getRowsInternal(
```
daniel-sanche, line 49: nit: It seems a but weird to have this helper that basically just defers to another helper, and is only used in a few places. I'd probably suggest to move this in with createReadStreamInternal (if not cut it out completely), since it seems like a pretty thin wrapper over it. I'm not sure if that would break a lot of tests or anything though

daniel-sanche (changes requested):
I left a new more comments. Along with those, it looks like there are these conversations that seem unresolved:
- createServiceTimeSeries: [https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2114495885](https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2114495885)
- remove retries? [https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2114496978](https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2114496978)
- READ_ROW vs READ_ROWS [https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2114556372](https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2114556372)
- Exceptions [https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2124659124](https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2124659124)
- Default METRICS_DEBUG to true? [https://github.com/googleapis/nodejs-bigtable/pull/1571/files#r2114536705](https://github.com/googleapis/nodejs-bigtable/pull/1571/files#r2114536705)

Overall, it's looking good!

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -13,13 +13,22 @@
...
+import {IMetricsHandler} from './metrics-handler';
+
+// When this environment variable is set then print any errors associated
+// with failures in the metrics collector.
+const METRICS_DEBUG = process.env.METRICS_DEBUG;
+
+const protoPath = path.join(
+  __dirname,
+  '../../protos/google/bigtable/v2/response_params.proto',
 );
```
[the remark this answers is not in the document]
daniel-sanche: She also said:
>The main thing is it might be good to see if you can pull the parsed protos via gax or gapic-loaded versions. The parsing is expensive and sometimes leads to memory leakage if it happens a lot. (The text of the protos will get stuck for some reason.)

Let me know if you think there's a way to load them directly from gapic, because that would be cleaner. But I'm not familiar enough with the node stack to say much more than that

src/utils/getRowsInternal.ts
danieljbruce: These changes were introduced in https://github.com/googleapis/nodejs-bigtable/pull/1612.

This is necessary and here is why:
`table.getRows` calls `getRowsInternal`.
`table.createReadStream` calls `createReadStreamInternal`.

`getRowsInternal` and `createReadStreamInternal` both need a parameter telling them if the call is for a single row or for a set of rows. We cannot add this parameter to `table.getRows` or `table.createReadStream` because it changes the API surface.

danieljbruce: If `createReadStreamInternal` was pasted inline into `getRowsInternal` then `createReadStream` would have nothing to call.

src/index.ts
danieljbruce: Removed. At this point there may be some automation or a bot we can set up for little changes like this. It could be a real time saver.

src/client-side-metrics/generate-client-uuid.ts
```diff
@@ -0,0 +1,39 @@
...
+ * - `<hostname>` is the hostname of the machine.
+ *
+ * @returns {string} A unique client identifier string.
+ */
+export function generateClientUuid() {
+  const hostname = os.hostname() || 'localhost';
+  const currentPid = process.pid || '';
+  const uuid4 = crypto.randomUUID();
+  return `node-${uuid4}-${currentPid}${hostname}`;
+}
```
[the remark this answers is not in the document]
danieljbruce: I implemented the alternative I suggested so that we could move this function.

src/utils/createReadStreamInternal.ts
danieljbruce: Yes. This is better and I added this change.

src/tabular-api-surface.ts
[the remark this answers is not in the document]
danieljbruce: This is more of a requirements question so I wouldn't know.

Right now, `table.getRows` and `table.createReadStream` both record for `READ_ROWS` and `STREAMING`
`row.get` records for `READ_ROW` and `Unary`.

@mutianf Let me know if that's correct.

system-test/client-side-metrics.ts
```diff
@@ -0,0 +1,514 @@
...
+                tableId2,
+              ]);
+              const instance = bigtable.instance(instanceId);
+              const table = instance.table(tableId1);
+              await table.getRows();
+              const table2 = instance.table(tableId2);
+              await table2.getRows();
+            }
+          }
+        } catch (e) {
```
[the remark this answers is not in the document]
danieljbruce: I mention the test function swallowing the error because async functions swallow errors.

This comment is not about exporter errors. The catch block would not catch exporter errors.

For exporter errors it looks like the error gets logged to the console.

For exporter errors we have this comment in the code:
// The resultCallback typically accepts a value equal to {code: x}
// for some value x along with other info. When the code is equal to 0
// then the operation completed successfully. When the code is not equal
// to 0 then the operation failed. Open telemetry logs errors to the
// console when the resultCallback passes in non-zero code values and
// logs nothing when the code is 0.

Actually I just did an experiment on https://github.com/googleapis/nodejs-bigtable/compare/359913994-third-PR-CSM-experiment?expand=1 and it looks like it doesn't log anything to the console at all. We have to correct this comment. Do we want errors to get printed to the console?

danieljbruce:
> I left a new more comments. Along with those, it looks like there are these conversations that seem unresolved:
> 
> * createServiceTimeSeries: [feat: Add client side metrics collection to readRows calls #1571 (comment)](https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2114495885)
> * remove retries? [feat: Add client side metrics collection to readRows calls #1571 (comment)](https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2114496978)
> * READ_ROW vs READ_ROWS [feat: Add client side metrics collection to readRows calls #1571 (comment)](https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2114556372)
> * Exceptions [feat: Add client side metrics collection to readRows calls #1571 (comment)](https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2124659124)
> * Default METRICS_DEBUG to true? https://github.com/googleapis/nodejs-bigtable/pull/1571/files#r2114536705
> 
> Overall, it's looking good!

Do we want to set METRICS_DEBUG to true?

system-test/client-side-metrics.ts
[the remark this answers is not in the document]
danieljbruce: Corrected the comment

daniel-sanche (dismissed):
It looks like my concerns have all been addressed. The two remaining questions for Mattie that I see are:
- do we want to record the read_row metric as READ_ROWS (like Java), or READ_ROW (like Igor suggested for Python)? (https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2114556372)
- Should METRICS_DEBUG default on or off? (https://github.com/googleapis/nodejs-bigtable/pull/1571/files#r2114536705)

Neither of those need to block this, so we can open issues to pick them up later if needed

daniel-sanche (dismissed):

mutianf (dismissed):
LGTM after the final nits

mutianf (dismissed):

daniel-sanche (approved):

PR #1575, opened 2025-03-17
feat: Add the plumbing for application blocking latencies client side metrics

danieljbruce:
**Summary:**

This PR adds the plumbing for application blocking latencies. Right now plumbing exists for all the other client side metrics that we want to support so we just add the application blocking latencies to that plumbing.

**Background:**

(Explain application blocking latencies)

**Changes:**

Important source code changes:

`src/client-side-metrics/gcp-metrics-handler.ts`: When operations complete this now records application latencies to the open telemetry instruments ready for export.

`src/client-side-metrics/metrics-handler.ts:` The interface representing the operation complete data now includes application latencies.

`src/client-side-metrics/operation-metrics-collector.ts:` Added a method to the metrics collector that is called when the row reaches the user and also code that passes application latencies to the metrics handler when the operation completes.

Test changes:

`test-common/expected-otel-export-input.ts:` The numbers are slightly different for the fixture because of the calls to `onRowReachesUser` from the fake client.

`test-common/metrics-handler-fixture.ts:`  The numbers are slightly different for the fixture because of the calls to `onRowReachesUser` from the fake client.

`test/metrics-collector/metrics-collector.ts:` The fake client is changed so that it makes calls to `onRowReachesUser` when it receives rows.

`test/metrics-collector/typical-method-call.txt`: The numbers for the test output are slightly different because of the calls to onRowReachesUser.

**Next Steps:**

- Merge [the code that adds client side metrics to readRows calls](https://github.com/googleapis/nodejs-bigtable/pull/1571)
- Add a PR to start collecting application latencies for readRows calls that makes use of the plumbing in this PR.


src/client-side-metrics/exporter.ts
```diff
@@ -157,15 +157,16 @@ function getMetric(
...
   const streaming = dataPoint.attributes.streaming;
   const app_profile = dataPoint.attributes.app_profile;
+  const status = dataPoint.attributes.status;
   return {
     type: metricName,
     labels: Object.assign(
       {
         method: dataPoint.attributes.method,
         client_uid: dataPoint.attributes.client_uid,
-        status: dataPoint.attributes.status,
```
danieljbruce, line 166: If `status: undefined` is included for application latencies then the MetricServiceClient call will fail with an error so this step is necessary.

daniel-sanche: The reason for this is because application_latencies doesn't include status, right?

So status on line 169 will be null, meaning it won't be sent? Or am I reading this wrong?

Just want to make sure I understand why status would be undefined here

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -76,6 +76,8 @@ export class OperationMetricsCollector {
   private serverTime: number | null;
   private connectivityErrorCount: number;
   private streamingOperation: StreamingState;
+  private applicationLatencies: number[];
+  private lastRowReceivedTime: Date | null;
```
daniel-sanche, line 80: is Date a safe way to store timestamps in node? 

My understanding is that Date is based on the system time, so it's not monotonic, and could move backwards. A quick google search shows [process.hrtime](https://nodejs.org/api/process.html#processhrtimebigint) might be a better fit for this kind of thing

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -281,6 +288,25 @@ export class OperationMetricsCollector {
...
+   * Called when a row from the Bigtable stream reaches the application user.
+   *
+   * This method is used to calculate the latency experienced by the application
+   * when reading rows from a Bigtable stream. It records the time between the
+   * previous row being received and the current row reaching the user. These
+   * latencies are then collected and reported as `applicationBlockingLatencies`
+   * when the operation completes.
+   */
+  onRowReachesUser() {
+    const currentTime = new Date();
```
daniel-sanche, line 301: Maybe you should check that you're in the expected state here?

danieljbruce: Alright. I added a guard here.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: We only measure deltas between dates, but just to be safe I switched to using hrtime everywhere.

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -148,9 +154,11 @@ export class OperationMetricsCollector {
       this.state =
         MetricsCollectorState.OPERATION_STARTED_ATTEMPT_NOT_IN_PROGRESS;
       this.attemptCount++;
-      const endTime = new Date();
+      const endTime = hrtime.bigint();
       if (projectId && this.attemptStartTime) {
-        const totalTime = endTime.getTime() - this.attemptStartTime.getTime();
+        const totalTime = Number(
+          (endTime - this.attemptStartTime) / BigInt(1000000)
```
daniel-sanche, line 160: nit: it would be good to capture that this is in milliseconds somewhere. Either a comment beside this conversion, or renaming variables to something like `totalMilliseconds` and `operationStartTimeNanos`

daniel-sanche (approved):
LGTM, but consider making the time formats more explicit

src/client-side-metrics/operation-metrics-collector.ts
daniel-sanche: Also, it looks like this i a float, correct? Is that right?

danieljbruce: > Also, it looks like this i a float, correct? Is that right?

It looks like it rounds it off for a number type.

image

danieljbruce: image

danieljbruce: And I added comments / variable name changes

PR #1582, opened 2025-04-03
chore!: migrate to Node 18

danieljbruce:
This PR is the PR with the fixes necessary for upgrading to Node 18.

To produce this PR I:
- Took changes from the owl-bot-copy generator branch
- Fixed some compiler errors
- Ran the pre-processor
- Fixed some more compiler errors
- Fixed the linting issues


.github/.OwlBot.lock.yaml
```diff
@@ -12,6 +12,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 docker:
-  image: gcr.io/cloud-devrel-public-resources/owlbot-nodejs:latest
-  digest: sha256:e92044720ab3cb6984a70b0c6001081204375959ba3599ef6c42dd99a7783a67
-# created: 2023-11-10T00:24:05.581078808Z
+  image: 'gcr.io/cloud-devrel-public-resources/owlbot-nodejs:latest'
+  digest: 'sha256:c7e4968cfc97a204a4b2381f3ecb55cabc40c4cccf88b1ef8bef0d976be87fee'
```
sofisl, line 16: not sure if it matters but it doesn't seem like these had quotes beforehand.

src/mutation.ts
```diff
@@ -14,7 +14,7 @@
 
 import arrify = require('arrify');
 import * as is from 'is';
-import * as Long from 'long';
+import Long = require('long');
```
sofisl, line 17: ah, this is probably why long needs to be a dependency. Do you know why the way its being imported is changing to require?

test/chunktransformer.ts
```diff
@@ -14,7 +14,7 @@
 
 import * as assert from 'assert';
 import {describe, it, before, beforeEach, afterEach} from 'mocha';
-import * as Long from 'long';
+import Long = require('long');
```
sofisl, line 17: why is this import statement changing?

.github/.OwlBot.lock.yaml
danieljbruce: Good point. I changed it back.

src/mutation.ts
danieljbruce: IIRC, some compiler or linting error wasn't letting us do it the old way, but we must have made a recent change because it doesn't seem to be a problem anymore so I changed it back to an import.

test/chunktransformer.ts
danieljbruce: IIRC, some compiler or linting error wasn't letting us do it the old way, but we must have made a recent change because it doesn't seem to be a problem anymore so I changed it back to an import.

sofisl (approved):

PR #1583, opened 2025-04-08
fix: Fix plumbing errors for client side metrics collection

danieljbruce:
**Summary:**

While testing client side metrics collection with the client library some errors were observed and investigated. This PR fixes those errors and some of the tests that used to be flakey will pass much more consistently now.

**Changes:**

**`src/client-side-metrics/exporter.ts:`** The change in this file ensures the call with the MetricsServiceClient to record the metric includes nanoseconds. We observed an error that said `The start time must be before the end time (2025-04-04T11:12:56-07:00) for the non-gauge metric` because the start time and end time were the same since they both only included seconds. We don't want errors if the span of the metrics collection is under a second so we should include nanoseconds as well so that this error doesn't appear.

**`src/client-side-metrics/gcp-metrics-handler.ts:`** A change is applied so that we have one instrument stack per project instead of one instrument stack per client/handler. This is necessary because if a user creates multiple clients/handlers then they could get many exports happening at the same time resulting in a `One or more points were written more frequently than the maximum sampling period configured for the metric` error. Therefore, the GCP Metrics handler now uses a global singleton map of project ids to instrument stacks instead of an instrument stack per handler.

**`test-common/expected-otel-export-input.ts:`** For existing fixtures the converted export input sent to the metric service backend will now contain nanoseconds as per the change in `src/client-side-metrics/exporter.ts`. Also, a fixture is added containing the one export that now happens with 100 operations when 100 metrics handlers are used with one instrument stack.

**`system-test/gcp-metrics-handler.ts:`** Now that the instrument stacks are one per project instead of one per handler, the tests need to change to only expect one export when multiple handlers are used, but also to expect all the data in that export. In addition to this, an empty map is used in each test for the singleton map from project ids to instrument stacks so that the tests do not have dependencies on one another. Finally, if exports don't happen then the test will throw an error instead of exiting with code 0 quietly.

### DIFF
--- a/src/client-side-metrics/exporter.ts
+++ b/src/client-side-metrics/exporter.ts
@@ -23,6 +23,7 @@ import {ServiceError} from 'google-gax';
 import {MetricServiceClient} from '@google-cloud/monitoring';
 import {google} from '@google-cloud/monitoring/build/protos/protos';
 import ICreateTimeSeriesRequest = google.monitoring.v3.ICreateTimeSeriesRequest;
+import {RetryOptions} from 'google-gax';
 
 export interface ExportResult {
   code: number;
@@ -304,8 +305,20 @@ export class CloudMonitoringExporter extends MetricExporter {
     (async () => {
       try {
         const request = metricsToRequest(metrics);
+        // In order to manage the "One or more points were written more
+        // frequently than the maximum sampling period configured for the
+        // metric." error we should have the metric service client retry a few
+        // times to ensure the metrics do get written.
+        const retry = new RetryOptions([3, 4, 8, 10, 14], {
+          initialRetryDelayMillis: 10,
+          retryDelayMultiplier: 2,
+          maxRetryDelayMillis: 10000,
+        });
         await this.monitoringClient.createTimeSeries(
-          request as ICreateTimeSeriesRequest
+          request as ICreateTimeSeriesRequest,
+          {
+            retry,
+          }
         );
         // The resultCallback typically accepts a value equal to {code: x}
         // for some value x along with other info. When the code is equal to 0
--- a/src/client-side-metrics/exporter.ts
+++ b/src/client-side-metrics/exporter.ts
@@ -25,7 +25,9 @@ import {google} from '@google-cloud/monitoring/build/protos/protos';
 import ICreateTimeSeriesRequest = google.monitoring.v3.ICreateTimeSeriesRequest;
 import {Mutex} from 'async-mutex';
 
-// This mutex is a global singleton that ensures the
+// This mutex is a global singleton that ensures the metric exports to GCM
+// happen in the order they are passed into the export function to avoid
+// "Points must be written in order" errors.
 const mutex = new Mutex();
 
 export interface ExportResult {
--- a/src/client-side-metrics/exporter.ts
+++ b/src/client-side-metrics/exporter.ts
@@ -309,6 +309,8 @@ export class CloudMonitoringExporter extends MetricExporter {
         // frequently than the maximum sampling period configured for the
         // metric." error we should have the metric service client retry a few
         // times to ensure the metrics do get written.
+        // We use all the usual retry codes plus code 3 because 3 corresponds
+        // to the maximum sampling error.
         const retry = new RetryOptions([3, 4, 8, 10, 14], {
           initialRetryDelayMillis: 10,
           retryDelayMultiplier: 2,
--- a/src/client-side-metrics/exporter.ts
+++ b/src/client-side-metrics/exporter.ts
@@ -312,9 +312,9 @@ export class CloudMonitoringExporter extends MetricExporter {
         // We use all the usual retry codes plus code 3 because 3 corresponds
         // to the maximum sampling error.
         const retry = new RetryOptions([3, 4, 8, 10, 14], {
-          initialRetryDelayMillis: 10,
+          initialRetryDelayMillis: 5000,
           retryDelayMultiplier: 2,
-          maxRetryDelayMillis: 10000,
+          maxRetryDelayMillis: 50000,
         });
         await this.monitoringClient.createTimeSeries(
           request as ICreateTimeSeriesRequest,
--- a/src/client-side-metrics/exporter.ts
+++ b/src/client-side-metrics/exporter.ts
@@ -23,6 +23,10 @@ import {ServiceError} from 'google-gax';
 import {MetricServiceClient} from '@google-cloud/monitoring';
 import {google} from '@google-cloud/monitoring/build/protos/protos';
 import ICreateTimeSeriesRequest = google.monitoring.v3.ICreateTimeSeriesRequest;
+import {Mutex} from 'async-mutex';
+
+// This mutex is a global singleton that ensures the
+const mutex = new Mutex();
 
 export interface ExportResult {
   code: number;
@@ -307,9 +311,12 @@ export function metricsToRequest(exportArgs: ResourceMetrics) {
       }
     }
   }
+  console.log('exported');
+  // TODO: This sortTimeSeriesByStartTime call isn't necessary.
   const sortedTimeSeriesArray = sortTimeSeriesByStartTime(
     timeSeriesArray as google.monitoring.v3.ITimeSeries[]
   );
+  console.log(JSON.stringify(sortedTimeSeriesArray));
   return {
     name: `projects/${projectId}`,
     timeSeries: sortedTimeSeriesArray,
@@ -371,16 +378,28 @@ export class CloudMonitoringExporter extends MetricExporter {
     (async () => {
       try {
         const request = metricsToRequest(metrics);
-        await this.monitoringClient.createTimeSeries(
-          request as ICreateTimeSeriesRequest
-        );
-        // The resultCallback typically accepts a value equal to {code: x}
-        // for some value x along with other info. When the code is equal to 0
-        // then the operation completed successfully. When the code is not equal
-        // to 0 then the operation failed. Open telemetry logs errors to the
-        // console when the resultCallback passes in non-zero code values and
-        // logs nothing when the code is 0.
-        resultCallback({code: 0});
+        // Even if the data from various metrics handlers gets passed into
+        // export in order, if we allow the second export call to call
+        // createTimeSeries before the first export call completes then there is
+        // a chance the second createTimeSeries call reaches the server first
+        // causing a "Points must be written in order" error. Therefore, we
+        // need a mutex to ensure points are written in order.
+        await mutex.runExclusive(async () => {
+          try {
+            await this.monitoringClient.createTimeSeries(
+              request as ICreateTimeSeriesRequest
+            );
+            // The resultCallback typically accepts a value equal to {code: x}
+            // for some value x along with other info. When the code is equal to 0
+            // then the operation completed successfully. When the code is not equal
+            // to 0 then the operation failed. Open telemetry logs errors to the
+            // console when the resultCallback passes in non-zero code values and
+            // logs nothing when the code is 0.
+            resultCallback({code: 0});
+          } catch (e: any) {
+            resultCallback(e as ServiceError);
+          }
+        });
       } catch (error) {
         resultCallback(error as ServiceError);
       }
--- a/src/client-side-metrics/exporter.ts
+++ b/src/client-side-metrics/exporter.ts
@@ -59,7 +59,7 @@ function getInterval(
     },
     startTime: {
       seconds: dataPoint.startTime[0],
-      nanos: dataPoint.endTime[1],
+      nanos: dataPoint.startTime[1],
     },
   };
 }
--- a/src/client-side-metrics/exporter.ts
+++ b/src/client-side-metrics/exporter.ts
@@ -175,6 +175,65 @@ function getMetric(
 }
 
 /**
+ * In order to avoid encountering an error that says "Points must be written in
+ * order. One or more of the points specified had an older start time than the
+ * most recent point." we need to sort the data points before sending them
+ * to the metrics service client backend.
+ *
+ * @param dataPoints The data points to be sorted
+ */
+function sortDataPointsByStartTime(
+  dataPoints:
+    | DataPoint<number>[]
+    | DataPoint<Histogram>[]
+    | DataPoint<ExponentialHistogram>[]
+) {
+  if (!Array.isArray(dataPoints)) {
+    console.warn('Invalid dataPoints array provided.');
+    return [];
+  }
+
+  // Create a copy to avoid modifying the original array
+  const sortedDataPoints = [...dataPoints].sort((a, b) => {
+    // Assuming startTime is an array [seconds, nanoseconds]
+    if (
+      Array.isArray(a.startTime) &&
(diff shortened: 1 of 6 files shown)


system-test/gcp-metrics-handler.ts
```diff
@@ -147,45 +156,7 @@ describe('Bigtable/GCPMetricsHandler', () => {
...
-                assert.deepStrictEqual(
-                  parsedExportInput.scopeMetrics[0].metrics[index],
-                  expectedOtelExportInput.scopeMetrics[0].metrics[index]
-                );
-              }
-            } catch (e) {
-              // The error needs to be caught so it can be reported to the mocha
-              // test runner.
-              done(e);
-            }
```
danieljbruce, line 188: The checks above have been removed due to the fact that they would fail now since there is now one instrument stack with 1 export instead of 2, but twice the data. `Should export a value to a hundred GCPMetricsHandlers` basically does these checks so we don't really need a new fixture and test here just for this check.

src/client-side-metrics/exporter.ts
```diff
@@ -55,9 +55,11 @@ function getInterval(
   return {
     endTime: {
       seconds: dataPoint.endTime[0],
+      nanos: dataPoint.endTime[1],
     },
     startTime: {
       seconds: dataPoint.startTime[0],
+      nanos: dataPoint.endTime[1],
```
danieljbruce, line 62: Oops. This should be startTime

daniel-sanche:
> src/client-side-metrics/gcp-metrics-handler.ts: A change is applied so that we have one instrument stack per project instead of one instrument stack per client/handler. This is necessary because if a user creates multiple clients/handlers then they could get many exports happening at the same time resulting in a One or more points were written more frequently than the maximum sampling period configured for the metric error. Therefore, the GCP Metrics handler now uses a global singleton map of project ids to instrument stacks instead of an instrument stack per handler.

Does this mean the problem still exists if many projects are used? 

What is throwing the error? Is it from OTel or GCP?

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -63,8 +69,14 @@ export class GCPMetricsHandler implements IMetricsHandler {
    *   metrics to Google Cloud Monitoring. This exporter is responsible for
    *   sending the collected metrics data to the monitoring backend. The provided exporter must be fully configured, for example the projectId must have been set.
    */
-  constructor(exporter: PushMetricExporter) {
+  constructor(
+    exporter: PushMetricExporter,
```
daniel-sanche, line 73: Shouldn't this be a CloudMonitoringExporter? Where is that configured?

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -28,6 +28,12 @@ const {
   PeriodicExportingMetricReader,
 } = require('@opentelemetry/sdk-metrics');
 
+// The variable below is the singleton map from projects to instrument stacks
+// which exists so that we only create one instrument stack per project. This
+// will eliminate errors due to the maximum sampling period.
+const defaultProjectToInstruments: {[projectId: string]: MetricsInstruments} =
```
daniel-sanche, line 34: How does this fix the issue? Won't there still be an issue if multiple projects write at the same time?

daniel-sanche (changes requested):

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: If you change this to CloudMonitoringExporter then it will break [this test](https://github.com/googleapis/nodejs-bigtable/blob/c37a451c52ff5639fc9f5e76c88e93ba549fe6c9/test/metrics-collector/gcp-metrics-handler.ts#L118).

In practice only `CloudMonitoringExporter` objects will be passed in here when the client library is used, but if we want to unit test this object by seeing what gets passed into the exporter then we have to use the more general `PushMetricExporter` interface.
[CloudMonitoringExporter](https://github.com/googleapis/nodejs-bigtable/blob/c37a451c52ff5639fc9f5e76c88e93ba549fe6c9/src/client-side-metrics/exporter.ts#L295C46-L295C60) extends MetricExporter which implements PushMetricExporter interface.

danieljbruce: image

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -79,7 +91,7 @@ export class GCPMetricsHandler implements IMetricsHandler {
     // The projectId is needed per metrics handler because when the exporter is
     // used it provides the project id for the name of the time series exported.
     // ie. name: `projects/${....['monitored_resource.project_id']}`,
-    if (!this.otelInstruments) {
+    if (!this.projectToInstruments[projectId]) {
```
daniel-sanche, line 94: nit: this might be a bit easier to read if you break out the instrument creation into a helper:

```
private createInstrumentsForProject(projectId: String): MetricInstruments {
    ...
}

private getInstruments(projectId: string): MetricsInstruments {
    if (!this.projectToInstruments[projectId]) {
       this.projectToInstruments[projectId] = createInstrumentsForProject(projectId)
    }
    return  this.projectToInstruments[projectId]
}
```

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -49,8 +55,8 @@ interface MetricsInstruments {
  * associating them with relevant attributes for detailed analysis in Cloud Monitoring.
  */
 export class GCPMetricsHandler implements IMetricsHandler {
-  private otelInstruments?: MetricsInstruments;
   private exporter: PushMetricExporter;
+  private projectToInstruments: {[projectId: string]: MetricsInstruments};
```
daniel-sanche, line 59: This is supposed to be global shared state, right? Can't we make this a static variable to be shared across all instances?

I find it confusing that this looks like an instance variable, but we actually expect it to point to a shared singleton in practice.

daniel-sanche, line 59: nit: `instrumentsForProject` feels more descriptive, but maybe that's just me

Or maybe just `instrumentStore`, since it's meant to be internal, and accessed through getInstruments?

daniel-sanche (changes requested):

src/client-side-metrics/exporter.ts
```diff
@@ -302,8 +305,22 @@ export class CloudMonitoringExporter extends MetricExporter {
...
       try {
         const request = metricsToRequest(metrics);
+        // In order to manage the "One or more points were written more
+        // frequently than the maximum sampling period configured for the
+        // metric." error we should have the metric service client retry a few
+        // times to ensure the metrics do get written.
+        // We use all the usual retry codes plus code 3 because 3 corresponds
+        // to the maximum sampling error.
+        const retry = new RetryOptions([3, 4, 8, 10, 14], {
+          initialRetryDelayMillis: 10,
```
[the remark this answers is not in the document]
danieljbruce: Changed to 5000.

src/client-side-metrics/exporter.ts
```diff
@@ -302,8 +305,22 @@ export class CloudMonitoringExporter extends MetricExporter {
     (async () => {
       try {
         const request = metricsToRequest(metrics);
+        // In order to manage the "One or more points were written more
+        // frequently than the maximum sampling period configured for the
+        // metric." error we should have the metric service client retry a few
+        // times to ensure the metrics do get written.
+        // We use all the usual retry codes plus code 3 because 3 corresponds
+        // to the maximum sampling error.
+        const retry = new RetryOptions([3, 4, 8, 10, 14], {
```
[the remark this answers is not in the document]
danieljbruce: > Also, isn't error code 3 [INVALID_ARGUMENT](https://grpc.io/docs/guides/status-codes/)? That seems unexpected for this kind of error... Do you have a stack trace you can share?

[this comment from earlier](https://b.corp.google.com/issues/408462562#comment4) contains a copy/paste of the error. It is code 3.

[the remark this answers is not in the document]
danieljbruce: > Can this be structured in a way that doesn't use magic numbers of all the retries?

Yes. I made this update

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -63,8 +69,14 @@ export class GCPMetricsHandler implements IMetricsHandler {
    *   metrics to Google Cloud Monitoring. This exporter is responsible for
    *   sending the collected metrics data to the monitoring backend. The provided exporter must be fully configured, for example the projectId must have been set.
    */
-  constructor(exporter: PushMetricExporter) {
+  constructor(
+    exporter: PushMetricExporter,
+    projectToInstruments: {
+      [projectId: string]: MetricsInstruments;
+    } = defaultProjectToInstruments
+  ) {
```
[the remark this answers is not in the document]
danieljbruce: In our tests, ie. `Should export a value to a hundred GCPMetricsHandlers` etc, the tests need to start with a map that doesn't have the OTEL stack initialized yet. This ensures that if we run them in isolatation with `.only` or if we run them with the other tests then they will export a fixed number of metrics that match the assertion statement. For instance, for the operation latencies with `Should export a value to a hundred GCPMetricsHandlers` this amount is 100. We don't want it to reuse the OTEL stack from the previous test because then the OTEL stack will have 101 metrics.

We can try to eliminate this `projectToInstruments` instance variable and constructor argument by resetting the global singleton to an empty map, but then we run into other issues when we try to reset this value to an empty map in each test:

image

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: Sure. Done.

src/client-side-metrics/gcp-metrics-handler.ts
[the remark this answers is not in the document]
danieljbruce: Actually, now that we use the static variable like you suggested later, this argument isn't needed anymore so has been removed.

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -121,7 +133,7 @@ export class GCPMetricsHandler implements IMetricsHandler {
         ],
       });
       const meter = meterProvider.getMeter('bigtable.googleapis.com');
-      this.otelInstruments = {
+      this.projectToInstruments[projectId] = {
```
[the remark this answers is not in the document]
danieljbruce: Oops. I accidentally edited your comment. I meant to say "as far as I know, Node is single threaded. So if a thread encounters if (!this.projectToInstruments[projectId]) { then it will reach this line above before any other events on the event loop are processed."

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: Yes. This is a really good idea and it solves the problem of resetting the map in the tests too.

danieljbruce: Renamed to `instrumentsForProject`.

src/client-side-metrics/exporter.ts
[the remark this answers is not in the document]
daniel-sanche: Ok, that doesn't seem like the right error code to me (["INVALID_ARGUMENT indicates arguments that are problematic regardless of the state of the system"](https://grpc.io/docs/guides/status-codes/)). But that would be a backend problem

Thanks!

src/client-side-metrics/gcp-metrics-handler.ts
daniel-sanche: I'm not too familiar with node, but do you have to pass all mocks through the constructor? Can't you monkey patch the variable with `handler.exporter = mock()` in the test? Or use a testing framework to help inject mocks? 

Do we really need to expose up the internals for every class to make them testable? How do we handle this for public classes?

daniel-sanche: > I think this also falls under the principle of [programming to an interface instead of an implementation](https://ahmed-waleed.medium.com/program-to-an-interface-not-an-implementation-27c9a6c8fb61) we should be following more often.

Adding abstractions comes with a cost. We should decide what we actually want to make configurable, and aggressively encapsulate the rest of the complexity. A lot of my feedback on these PRs has been asking for more focused, concrete class structures. I'm happy to talk about this in more detail if you want

> An alternative to consider is that we could create a DefaultGCPMetricsHandler that is used in the client library in practice that only allows the CloudMonitoringExporter objects to be passed in. This way we don't have to modify the test with the MockExporter.

No, the motivation for this comment was to keep the classes simple. If we can get away with only having the DefaultGCPMetricsHandler that would be great, otherwise we can leave it as-is

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -43,14 +43,151 @@ interface MetricsInstruments {
...
 }
 
+/**
+ * This method gets the open telemetry instruments that will store GCP metrics
+ * for a particular project.
+ *
+ * @param projectId The project for which the instruments will be stored.
+ * @param exporter The exporter the metrics will be sent to.
+ */
+function getInstruments(projectId: string, exporter: PushMetricExporter) {
```
daniel-sanche, line 53: It's a bit confusing that there are two `getInstruments` functions. I'd suggest calling this `createInstruments` or `initilizeInstrumentsForProject` or something

daniel-sanche (approved):
Looks good overall!

I think better mock injection could help reduce some complexity, as we discussed [in this thread](https://github.com/googleapis/nodejs-bigtable/pull/1583#discussion_r2034178097), but that's an entirely separate discussion from this PR

other than that, just a few optional notes

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: >Or use a testing framework to help inject mocks?

We use proxyquire for mocking in some places which might work here, but I find proxyquire confusing to work with so I think it's best to avoid it if possible.

>We should decide what we actually want to make configurable, and aggressively encapsulate the rest of the complexity.

I think this is definitely true if the user has access to the class and is also a good point in general. In this case though, since we have to support the test then we have to make the exporter configurable or use proxyquire, but proxyquire is a bit confusing sometimes.

src/client-side-metrics/exporter.ts
```diff
@@ -302,8 +305,31 @@ export class CloudMonitoringExporter extends MetricExporter {
     (async () => {
       try {
         const request = metricsToRequest(metrics);
+        // In order to manage the "One or more points were written more
+        // frequently than the maximum sampling period configured for the
+        // metric." error we should have the metric service client retry a few
+        // times to ensure the metrics do get written.
+        // We use all the usual retry codes plus code 3 because 3 corresponds
+        // to the maximum sampling error.
```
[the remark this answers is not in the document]
danieljbruce: Done

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: It is now createInstruments. Done.

PR #1587, opened 2025-04-16
fix: For client side metrics, localize the MetricServiceClient

danieljbruce:
**Summary:**

Right now the MetricServiceClient will properly record metrics to a second project if the client is initialized with the second project, but this PR addresses a particular edge case we should cover.

Before this change, if the user cannot authenticate with their default project and is instead using a second project when initializing the client then the library will try to authenticate MetricServiceClient with the default project and fail. Therefore, it makes sense to localize the instantiation of the MetricServiceClient when the projectId is available and record metrics with the local MetricServiceClient.


danieljbruce:
This PR is dependent on https://github.com/googleapis/nodejs-bigtable/pull/1571 and is a small change so I'm just going to make this part of that PR.

PR #1590, opened 2025-04-17
fix: In Client-side metrics makes sure that the right views get created for the right metrics

danieljbruce:
## Description

This is a one line fix. If the aggregation ends with 'latencies' then the explicit bucket histogram should be used, not the sum histogram and vice versa.


daniel-sanche (dismissed):
LGTM, but consider a test so this can't slip through in the future

daniel-sanche (approved):

PR #1604, opened 2025-05-02
feat: Add full support for Universe Domain

danieljbruce:
## Description

Adds local testing for all universe domain use cases with a service account. Adds support for `universeDomain` option passed into the handwritten client. Passes universeDomain to the Gapic client to solve the `Error: The configured universe domain (googleapis.com) does not match the universe domain found in the credentials (apis-tpczero.goog)` error.

Also, an upstream change from google-gax is now providing more information in errors so some assert statements needed to be slightly modified.

## Impact

Fixes all universe domain use case and provides code so that we can continue to validate it going forward.

## Testing

Added tests to be run locally with the service account credentials for the project with the universe domain. Adding new service account credentials so that this can be tested in the CI pipeline is out of scope.

## Additional Information

An Alternative was considered where we only pass a service path into the gapic client when a custom API is provided, but that produced an error that says `Error: 14 UNAVAILABLE: Name resolution failed for target dns:null:443  before any response was received.`.


src/index.ts
```diff
@@ -467,10 +519,11 @@ export class Bigtable {
     const dataOptions = Object.assign(
       {},
       baseOptions,
+      getUniverseDomainOptions(options, options.BigtableClient),
```
danieljbruce, line 522: Without this, ie. without providing a universeDomain to the gapic client we see an error that says: `Error: The configured universe domain (googleapis.com) does not match the universe domain found in the credentials (apis-tpczero.goog). If you haven't configured the universe domain explicitly, googleapis.com is the default.`

src/index.ts
```diff
@@ -488,20 +541,26 @@ export class Bigtable {
     const adminOptions = Object.assign(
       {},
       baseOptions,
+      getUniverseDomainOptions(options, options.BigtableTableAdminClient),
```
danieljbruce, line 544: Without this, ie. without providing a universeDomain to the gapic client we see an error that says: `Error: The configured universe domain (googleapis.com) does not match the universe domain found in the credentials (apis-tpczero.goog). If you haven't configured the universe domain explicitly, googleapis.com is the default.`

src/index.ts
```diff
@@ -488,20 +541,26 @@ export class Bigtable {
...
           customEndpointBaseUrl ||
-          getDomain('bigtableadmin', options.BigtableClient),
+          getDomain('bigtableadmin', options, options.BigtableTableAdminClient),
       },
       options,
     );
     const instanceAdminOptions = Object.assign(
       {},
       baseOptions,
+      getUniverseDomainOptions(options, options.BigtableInstanceAdminClient),
```
danieljbruce, line 555: Without this, ie. without providing a universeDomain to the gapic client we see an error that says: `Error: The configured universe domain (googleapis.com) does not match the universe domain found in the credentials (apis-tpczero.goog). If you haven't configured the universe domain explicitly, googleapis.com is the default.`

src/index.ts
```diff
@@ -467,10 +519,11 @@ export class Bigtable {
     const dataOptions = Object.assign(
       {},
       baseOptions,
+      getUniverseDomainOptions(options, options.BigtableClient),
       {
         servicePath:
           customEndpointBaseUrl ||
-          getDomain('bigtable', options.BigtableClient),
+          getDomain('bigtable', options, options.BigtableClient),
```
danieljbruce, line 526: This now supports the use case where the universe domain is provided in the handwritten client options as well as maintaining support for when the universe domain is provided just for the gapic client.

src/index.ts
```diff
@@ -488,20 +541,26 @@ export class Bigtable {
     const adminOptions = Object.assign(
       {},
       baseOptions,
+      getUniverseDomainOptions(options, options.BigtableTableAdminClient),
       {
         servicePath:
           customEndpointBaseUrl ||
-          getDomain('bigtableadmin', options.BigtableClient),
+          getDomain('bigtableadmin', options, options.BigtableTableAdminClient),
```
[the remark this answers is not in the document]
danieljbruce: Note that this should be `BigtableTableAdminClient` as this is a setting for the adminOptions.

src/index.ts
danieljbruce: Note that this should be BigtableInstanceAdminClient as this is a setting for the instance admin client.

src/index.ts
```diff
@@ -488,20 +541,26 @@ export class Bigtable {
...
+      getUniverseDomainOptions(options, options.BigtableInstanceAdminClient),
       {
         servicePath:
           customEndpointBaseUrl ||
-          getDomain('bigtableadmin', options.BigtableClient),
+          getDomain(
+            'bigtableadmin',
+            options,
+            options.BigtableInstanceAdminClient,
+          ),
```
danieljbruce, line 563: This now supports the use case where the universe domain is provided in the handwritten client options as well as maintaining support for when the universe domain is provided just for the gapic client.

danieljbruce: Note that this should be BigtableInstanceAdminClient as this is a setting for the instance admin client.

src/index.ts
```diff
@@ -104,28 +104,80 @@ export interface BigtableOptions extends gax.GoogleAuthOptions {
...
+ * Otherwise, the function returns `undefined`.
+ *
+ * @param {BigtableOptions} options - The Bigtable client options.
+ * @param {gax.ClientOptions} [opts] - Optional gax client options.
+ * @returns {string | undefined} The universe domain, or `undefined` if not found.
  */
-function getDomain(prefix: string, opts?: gax.ClientOptions) {
+function getUniverseDomainOnly(
+  options: BigtableOptions,
+  opts?: gax.ClientOptions,
```
feywind, line 123: It might be good to give these more different names - opts and options are pretty close. I think in pubsub we tend to use gaxOpts, but it's variable.

src/index.ts
```diff
@@ -104,28 +104,80 @@ export interface BigtableOptions extends gax.GoogleAuthOptions {
...
+ * gax.ClientOptions object to determine the universe domain to be used.
+ * It prioritizes the `universeDomain` property in the options, then checks
+ * for `universeDomain` or `universe_domain` in the gax options, and finally
+ * falls back to the `GOOGLE_CLOUD_UNIVERSE_DOMAIN` environment variable.
+ * If a universe domain is found, it returns an object containing the
+ * `universeDomain` property; otherwise, it returns `null`.
+ *
+ * @param {BigtableOptions} options - The Bigtable client options.
+ * @param {gax.ClientOptions} [opts] - Optional gax client options.
+ * @returns {object | null} An object containing the `universeDomain` property if found,
```
feywind, line 153: I'd try to give this a type of some kind - TypeScript will likely assign it `{ universeDomain: string }` but it takes a while to figure that out. (And the JSdoc won't be great.)

feywind: Separate comment, it's good to give all these functions an explicit return type when you can. Even if it's super obvious what it is, if something changes inside the function, it'll help catch it.

system-test/tpc-tests.ts
```diff
@@ -0,0 +1,112 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {after, describe, it} from 'mocha';
+import {Bigtable} from '../src';
+
+describe.skip('TPC tests', () => {
```
feywind, line 18: re: chat comment, I'm not sure if TPC is an accepted term to use for this stuff; might want to double check with @bshaffer.

system-test/tpc-tests.ts
```diff
@@ -0,0 +1,112 @@
...
+  // These tests are only designed to pass when using the service account
+  // credentials for the TPC environment so we skip them in the CI pipeline.
+  //
+  // To see successful tests, uncomment the following line:
+  // process.env.GOOGLE_APPLICATION_CREDENTIALS = '/path-to/tpc_sa_key.json';
+
+  function runTest(done: Mocha.Done, bigtable: Bigtable) {
```
feywind, line 25: I'd make runTest an async function here and remove the inner lambda with async. Down below, you can just make your test functions passed to mocha async and then await on runTest.

system-test/tpc-tests.ts
```diff
@@ -0,0 +1,112 @@
...
+    const options = {
+      universeDomain,
+    };
+    const bigtable = new Bigtable(options);
+    runTest(done, bigtable);
+  });
+
+  it('should set the TPC universe with a client option for the gapic clients', done => {
+    const universeDomain = 'apis-tpczero.goog'; // or your universe domain if not using emulator
+    const options = {
```
feywind, line 98: I know basically nothing at all about bigtable libraries, but it seems odd to me that each of the client has its own options variants. Maybe something we think about later, if they could be combined.

feywind (commented):
Node comments added, I'll wait for @bshaffer to chime in on the approval.

src/index.ts
danieljbruce: Done

system-test/tpc-tests.ts
danieljbruce: Done

system-test/tpc-tests.ts
danieljbruce: This is just one potential option. The user can pass universe domain in here for each individual Gapic client or like one of the other tests they can just pass in the universeDomain for the handwritten client and that universe domain option will be applied to all three. So the combined option is supported.

Do we want to remove this test and eliminate the support for universe domain on the Gapic clients?

system-test/tpc-tests.ts
bshaffer: It is not, please remove it!

src/index.ts
```diff
@@ -104,28 +104,81 @@ export interface BigtableOptions extends gax.GoogleAuthOptions {
...
-    opts?.universeDomain ??
-    opts?.universe_domain ??
-    universeDomainEnvVar ??
-    'googleapis.com'
-  }`;
+  return (
+    options?.universeDomain ??
+    gaxOpts?.universeDomain ??
+    gaxOpts?.universe_domain ??
+    universeDomainEnvVar
```
bshaffer, line 136: what is the difference between `options.universeDomain` and `gaxOpts.universeDomain` from a user perspective? What does it look like when one overrides the other?

src/index.ts
bshaffer: Why not just be opinionated here and pass in `getDomain`?

bshaffer (commented):
I am a bit confused by these changes. 

1. It seems like for the GAPIC clients, the universe domain needs to be set in two places - in the `servicePath` and the `options`. Why not just let the GAPIC determine their own `servicePath`? If this is just keeping with the existing paradigm, then that's fine
2. The `adminOptions` and `instanceAdminOptions` are formed by merging the options from the existing `BigtableTableAdminClient` and `BigtableInstanceAdminClient`. Why? It seems circular to me - we create the options for those GAPICs and then save the options back from the GAPICs?

system-test/tpc-tests.ts
danieljbruce: This was done earlier

src/index.ts
danieljbruce: This was done. See "Outdated".

src/index.ts
```diff
@@ -104,28 +104,81 @@ export interface BigtableOptions extends gax.GoogleAuthOptions {
...
-  return `${prefix}.${
-    opts?.universeDomain ??
-    opts?.universe_domain ??
-    universeDomainEnvVar ??
-    'googleapis.com'
-  }`;
+  return (
+    options?.universeDomain ??
+    gaxOpts?.universeDomain ??
+    gaxOpts?.universe_domain ??
```
bshaffer, line 135: Why do we check the snake-case `universe_domain` for the `gaxOpts` but not for the `options` opts?

system-test/tpc-tests.ts
feywind: Ah, I see. Thanks for the clarification.

src/index.ts
feywind: This something that Alex decided on a while back. I'm not entirely sure why we do it, but we do it in generated gapic clients too, so it's in all of the libraries. (At least the ones I've seen.)

src/index.ts
danieljbruce: `gaxOpts` is the settings the user provided for the specific Gapic client and `options` is the options provided to the handwritten layer. We discussed offline that we should really allow the gapic options to take priority over the handwritten layer options so I can apply the change to make gaxOpts?.universeDomain come first.

src/index.ts
bshaffer: Should it be added to `options` then as well? e.g.

```
  return (
    options?.universeDomain ??
    options?.universe_domain ??
    gaxOpts?.universeDomain ??
    gaxOpts?.universe_domain ??
    universeDomainEnvVar ??
 }
```

Or should snake-case be marked deprecated so that newer clients (e.g. this one) can ignore it?

src/index.ts
```diff
@@ -104,28 +104,81 @@ export interface BigtableOptions extends gax.GoogleAuthOptions {
...
+/**
+ * Retrieves the universe domain options from the provided options.
+ *
+ * This function examines the provided BigtableOptions and an optional
+ * gax.ClientOptions object to determine the universe domain to be used.
+ * It prioritizes the `universeDomain` property in the options, then checks
+ * for `universeDomain` or `universe_domain` in the gax options, and finally
+ * falls back to the `GOOGLE_CLOUD_UNIVERSE_DOMAIN` environment variable.
+ * If a universe domain is found, it returns an object containing the
+ * `universeDomain` property; otherwise, it returns `null`.
```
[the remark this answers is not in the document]
danieljbruce: The reason getUniverseDomainOptions returns null is that the return value is used in an `Object.assign` call. The behaviour of Object.assign is that it does nothing to the object being assigned, in this case, the Gapic client options if the object being assigned to it is null.

So the behaviour we want that we are getting with this code is don't modify the options when null gets returned.

src/index.ts
bshaffer: You don't need to pass the universe domain option through with the Env Var because that'll be done by the GAPICs. Even though you need it to calculate the servicePath, you don't need to set it as the UniverseDomain option as well. 

I would suggest taking the path of either 1. "configure the bare minimum and let the GAPICs do the rest" or 2. "always explicitly set it so the GAPICs do nothing". So this would be, either always ensure the `universeDomain` option is set (make it `googleapis.com` if not), or make it minimally set (so that it would be `null` unless explicitly configured).

1. `getUniverseDomain` _always_ returns a string, and is used in `getDomain` to append the service name
2. `getUniverseDomainOption` _only_ returns a string if a universe domain option was explicitly configured in the client options or GAPIC client options

src/index.ts
danieljbruce: Then it would be set to `bigtable.${universeDomain}` where universeDomain is the value provided in the Gapic client.

src/index.ts
danieljbruce: Ok. gaxOpts?.universeDomain comes first now

danieljbruce:
> I am a bit confused by these changes.
> 
> 1. It seems like for the GAPIC clients, the universe domain needs to be set in two places - in the `servicePath` and the `options`. Why not just let the GAPIC determine their own `servicePath`? If this is just keeping with the existing paradigm, then that's fine
> 2. The `adminOptions` and `instanceAdminOptions` are formed by merging the options from the existing `BigtableTableAdminClient` and `BigtableInstanceAdminClient`. Why? It seems circular to me - we create the options for those GAPICs and then save the options back from the GAPICs?

In response to 1. Because as I said in one of the PR comments, if we don't set universeDomain in the options then the test fails! When universe domain is set in a client option or using an environment variable by the user then universe domain has to be set on the gapic client options for the test to pass. Try it out yourself.

In response to 2. The client library worked this way before and I don't want to change the way it works because the minimal amount of changes necessary to support universe domains is ideal. I think these values adminOptions and dataOptions are saved so that they are ready when a gapic client gets initialized just before a request is made. If you want to change the way gapic options are handled in nodejs-bigtable then I would suggest filing a separate bug.

feywind:
> In response to 2. The client library worked this way before and I don't want to change the way it works because the minimal amount of changes necessary to support universe domains is ideal.

I'm behind this idea. If it was already that way before, let's leave it for now and circle back if it's weird.

feywind (dismissed):

feywind (approved):

bshaffer (approved):

PR #1605, opened 2025-05-13
fix: Get rid of the global singletons thereby improving client side metrics design

danieljbruce:
## Description

This change eliminates the instrumentsForProject global singleton and the metricsHandlers global singleton replacing it with a ClientSideMetricsConfigManager object. This allows us to pass down the credentials from the client library to the metric service client so that the MetricServiceClient always authenticates correctly when custom credentials are supplied.

## Impact

This will ensure that no matter if custom project settings are supplied or default project settings are supplied that the MetricServiceClient will authenticate correctly. It also encapsulates work that was previously done by some global singletons for better factoring.

## Testing

Since some items moved around, mocks of those items needed to change. Tests have been added for the ClientSideMetricsConfigManager class.

## Additional Information

We can further explore other design suggestions like:
- Not passing the project through the OTEL Resource and passing it as a parameter instead
- Passing projectId to the Metrics Collector earlier in the lifecycle


src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -211,26 +211,29 @@ export class OperationMetricsCollector {
...
-              metricsCollectorData: this.getMetricsCollectorData(),
-              projectId,
-            });
-          }
-        });
+        const metricsHandler =
+          ClientSideMetricsConfigManager.getGcpHandlerForProject(
+            projectId,
+            this.tabularApiSurface.bigtable.options as ClientOptions,
+          );
```
daniel-sanche, line 222: I still really don't like getting the handler this way. It's a really unintuitive coupling, calling out to a static class to grab a singleton on each attempt/operation completion. It would be to describe the relationship between these classes to an outsider. I think these classes should have a more conventional coupling

daniel-sanche (commented):
I wanted to leave some of my thoughts on this before the end of the day, but I don't have a full solution in mind yet. I'm going to look at it more tomorrow

I still think we should aim to have the MetricsHandlers live on the client, and be created at init time. It seems like we're real close, and that would simplify a lot of things, and give us a better architecture going forward. Just need to figure out how to handle that projectId, so we can cache the exporters to avoid the rate limit issue

src/client-side-metrics/operation-metrics-collector-factory.ts
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {MethodName, StreamingState} from './client-side-metrics-attributes';
+import {
+  ITabularApiSurface,
+  OperationMetricsCollector,
+} from './operation-metrics-collector';
+
+export class OperationMetricsCollectorFactory {
```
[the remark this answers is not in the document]
danieljbruce: Instead of just a function, I think I can made this a class so that it could be easily mocked out with proxyquire.

https://github.com/googleapis/nodejs-bigtable/pull/1605/files#diff-0db6bc878429294fec782c3aa5f803212e2a6cb42ddc0c61fef87b59142b9c50R69

I can try to make our tests work with it as a function, but I don't know if that's going to be possible.

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {CloudMonitoringExporter} from './exporter'; // Added import
```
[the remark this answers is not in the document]
danieljbruce: Removed comment and other "Added" comments.

src/client-side-metrics/operation-metrics-collector.ts
danieljbruce: I understand your concern that calling a static class like this isn't great and it isn't the most modular, but at some point in the code it has to be done because there is a global map from projectIds to instrument stacks that is a shared resource between clients so either we use it earlier, like in the method call or we use it later.

We could move the manager object back to its position on the Bigtable client like you had in your design sketch, but the thing I don't like about this is that it exposes another variable on the client. What do you think?

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -216,13 +215,10 @@ export class GCPMetricsHandler implements IMetricsHandler {
...
-    if (!GCPMetricsHandler.instrumentsForProject[projectId]) {
-      GCPMetricsHandler.instrumentsForProject[projectId] = createInstruments(
-        projectId,
-        this.exporter,
-      );
+    if (!this.otelInstruments) {
+      this.otelInstruments = createInstruments(projectId, this.exporter);
     }
-    return GCPMetricsHandler.instrumentsForProject[projectId];
+    return this.otelInstruments;
```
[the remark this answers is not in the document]
danieljbruce: Yes. I made the change to provide the projectId at constructor time and then fix the tests. ProjectId won't be provided for each operation anymore.

src/client-side-metrics/metrics-config-manager.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  static getGcpHandlerForProject(
+    projectId: string,
+    options: ClientOptions,
+  ): GCPMetricsHandler {
+    // share a single GCPMetricsHandler for each project, to avoid sampling errors
+    if (this.gcpHandlerStore.has(projectId)) {
+      return this.gcpHandlerStore.get(projectId)!;
+    } else {
+      const newHandler = new GCPMetricsHandler(options);
```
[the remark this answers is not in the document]
danieljbruce: This is the same suggestion as the following quote:

> One thing I don't like about this: each operation has it's own projectId, and it uses that to call getInstruments(project_id). But getInstruments only uses the projectId once, the first time it creates the instruments, and then it always returns the same cached instruments.

I pass projectId in the constructor now.

danieljbruce:
> I wanted to leave some of my thoughts on this before the end of the day, but I don't have a full solution in mind yet. I'm going to look at it more tomorrow
> 
> I still think we should aim to have the MetricsHandlers live on the client, and be created at init time. It seems like we're real close, and that would simplify a lot of things, and give us a better architecture going forward. Just need to figure out how to handle that projectId, so we can cache the exporters to avoid the rate limit issue

> I still think we should aim to have the MetricsHandlers live on the client

Doesn't this create some data duplication though with the metrics handlers that live on the client + the singleton map that stores the metrics handlers?

> be created at init time

If we create the metrics handlers at client initialization time then we have to do an asynchronous fetch to get the projectId. We don't know when this operation will be complete. Even if this isn't a problem now, if there is any time in the future as we modify and maintain the codebase that we are counting on this async fetch to be done on time then that could be a bug that  is really hard to find and solve.

Just a reminder since it may help with our discussion. Here are the order of operations:

1. Initialize client
2. Get instance from client, table from instance
3. Make method call from table
4. Initialize metrics collector
5. Make request which initializes project <- *we get the project here*
6. Call metrics collector method (onAttemptComplete)
7. Call getGcpHandlerForProject <- *projectId needed here to fetch instrument stack*
8. Initialize metrics handler passing in project 
9. Metrics handler creates exporter
10. Exporter is used to create meter provider
11. Instrument stack gets created from meter provider
12. getGcpHandlerForProject returns GCPHandler which has the instrument stack
13. onAttemptComplete is called and metrics are recorded on instrument stack
14. 30 seconds later, metrics automatically sent to exporter

We can't push Step 7 to before Step 5.

src/client-side-metrics/gcp-metrics-handler.ts
```diff
@@ -196,33 +199,25 @@ export class GCPMetricsHandler implements IMetricsHandler {
...
    *
-   * @param exporter - The `PushMetricExporter` instance to use for exporting
-   *   metrics to Google Cloud Monitoring. This exporter is responsible for
-   *   sending the collected metrics data to the monitoring backend. The provided exporter must be fully configured, for example the projectId must have been set.
    */
-  constructor(exporter: PushMetricExporter) {
-    this.exporter = exporter;
+  constructor(projectId: string, options: ClientOptions) {
+    this.options = options;
+    this.projectId = projectId;
```
daniel-sanche, line 205: This isn't going to work in the same was as the code I sent you. client.projectId can be the placeholder `{{projectId}}` value before the first rpc is sent, but it will be replaced by the real value by the time data arrives. If we store the `client` object, we can grab the projectId onOperationComplete. But if we store a string here, we may be stuck with the placeholder

src/client-side-metrics/operation-metrics-collector-factory.ts
```diff
@@ -0,0 +1,33 @@
...
+  static createOperation(
+    methodName: MethodName,
+    streaming: StreamingState,
+    table: ITabularApiSurface,
+  ): OperationMetricsCollector | undefined {
+    if (table.bigtable.metricsEnabled) {
+      return new OperationMetricsCollector(table, methodName, streaming);
+    } else {
+      return;
```
daniel-sanche, line 30: I think we should always write to the OperationMetricsCollector, but just do nothing if no handlers are attached. That makes the testing easier, since we don't have to worry about the two cases, optional values, etc. Let me know what you think

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -211,26 +211,27 @@ export class OperationMetricsCollector {
...
-              client_name: `nodejs-bigtable/${version}`,
-              metricsCollectorData: this.getMetricsCollectorData(),
-              projectId,
-            });
-          }
-        });
+        const projectId = this.tabularApiSurface.bigtable.projectId as string;
+        const options = this.tabularApiSurface.bigtable
+          .options as ClientOptions;
+        const metricsHandler = new GCPMetricsHandler(projectId, options);
```
daniel-sanche, line 221: I really don't want us creating new handlers each time onAttemptComplete is called. 

That's a lot of extra work to do on each rpc, and we should have all the data we need to do this once per client, and share them

src/client-side-metrics/operation-metrics-collector-factory.ts
danieljbruce: It makes testing easier, but also means we run metrics collector code for no reason increasing latency.

src/client-side-metrics/gcp-metrics-handler.ts
danieljbruce: Yes. We will be stuck with the placeholder if we create this object too early. So if we decide not to create a metrics handler in the metrics collector and pass it in instead then this needs to change.

src/index.ts
```diff
@@ -537,10 +539,12 @@ export class Bigtable {
...
 
-    if (options.metricsEnabled === false) {
-      this.metricsEnabled = false;
+    if (options.metricsEnabled === true) {
+      this._metricsConfigManager = new ClientSideMetricsConfigManager([
+        new GCPMetricsHandler(options as ClientOptions),
+      ]);
     } else {
-      this.metricsEnabled = true;
+      this._metricsConfigManager = new ClientSideMetricsConfigManager([]);
```
daniel-sanche, line 547: not: looking at this again, maybe we should structure this like this?

```
if (options.metricsEnabled === true) {
     metric_handlers = [new GCPMetricsHandler(options as ClientOptions)]
} else {
    metric_handlers = []
}
this._metricsConfigManager = new ClientSideMetricsConfigManager(metric_handlers);
```

daniel-sanche (approved):
Looks good! Thanks for putting this together!

src/index.ts
mutianf, line 547: why do we still create the MetricsConfigManager when metrics are not enabled? I think we can bypass it entirely.

danieljbruce: We should loop @daniel-sanche in this discussion. My original design would ignore any metrics collection if metrics aren't enabled, but [this discussion](https://github.com/googleapis/nodejs-bigtable/pull/1605#discussion_r2091929506) makes a case for a different option.

By the way I think we should just merge this PR and continue comments on https://github.com/googleapis/nodejs-bigtable/pull/1571. This PR was just meant to separate some changes Daniel and I wanted to talk about from the rest of the readRows stuff, but now that he has approved it will be easier to discuss everything in the other PR.

danieljbruce: Yup. Done. Simple change.

PR #1606, opened 2025-05-21
test: Add a test that confirms errors are properly passed through for getInstance calls

danieljbruce:
## Description

This test confirms that errors will bubble up to the user for getInstances calls which allows us to close b/388499540 as not reproducible.

## Impact

Just adding a test to confirm correct client library behaviour

## Additional Information

Two files were edited:
-  One is the test file where we added the new test
-  The other is the file that defines mock services where we needed to add a mock service for the admin client


test/errors.ts
```diff
@@ -147,3 +150,56 @@ describe('Bigtable/Errors', () => {
...
+    bigtable = new Bigtable({
+      apiEndpoint: `localhost:${port}`,
+    });
+    service = new BigtableAdminClientMockService(server);
+  });
+
+  describe('with getInstances', () => {
+    const emitGetInstancesError = (stream: any) => {
+      const metadata = new grpc.Metadata();
+      metadata.set(
```
feywind, line 173: Might add a comment here about what the magic buffer numbers mean.

feywind (dismissed):

cindy-peng (approved):

PR #1612, opened 2025-06-02
feat: Record READ_ROW metrics for single row calls

danieljbruce:
## Description

Mattie pointed out an important requirement [here](https://github.com/googleapis/nodejs-bigtable/pull/1571#discussion_r2116388538) which needs this change in place. This PR ensures the right client side metrics are collected for the right methods.

## Impact

User calls with `row.get` will now write client side metrics for `READ_ROW` instead of `READ_ROWS` as required. This change ensures the right metrics are collected for the right methods.

## Testing

- A whole bunch of the mocks needed to change as is usually the case when you move source code around
- Added tests to ensure the metrics are being written properly for READ_ROW case

## Additional Information


src/tabular-api-surface.ts
```diff
@@ -212,376 +203,10 @@ Please use the format 'prezzy' or '${instance.name}/tables/prezzy'.`);
...
-        .on('end', () => {
-          activeRequestStream = null;
-          metricsCollector.onOperationComplete(grpc.status.OK);
-        });
-      rowStreamPipe(rowStream, userStream);
-    };
-
-    makeNewRequest();
-    return userStream;
+    return createReadStreamInternal(this, true, opts);
```
mutianf, line 209: why is this true? I think createReadStream is for scan not reading a single row?

src/utils/createReadStreamInternal.ts
```diff
@@ -0,0 +1,431 @@
...
+ * @param {string} [opts.prefix] Prefix that the row key must match.
+ * @param {string[]} [opts.prefixes] List of prefixes that a row key must
+ *     match.
+ * @param {object[]} [opts.ranges] A list of key ranges.
+ * @param {string} [opts.start] Start value for key range.
+ * @param {string} [viewName] The name of the authorized view, if applicable.
+ * @returns {stream} A readable stream of {@link Row} objects.
+ *
+ */
+export function createReadStreamInternal(
```
mutianf, line 61: I didn't look at this file in detail except for changes related to `singleRow`. I assume everything is copy pasted from createReadStream. Let me know if there are other lines i should take a look.

src/utils/getRowsInternal.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @param {function} cb The callback function.
+ * @param {?error} cb.err An error returned while making this request.
+ * @param {Row[]} cb.rows List of Row objects.
+ *
+ * @returns {Promise<GetRowsResponse>|void} Returns a promise that resolves with the rows if no callback is provided, otherwise calls the callback with the rows.
+ *
+ * @example <caption>include:samples/api-reference-doc-snippets/table.js</caption>
+ * region_tag:bigtable_api_get_rows
+ */
+export function getRowsInternal(
```
mutianf, line 35: why do we need getRowsInternal? Can we call createReadStreamInternal directly?

danieljbruce: row.get used to call getRows. Since getRows calls createReadStream for both single rows and multi row calls then getRows needs a `singleRow` parameter so should be pulled out into getRowsInternal.

In `row.get` we should not call `createReadStreamInternal` because then we need to rewrite the `getRows` function inside `row.get` so that `row.get` does what it did before. This will make it so that we have to implement the `getRows` logic in two places which isn't ideal.

danieljbruce:
@mutianf Thanks for the reviews. The main purpose of this PR was to provide reviews for one particular change in isolation and get some eyes on it before applying it to all the readRows work. Now that we agree the changes are directionally correct, I think its fine if we merge this into the ReadRows branch and continue review on this code there to unblock progress.

PR #1624, opened 2025-06-25
test: Remove sinon clock dependency to unblock the CI pipeline

danieljbruce:
## Description

Two days ago the system tests suddenly started failing on main even though no changes were made to the codebase. In the tests, the line of code that says clock.runAll(); would wait for all the events to be emitted and processed by the client library so that all the variables used in assertion checks would have the right values and the assertion checks would pass. Suddenly the clock.runAll(); line of code doesn't wait for the events to be processed by the client library so the assertion checks fail. It also seems that even when the assertion checks are omitted the test passes, but the events still don't get processed.

The test framework is changed so that we don't rely on the sinon clock dependency anymore because an update to this dependency seems to have broke the test. Instead, we do all the assertion checks when an error or end event passes through the client library and reaches the test runner since this is when we were doing the assertion checks before anyway when the sinon clock was working properly.

## Impact

Now system tests will pass so we can develop on nodejs-bigtable again.

## Testing

Modified a test. This doesn't reduce test coverage. It just removes the dependency on the sinon clock.

## Additional Information

Before these changes 9 of the tests would fail. The eventemitter was emitting the events, but for some reason the events were never reaching the chunk transformer in the client library.


system-test/read-rows.ts
```diff
@@ -133,18 +133,6 @@ describe('Bigtable/Table', () => {
     let stub: sinon.SinonStub;
 
     beforeEach(() => {
-      clock = sinon.useFakeTimers({
-        toFake: [
-          'setTimeout',
-          'clearTimeout',
-          'setImmediate',
```
feywind, line 140: It's possible you could just remove setImmediate and clearImmediate here (and also probably nextTick) and that would also fix it. I seem to remember having troubles with this same issue on Pub/Sub, now that you mention it, because they started hooking these by default.

system-test/read-rows.ts
```diff
@@ -186,34 +174,46 @@ describe('Bigtable/Table', () => {
     });
 
     afterEach(() => {
-      clock.restore();
       stub.restore();
     });
 
     tests.forEach(test => {
-      it(test.name, () => {
+      it(test.name, (done: mocha.Done) => {
```
feywind, line 181: If you want to avoid the extra mocha dependency, you can also just do `done => {` here. TS will figure it out.

daniel-sanche (approved):

system-test/read-rows.ts
feywind: [here](https://github.com/googleapis/nodejs-pubsub/pull/2024/files#diff-2cf01d03222a34294066e6c173aea4831d26113df10f6a8969dc29a3f97cb467)

feywind (approved):

PR #1631, opened 2025-07-03
feat: Add plumbing to support unary calls for client side metric collection

danieljbruce:
## Description

This PR adds source code changes with plumbing so that client side metric collection can be done for unary calls. No client side metrics are actually collected yet for readModifyWriteRow or checkAndMutate calls, but future PRs will use this plumbing to collect metrics for those calls. A test is provided that makes a fake method on the table to demonstrate how a method would collect these metrics to serve as an example.

## Impact

Doesn't change the way the client library works. Just provides support for when we open a PR that collects client side metrics for unary calls.

## Testing

A unit test is created that adds a fake method to a table that does a basic unary readModifyWriteRow call. It ensures that the metrics make it to the test metrics handler correctly.

## Additional Information

**Changes:**

`src/interceptor.ts`: This file contains code for attaching interceptors to a call that will call the correct metrics collector methods when data arrives

`src/client-side-metrics/operation-metrics-collector.ts`: The types are cleaned up in this file. None of the functionality has changed in this file

`system-test/read-modify-write-row-interceptors.ts`: This adds a test that ensures the plumbing for collecting metadata trailers and headers works as intended for unary calls.


danieljbruce:
We should merge https://github.com/googleapis/nodejs-bigtable/pull/1634 first. It will resolve the system-test issues.

protos/protos.d.ts
```diff
@@ -28903,6 +28903,9 @@ export namespace google {
 
             /** CommonLanguageSettings destinations */
             destinations?: (google.api.ClientLibraryDestination[]|null);
+
+            /** CommonLanguageSettings selectiveGapicGeneration */
+            selectiveGapicGeneration?: (google.api.ISelectiveGapicGeneration|null);
```
daniel-sanche, line 28908: are these protos supposed to be part of the PR?

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -323,17 +317,20 @@ export class OperationMetricsCollector {
...
       // Check serverTimeRead, connectivityErrorCount here to reduce latency.
       const mappedEntries = new Map(
-        Array.from(metadata.internalRepr.entries(), ([key, value]) => [
-          key,
-          value.toString(),
-        ]),
+        // Note: The cast on metadata is required to satisfy the compiler.
+        // In practice the internalRepr property is available for us to read.
+        // But the internalRepr is protected on the Metadata type so we need
+        // to cast.
```
daniel-sanche, line 327: is accessing the internalRepr necessary? Can you link to the class definition?

src/interceptor.ts
```diff
@@ -0,0 +1,85 @@
...
+        next(metadata, newListener);
+      },
+      sendMessage: (message, next) => next(message),
+      halfClose: next => next(),
+      cancel: next => next(),
+    });
+  };
+}
+
+export function withInterceptors(
```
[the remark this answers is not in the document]
danieljbruce: In a method call we write something like this:

```
        bigtable.request(
          {
            client: 'BigtableClient',
            method: 'readModifyWriteRow',
            reqOpts: {
              tableName: table.name,
              rowKey: Buffer.from(ROW_KEY),
              rules: [
                {
                  familyName: COLUMN_FAMILY,
                  columnQualifier: Buffer.from(COLUMN),
                  appendValue: Buffer.from('-wood'),
                },
              ],
              appProfileId: undefined,
            },
            gaxOpts: **_withInterceptors_**({}, metricsCollector),
          },
          (err: ServiceError | null, resp?: any) => {
            if (err) {
              reject(err);
            } else {
              resolve(resp);
            }
          },
        );
```

system-test/read-modify-write-row-interceptors.ts
```diff
@@ -0,0 +1,294 @@
...
+      attemptCompleteData.metricsCollectorData.table,
+      TABLE_ID,
+    );
+    assert.strictEqual(
+      attemptCompleteData.metricsCollectorData.instanceId,
+      INSTANCE_ID,
+    );
+    assert.ok(attemptCompleteData.attemptLatency >= 0);
+    assert(attemptCompleteData.serverLatency);
+    assert.ok(attemptCompleteData.serverLatency >= 0);
```
danieljbruce, line 264: This is where we prove we are getting the server latency from the metadata headers.

system-test/read-modify-write-row-interceptors.ts
```diff
@@ -0,0 +1,294 @@
...
+      operationCompleteData.metricsCollectorData.instanceId,
+      INSTANCE_ID,
+    );
+    assert.ok(operationCompleteData.operationLatency >= 0);
+    assert.strictEqual(operationCompleteData.retryCount, 0);
+    assert.strictEqual(operationCompleteData.metricsCollectorData.zone, ZONE);
+    assert.strictEqual(
+      operationCompleteData.metricsCollectorData.cluster,
+      CLUSTER,
+    );
```
danieljbruce, line 291: This is where we prove we are getting the Zone/Cluster from the trailer.

src/interceptor.ts
```diff
@@ -0,0 +1,91 @@
...
+    const interceptor = createMetricsInterceptorProvider(metricsCollector);
+    if (!gaxOptions.otherArgs) {
+      gaxOptions.otherArgs = {};
+    }
+    if (!gaxOptions.otherArgs.options) {
+      gaxOptions.otherArgs.options = {};
+    }
+    if (!gaxOptions.otherArgs.options.interceptors) {
+      gaxOptions.otherArgs.options.interceptors = [interceptor];
+    }
```
daniel-sanche, line 88: don't you need to add the interceptor, even if other interceptors were added?

src/interceptor.ts
```diff
@@ -0,0 +1,91 @@
...
+            );
+            nextMd(metadata);
+          },
+          onReceiveMessage: (message, nextMsg) => {
+            savedReceiveMessage = message; // Still need to know if a message was received for onResponse
+            nextMsg(message);
+          },
+          onReceiveStatus: (status, nextStat) => {
+            if (status.code === GrpcStatus.OK && savedReceiveMessage) {
+              collector.onResponse(); // Call onResponse for successful unary calls with a message
```
daniel-sanche, line 56: is this the right place for onResponse? Why not put this in onReceiveMessage?

Also, isn't onResponse only used by read rows? should createMetricsInterceptorProvider have a bool argument indicating whether to call onResponse?

daniel-sanche (commented):
Looking good for the most part, but I have a few small comments

src/interceptor.ts
danieljbruce: That really depends on how we want to design the system. Here are the pros and cons of adding the interceptor.

Pros:
- Existing users that are already using interceptors, likely a small minority of our users, will automatically be using the new interceptor now so they don't have to add the CSM interceptor to their interceptors
Cons:
- It means the user doesn't have full control of the interceptors which reduces flexibility

Yeah, I guess the pros outweigh the cons so we can add this change.

danieljbruce: Actually, I'm not sure I like this change anymore. If the interceptors are not an array (although they should be) then trying to push another interceptor onto interceptors will throw an error. I don't want users to experience that just in case they are using interceptors incorrectly. As you can see in the screenshot, the compiler doesn't guarantee any particular structure for the interceptors property.

image

src/interceptor.ts
danieljbruce: > Also, isn't onResponse only used by read rows?

This is a good point. I don't think we need the call to onResponse at all because this interceptor should only be used for unary calls. For streaming calls like readRows, if we really did want to call onResponse earlier like right when the client receives the message rather than after it passes through gax then we should have a different interceptor that doesn't do the status and metadata parse. Unless we plan on making changes to the streaming calls to change the way they do status and metadata parsing?

My suggestion is to remove the onResponse call for now since the intent is to only use these interceptors for unary calls.

Then as a backlog item we can decide:
1. If streaming calls should use interceptors as well. If not no changes are required
2. I did an experiment and confirmed that onReceiveMessage seems to be the right place to do the onResponseCall
3. Make the withInterceptors method accept a parameter that picks the right interceptor. ie. if ReadRows then call onResponse

src/interceptor.ts
daniel-sanche: Just to make sure I'm understanding this right: with the current code, if the user passes in their own interceptors, they would effectively be disabling csm, right?

What else could they be passing other than an array? There must be some docstring somewhere telling us what we need to support

src/interceptor.ts
daniel-sanche: Ok, yeah IIRC onResponse is only for read_rows, so probably don't need it here

I think interceptors are optional for streams. If it would make the code more clean, and let us strip out some extra logic, we should go for it. But if it's easier to grab the metadata using wrappers, that's fine too

src/interceptor.ts
```diff
@@ -0,0 +1,88 @@
...
+            collector.onMetadataReceived(
+              metadata as unknown as {
+                internalRepr: Map<string, string[]>;
+                options: {};
+              },
+            );
+            nextMd(metadata);
+          },
+          onReceiveMessage: (message, nextMsg) => {
+            savedReceiveMessage = message; // Still need to know if a message was received for onResponse
```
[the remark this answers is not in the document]
danieljbruce: Good catch. This variable has been removed now.

src/interceptor.ts
danieljbruce: All the [code examples](https://github.com/grpc/grpc-node/blob/01db2bc6206963a6678ee05fa4651fd2ee40068c/examples/interceptors/client.js#L108) pass interceptors in as an array. If you feel strongly about adding it to the existing interceptors then we can make that change, but I just don't like how the compiler doesn't guarantee that it is an array.

src/interceptor.ts
daniel-sanche: Was that last comment meant for the other thread, where we were discussing Partial Errors?

I don't think that section of code tells us much on its own though. Where is the error being handled?

src/interceptor.ts
daniel-sanche: Can we check the type of `gaxOptions.otherArgs.options.interceptors`, and only append if it is a list? And ignore unexpected input types?

I really don't think we should silently disable CSM if the user passes in an interceptor. That would present as a bug

src/interceptor.ts
```diff
@@ -0,0 +1,86 @@
...
+              metadata as unknown as {
+                internalRepr: Map<string, string[]>;
+                options: {};
+              },
+            );
+            nextMd(metadata);
+          },
+          onReceiveMessage: (message, nextMsg) => {
+            nextMsg(message);
+          },
```
daniel-sanche, line 51: is this block actually needed? If we left this off, doesn't it just call `nextMesg(message)` automatically?

src/interceptor.ts
```diff
@@ -0,0 +1,86 @@
...
+            );
+            // AttemptComplete and OperationComplete will be called by the calling code
+            nextStat(status);
+          },
+        };
+        next(metadata, newListener);
+      },
+      sendMessage: (message, next) => next(message),
+      halfClose: next => next(),
+      cancel: next => next(),
```
daniel-sanche, line 64: are these unimplemented callbacks needed?

daniel-sanche (commented):
I think my only remaining concern is that users passing in interceptors would disable CSM. 

We need to either document that and explain how to manually pass the CSM interceptor, or make sure we append the interceptor to whatever the user passes in

src/interceptor.ts
danieljbruce: Yes. I tested it and we can ignore this block.

src/interceptor.ts
danieljbruce: No. These aren't necessary either. If you remove them then messages still get sent etc. They were just copy/paste from the logging interceptor code.

src/interceptor.ts
danieljbruce: > Was that last comment meant for the other thread, where we were discussing Partial Errors?

Yes I think so. I see it contains the link to the same code.

daniel-sanche (approved):
LGTM

PR #1634, opened 2025-07-07
test: The test collecting CSM for a second project should record metrics for the second project

danieljbruce:
## Description

As indicated by [this test run](https://btx.cloud.google.com/invocations/d46a4997-1695-4527-b934-77b7a53e369c/targets/cloud-devrel%2Fclient-libraries%2Fnodejs%2Fpresubmit%2Fgoogleapis%2Fnodejs-bigtable%2Fnode18%2Fsystem-test/log) the CI pipeline is suddenly failing every time on the "should send the metrics to Google Cloud Monitoring for a ReadRows call with a second project" test. This is because the test incorrectly mocks the metric handler for the fake Bigtable client so that it doesn't include the same project as the Bigtable client. In practice this can never happen because options are always passed down from the Bigtable client to the metrics handler to the exporter as confirmed by "should pass the credentials to the exporter" test.

The assertion error occurs when we fetch the client side metrics, but there isn't any series in the returned array. This is because before with the test we were writing metrics with the default project instead of the secondary project:

image

## Impact

Unblocks the CI pipeline.

## Testing

The tests were changed to mock a more realistic situation.


system-test/client-side-metrics.ts
```diff
@@ -205,7 +209,7 @@ function checkSingleRowCall(
  *   Cloud Monitoring.
  */
 async function checkForPublishedMetrics(projectId: string) {
-  const monitoringClient = new MetricServiceClient(); // Correct instantiation
+  const monitoringClient = new MetricServiceClient({projectId}); // Correct instantiation
```
danieljbruce, line 212: The test only passes when the secondary projectId is passed into the metricServiceClient.

daniel-sanche (approved):

PR #1639, opened 2025-07-10
feat: Initial timed stream implementation for application latencies

danieljbruce:
## Description

Adds a timed stream object. This PR adds an implementation for the timed stream object which provides an additional mechanism on top of streams that measures the amount of time the user processes data from the stream. The timed stream can then be used as a tool for measuring application latencies because application latencies measure time processing data. Tests should accurately reflect what the timed stream should and should not do.

## Impact

Basically provides the right plumbing for application latencies measurement. This should allow us to make code changes to each method that make use of the TimedStream and measure application latencies.

## Testing

Unit tests for the timed stream are added to illustrate what the timed stream should do. This ensures the measurements captured by the timed stream will meet the needs required in order to measure application latencies. This mostly includes ensuring the measurement is correct under a variety of circumstances like if the server delays sending data, there is backpressure or if the loop/handlers contains synchronous or asynchronous calls.


test/timed-stream.ts
```diff
@@ -0,0 +1,91 @@
...
+              /* empty */
+            }
+          });
+          timedStream.on('end', () => {
+            // print results
+            try {
+              const totalMilliseconds = timedStream.getTotalDurationMs();
+              console.log(totalMilliseconds);
+              // totalMilliseconds should be around 10 seconds, 1 per row
+              assert(totalMilliseconds > 9000);
```
danieljbruce, line 71: When you check out this code locally this assert fails which is a problem.

daniel-sanche (changes requested):

src/timed-stream.ts
```diff
@@ -0,0 +1,116 @@
...
+    this.on('after_row', this.handleAfterRowRead);
+  }
+
+  /**
+   * read code is called when a row is consumed.
+   */
+  read(size: number) {
+    // calculate the time spent between iterations of read (i.e. processing the stream in a for loop)
+    const chunk = super.read(size);
+    if (chunk) {
```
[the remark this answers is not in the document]
danieljbruce: If you remove this check then a whole bunch of tests actually fail including `should measure the total time accurately for a series of 30 rows with a synchronous call`.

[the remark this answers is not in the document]
danieljbruce: If you eliminate this check then the total time recorded is 59995 instead of 30000 which is wrong. What seems to happen in practice is that chunk is null when using handlers so this mechanism ensures that the read method doesn't add any time when using handlers.

src/timed-stream.ts
```diff
@@ -0,0 +1,116 @@
...
+          options?.transformHook(event, _encoding, callback);
+        }
+        callback(null, event);
+        // Last, run code for time measurement after the transform callback is
+        // invoked. ie. Ensure that the timer is stopped and elapsed time is
+        // recorded.
+        const endTime = process.hrtime.bigint();
+        const duration = endTime - startTimeTransform;
+        this.totalDurationTransform += duration;
+        startTimeTransform = process.hrtime.bigint();
```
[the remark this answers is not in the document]
danieljbruce: I made this a lot cleaner by introducing a StreamTimer class. We shouldn't have this duplicate code.

src/timed-stream.ts
```diff
@@ -0,0 +1,113 @@
...
+ */
+type TimedStreamOptions = TransformOptions & {
+  transformHook?: (
+    event: any,
+    _encoding: BufferEncoding,
+    callback: TransformCallback,
+  ) => void;
+};
+
+class StreamTimer {
```
daniel-sanche, line 31: a docstring would be useful here

src/timed-stream.ts
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * The TimedStream class is used for measuring the time the user spends
+ * processing data from the stream. We need to measure this time for use cases
+ * like measuring the application latencies for client side metrics.
+ */
+export class TimedStream extends PassThrough {
+  private readTimer = new StreamTimer();
+  private transformTimer = new StreamTimer();
+  constructor(options?: TimedStreamOptions) {
+    // highWaterMark of 1 is needed to respond to each row
```
daniel-sanche, line 64: is this still relevant? The code itself is using highWaterMark of 0

daniel-sanche (approved):
Overall LGTM, but take a look at those comments

src/timed-stream.ts
danieljbruce: I'll address this in a [backlog bug](https://b.corp.google.com/issues/433566793) because this is merge ready.

PR #1646, opened 2025-07-24
feat: fix plumbing for application latencies to use timed stream

danieljbruce:
## Description

Application latencies are recorded in the metrics collector now by reading the data that gets collected on the TimedStream object. This PR prepares all the plumbing so that in a follow-up PR, when we pass in the TimedStream for readRows and other endpoints that the TimedStream will send the right application latencies measurements to Google Cloud monitoring dashboard.

## Impact

This PR means that in a follow-up PR, all we have to do is pass in the timed stream that the user has access to into the metrics collector and we can expect this timed stream to send the application blocking latencies to the Google Cloud Monitoring dashboard correctly.

## Testing

`test-common/expected-otel-export-input.ts`: The numbers for the test fixtures in this file have changed slightly because they depend on the values in metrics-handler-fixture.ts and those values have changed.

`test-common/metrics-handler-fixture.ts`: The numbers have changed for the metrics collected in this file because test/metrics-collector/metrics-collector.ts test simulates events that now exclude the event that occurs with a call to onRowReachesUser.

`test/metrics-collector/metrics-collector.ts`: This test has changed so that the fakeMethod doesn't call onRowReachesUser anymore because onRowReachesUser doesn't exist and instead a fake application latencies value is provided by a fake user stream.

`test/metrics-collector/typical-method-call.txt`: This fixture needs to change because it records the order of events that have occurred in `test/metrics-collector/metrics-collector.ts` and the order of events that have occurred in `test/metrics-collector/metrics-collector.ts` has changed.

`test/timed-stream.ts`: The user stream in this method now explicitly calls a callback in its transformHook method.

## Additional Information

Next steps:

- For readRows open a PR that sends the user stream to the metrics collector for readRows calls so that application latencies get recorded.


src/timed-stream.ts
```diff
@@ -79,7 +80,6 @@ export class TimedStream extends PassThrough {
         if (options?.transformHook) {
           options?.transformHook(event, _encoding, callback);
         }
-        callback(null, event);
```
danieljbruce, line 82: In the userStreams the callback of the transform function is used in several places and sometimes does not pass an event. So this change relies on the `transformHook` to use the callback in the right places and offers us the flexibility that we need.

src/timed-stream.ts
```diff
@@ -65,7 +65,8 @@ export class TimedStream extends PassThrough {
     super({
       ...options,
       objectMode: true,
-      highWaterMark: 0,
+      readableHighWaterMark: 0, // We need to disable readside buffering to allow for acceptable behavior when the end user cancels the stream early.
+      writableHighWaterMark: 0, // We need to disable writeside buffering because in nodejs 14 the call to _transform happens after write buffering. This creates problems for tracking the last seen row key.
```
danieljbruce, line 69: I think this doesn't change the way the stream works, but its nicer if we have the comments here that we had for readRows.

test/timed-stream.ts
```diff
@@ -24,13 +24,23 @@ function* numberGenerator(n: number) {
...
+class UserStream extends TimedStream {
+  constructor() {
+    super({
+      transformHook(event, _encoding, callback) {
+        callback(null, event);
+      },
+    });
+  }
+}
+
```
danieljbruce, line 36: This is now necessary because we have to tell the TimedStream to use the callback in the transformHook now.

PR #1650, opened 2025-07-29
feat: For client side metrics, record metrics as MUTATE_ROW for single row mutates

danieljbruce:
## Description

Currently when a user calls row.save then client side metrics for those calls will get saved for the MUTATE_ROWS method. We want metrics for these calls to get saved for the MUTATE_ROW method so this PR provides the changes to make that happen.

## Impact

Correctly records client side metrics for single point mutateRows calls.

## Testing

For each test that applies to MutateRows and ReadRows there is now a corresponding test for mutating a single row.

## Additional Information

We didn't want to change the table.mutate interface because that would affect users so we pull table.mutate functionality into a helper function and table.mutate as well as row.save call the helper function.


system-test/client-side-metrics-all-methods.ts
```diff
@@ -191,6 +191,18 @@ function checkMutateRowsCall(
   );
 }
 
+function checkMutateRowCall(
+  projectId: string,
+  requestsHandled: (OnOperationCompleteData | OnAttemptCompleteData)[] = [],
+) {
+  readRowsAssertionCheck(
```
daniel-sanche, line 198: Why readRows? Should this be MetricsAssertionCheck or something?

src/utils/mutateInternal.ts
```diff
@@ -0,0 +1,239 @@
...
+  RETRYABLE_STATUS_CODES,
+  TabularApiSurface,
+} from '../tabular-api-surface';
+import {ServiceError} from 'google-gax';
+import {google} from '../../protos/protos';
+import * as is from 'is';
+import {Mutation} from '../mutation';
+import arrify = require('arrify');
+
+export function mutateInternal(
```
daniel-sanche, line 35: this should probably have [a comment like getRowsInternal](https://github.com/googleapis/nodejs-bigtable/blob/c524df017f4afe75633b2ad9d29d087803629e71/src/utils/getRowsInternal.ts#L27)

daniel-sanche (approved):
Looks good, just a couple small comments

PR #1656, opened 2025-07-30
feat: Add client side metrics for readModifyWriteRow calls

danieljbruce:
## Description

This PR collects client side metrics for readModifyWriteRow calls and collects all the right types of metrics for these unary calls. They leverage the plumbing provided for unary calls to add interceptors to those calls so that we can record metrics for the right zone/cluster/server time.

## Impact

This PR collects client side metrics for readModifyWriteRow calls.

## Testing

For each readRows/mutateRows/mutateRow test we add a corresponding test for readModifyWriteRow.


src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -310,7 +319,11 @@ export class OperationMetricsCollector {
                 client_name: `nodejs-bigtable/${version}`,
                 operationLatency: totalMilliseconds,
                 retryCount: this.attemptCount - 1,
-                firstResponseLatency: this.firstResponseLatency ?? undefined,
+                // Conditionally add the firstResponseLatency property
+                ...(this.firstResponseLatency !== undefined &&
+                  this.firstResponseLatency !== null && {
+                    firstResponseLatency: this.firstResponseLatency,
+                  }),
```
daniel-sanche, line 326: can `this.firstResponseLatency` be undefined? It looks like it's [initialized to be null](https://github.com/googleapis/nodejs-bigtable/blob/21b37fd5bb9dc44c76fdaa7f7b84230ead4d66ee/src/client-side-metrics/operation-metrics-collector.ts#L141)

daniel-sanche: Also, doesn't this conflict with your other PR? As I described this in [there](https://github.com/googleapis/nodejs-bigtable/pull/1658/files#r2244059228), I think it's a better idea to always incude a firstResponseLatency, and let the handler decide when it's meaningful. We're using too many potential types for this

src/interceptor.ts
```diff
@@ -49,11 +49,16 @@ function createMetricsInterceptorProvider(
             collector.onStatusMetadataReceived(
               status as unknown as ServerStatus,
             );
+            collector.onAttemptComplete(status.code);
```
daniel-sanche, line 52: As I mentioned in the meeting, I think it's probably better to only run onAttemptComplete when there's an error code, and let the onOperationComplete finalize everything on success

It doesn't have to be that way, but consider it and let me know what you think

src/row-data-utils.ts
```diff
@@ -193,14 +200,32 @@ class RowDataUtils {
...
+    // 3. Make a unary call with gax options that include interceptors. The
+    // interceptors are built from a method that hooks them up to the
+    // metrics collector
     properties.requestData.bigtable.request<google.bigtable.v2.IReadModifyWriteRowResponse>(
       {
         client: 'BigtableClient',
         method: 'readModifyWriteRow',
         reqOpts,
-        gaxOpts: gaxOptions,
+        gaxOpts: withInterceptors(gaxOptions, metricsCollector),
```
daniel-sanche, line 222: This feedback probably should have come in the previous PR, but seeing this here, I think it the function would probably be better called `withMetricInterceptor`. The current name makes it seem like you use it to attach any interceptor, but it really is metrics-specific

And by the same logic `interceptor.ts` file should probably called `metric-intercetor.ts`, and location in `src/client-side-metrics` with the other csm-specific code

src/row-data-utils.ts
```diff
@@ -193,14 +200,32 @@ class RowDataUtils {
...
+      },
+      (...args) => {
+        const err = args[0];
```
daniel-sanche, line 225: can't this be `(err, ...args) => {` instead?

src/utils/createReadStreamInternal.ts
```diff
@@ -363,6 +363,9 @@ export function createReadStreamInternal(
     rowStream = pumpify.obj([requestStream, chunkTransformer, toRowStream]);
 
     metricsCollector.wrapRequest(requestStream);
+    requestStream.on('data', () => {
+      metricsCollector.onResponse();
+    });
```
daniel-sanche, line 368: Doesn't this conflict with https://github.com/googleapis/nodejs-bigtable/pull/1658?

src/utils/createReadStreamInternal.ts
```diff
@@ -421,9 +424,9 @@ export function createReadStreamInternal(
       .on('end', () => {
         activeRequestStream = null;
         const applicationLatency = userStream.getTotalDurationMs();
-        metricsCollector.onOperationComplete(
+        metricsCollector.onOperationAndAttemptComplete(
           grpc.status.OK,
-          userStream.getTotalDurationMs(),
+          applicationLatency,
```
daniel-sanche, line 429: why is applicationLatency a separate variable in only this case?

daniel-sanche (changes requested):

src/interceptor.ts
danieljbruce: The only trouble is that at this time we don't know if the code is a retryable error or not. That is up to gax to decide. And we will run onOperationComplete when we get non-retryable errors or with 0 status code which includes a whole bunch of edge cases like max retries exceeded and timeout exceeded. If onOperationComplete is implicitly running onAttemptComplete and we only want onAttemptComplete to be run once then here we have to know if onOperationComplete is going to get called which is a hard thing to determine.

My suggestion for now is to leave this as is, but let me know if there is another alternative that addresses the timing concern.

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -285,11 +282,23 @@ export class OperationMetricsCollector {
    * @param {grpc.status} finalOperationStatus Information about the completed operation.
    * @param {number} applicationLatency The application latency measurement.
    */
-  onOperationComplete(
+  onOperationAndAttemptComplete(
```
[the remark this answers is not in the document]
danieljbruce: With the current implementation we need `onOperationComplete` for unary calls and `onOperationAndAttemptComplete` for other calls because for unary calls `onAttemptComplete` is called in the interceptor. If we find a way so that only `onOperationAndAttemptComplete` functionality is necessary then I'm happy to hear it, but this change is contingent on how we handle attempt start / completion in the interceptor.

src/row-data-utils.ts
danieljbruce: I made these renames

src/row-data-utils.ts
danieljbruce: Yes! that is a good idea.

src/client-side-metrics/operation-metrics-collector.ts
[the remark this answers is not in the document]
daniel-sanche: I think you must have misunderstood what I'm suggesting

Before this PR, you have to finalize all attempts before you call `onOperationComplete`. if you call `onOperationComplete()` when there's an active attempt, it raises an error. 

To address this, you added the new `onOperationAndAttemptComplete()`, which does work when there's an active attempt, but fails if all attempts were already completed. 

With this change, we have two distinct methods for ending an operation. You have to choose the right one depending on the state you're in. If you choose the wrong one for the state, you get an error. There's no overlap

I'm just suggesting to merge these two, into a single `onOperationComplete()` that works in either state. To accomplish this, we just need to add adding that extra state check to the top of the method. That way, that single `onOperationComplete` makes the two redundant, and will "just work" no matter what the attempt state is

Does that make sense?

src/interceptor.ts
daniel-sanche: Think of it as two states: error or success. 
- If there's an error, we can end the attempt now, whether it retries or not. This current attempt is over.
- If it's successful, we can just leave the attempt running until we reach onOperationComplete, which closes them both

To me, this feels like a small change that makes the timing a bit more accurate, which is why I brought it up in the meeting. But honestly, it doesn't matter too much how we record that final successful attempt, so if you think it's confusing, we can just forget it and leave it as-is

daniel-sanche (commented):
The code overall looks good, but with the current state of the code, wouldn't it be calling onAttemptComplete twice? (Once in interceptor, and once in onOperationAndAttemptComplete)

My suggestion was to just make onOperationComplete more flexible, so it also ends the attempt if there is one active

But I didn't mean to cause so much confusion. I'm fine with just removing onOperationAndAttemptComplete, and ending all the attempts in the interceptor. (I'd just want to be careful to make sure we can guarantee the attempt is always ended before onOperationComplete is called, because right now it needs to be in that state to work)

src/interceptor.ts
danieljbruce: Sounds good. I think I misunderstood your suggestion before about implicitly ending the attempt if necessary in onOperationComplete. I ended up adopting that suggestion because I think it is good.

In this case, I kept things the way they are because I think there is an advantage to ending all attempts at the same time as it makes the results more consistent. I see the argument for not calling onAttemptComplete for the successful calls until onOperationComplete is called as being a more accurate measurement, but making this change wouldn't improve the accuracy of the failed attempts so I'm not sure it's worth it unless there is some way to make the measurements more accurate for all attempts.

daniel-sanche (commented):
Small comments (mostly the duplicated onResponse handler)

After that, it looks good

src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -283,14 +280,20 @@ export class OperationMetricsCollector {
    * Called when an operation completes (successfully or unsuccessfully).
    * Records operation latencies, retry counts, and connectivity error counts.
    * @param {grpc.status} finalOperationStatus Information about the completed operation.
-   * @param {number} applicationLatency The application latency measurement.
```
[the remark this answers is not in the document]
danieljbruce: No. Thanks for catching this.

danieljbruce:
The duplicated handler has been removed now. Good find

daniel-sanche (approved):

PR #1657, opened 2025-07-30
test: Move the timed stream to system tests

danieljbruce:
## Description

Right now `npm run test` takes a long time to run because of the timed stream tests. `npm run test` should be a quick way to see if there are any implementation mistakes so that we don't have to wait a whole hour to find simple problems like we do with the system tests. The timed stream tests should be moved to the system tests because the system tests are already expected to take a long time so by doing this we speed up development with `npm run test` without any real drawbacks.

## Impact

This will help us develop faster.

## Testing

Same tests are in place. The timed stream tests will just run under a different github check.


daniel-sanche:
I don't like the idea of moving unit tests into the system tests folder, just to keep the tests fast. Especially because this is a temporary fix, there will always be more tests coming in the future. I think we should explore other ideas to keep the development loop light

Are you concerned about running the tests locally, or the CI?

Locally, it looks like there are a number of ways to define different test configurations, so you can run a subset of the tests if needed

PR #1658, opened 2025-07-30
fix: firstResponseLatencies should only be collected for readRows calls

danieljbruce:
## Description

This fix ensures that first response latencies are only collected for ReadRows calls and not for the other grpc methods. This matches the requirements described in https://cloud.google.com/bigtable/docs/client-side-metrics-descriptions#first-response-latencies.

## Impact

Ensures that the right client side metrics are getting collected for the right methods.

## Testing

- More tolerance has been added to application latencies
- We change the test to expect first response latencies for readRows calls and not expect them otherwise

### DIFF
--- a/src/client-side-metrics/operation-metrics-collector.ts
+++ b/src/client-side-metrics/operation-metrics-collector.ts
@@ -181,10 +181,7 @@ export class OperationMetricsCollector {
         }) => {
           this.onStatusMetadataReceived(status);
         },
-      )
-      .on('data', () => {
-        this.onResponse();
-      });
+      );
   }
 
   /**
@@ -303,6 +300,14 @@ export class OperationMetricsCollector {
         {
           this.handlers.forEach(metricsHandler => {
             if (metricsHandler.onOperationComplete) {
+              const firstResponseLatencyExpression =
+                this.methodName === MethodName.READ_ROWS ||
+                this.methodName === MethodName.READ_ROW
+                  ? {
+                      firstResponseLatency:
+                        this.firstResponseLatency ?? undefined,
+                    }
+                  : {};
               metricsHandler.onOperationComplete({
                 status: finalOperationStatus.toString(),
                 streaming: this.streamingOperation,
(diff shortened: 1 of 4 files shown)


src/client-side-metrics/operation-metrics-collector.ts
```diff
@@ -303,14 +300,23 @@ export class OperationMetricsCollector {
...
           this.handlers.forEach(metricsHandler => {
             if (metricsHandler.onOperationComplete) {
+              const firstResponseLatencyExpression =
+                this.methodName === MethodName.READ_ROWS ||
+                this.methodName === MethodName.READ_ROW
+                  ? {
+                      firstResponseLatency:
+                        this.firstResponseLatency ?? undefined,
+                    }
+                  : {};
```
daniel-sanche, line 310: I think this check would be better [here, in onOperationComplete](https://github.com/googleapis/nodejs-bigtable/blob/bcb4cba1808064df11107ae9fc51c29cbbfee53a/src/client-side-metrics/gcp-metrics-handler.ts#L254).

If we put the validation here, we now have three meaningful states to keep track of for firstRepsonseLatency: float, undefined, or unset. And that opens us up to a bunch of weird edge cases if the type doesn't match what we expect when it comes time to record the metric.

All that type complexity goes away if we only record the metric for readRows, and ignore the field altogether for other rpcs

daniel-sanche: This is a bit of an aside, but why do we use undefined here instead of null? It seems a bit confusing to me that this.firstResponseLatency is float or null, but data.firstResponseLatency is float or undefined. We wouldn't need that extra ?? operator if they used the same representation for an unset value

src/utils/createReadStreamInternal.ts
```diff
@@ -324,6 +324,9 @@ export function createReadStreamInternal(
       gaxOpts,
       retryOpts,
     });
+    requestStream.on('data', () => {
+      metricsCollector.onResponse();
```
daniel-sanche, line 328: maybe add a comment that this is for firstResponseLatency?

(This does make me wonder if we can remove this handler after it is called once for performance reasons, to avoid all those extra method calls when we only want the first one. ReadRow's on('data') is the most performance-sensitive part of the library. But maybe that's out of scope here)

daniel-sanche (changes requested):

daniel-sanche (approved):

PR #1660, opened 2025-08-01
feat: Collect client side metrics for sampleRowKeys calls

danieljbruce:
## Description

This PR makes it so that when sampleRowKeys calls are made that metrics will be collected for those calls and available to view in the Google Cloud Monitoring dashboard.

## Impact

Now we collect client side metrics for sample row keys so that users can gain insights about what happens when a sampleRowKeys call gets made.

## Testing

For all client side metrics tests that exist for ReadRows/MutateRows calls, a test will be added for sample row keys calls.


daniel-sanche (approved):
I love how simple the metrics collection part is getting

PR #1661, opened 2025-08-01
feat: Add client side metrics for checkAndMutateRow calls

danieljbruce:
## Description

This PR collects client side metrics for checkAndMutateRow calls and collects all the right types of metrics for these unary calls. They leverage the plumbing provided for unary calls to add interceptors to those calls so that we can record metrics for the right zone/cluster/server time.

## Impact

This PR collects client side metrics for checkAndMutateRow calls.

## Testing

For each readRows/mutateRows/mutateRow test we add a corresponding test for checkAndMutateRow.

### DIFF
--- a/src/client-side-metrics/metric-interceptor.ts
+++ b/src/client-side-metrics/metric-interceptor.ts
@@ -63,6 +63,19 @@ function createMetricsInterceptorProvider(
   };
 }
 
+/**
+ * Attaches a metrics interceptor to unary calls for collecting client-side metrics.
+ *
+ * This method modifies the given `gaxOptions` to include an interceptor that
+ * will be triggered during the execution of a unary gRPC call. The interceptor
+ * uses the provided `OperationMetricsCollector` to record various metrics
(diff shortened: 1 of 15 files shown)


src/client-side-metrics/metric-interceptor.ts
```diff
@@ -49,16 +49,21 @@ function createMetricsInterceptorProvider(
...
             );
+            collector.onAttemptComplete(status.code);
             nextStat(status);
           },
         };
         next(metadata, newListener);
       },
+      sendMessage: function (message, next) {
+        collector.onAttemptStart();
+        next(message);
```
daniel-sanche, line 60: this is unrelated, but maybe we should actually rename this function to create createMetricsUnaryInterceptorProvider?

I'm realizing we don't seem to mention unary anywhere in this file, and that's actually pretty important, because iirc sendMessage wouldn't be a good place to start the attempt for stream calls

daniel-sanche (approved):

PR #1675, opened 2025-08-07
test: Better factoring and generate instance id for close test

danieljbruce:
## Description

Every once in a while the close client test acts up. It says the instance has already been created. This PR may fix this flakey test and if not at the very least it will make sure two running system tests don't interfere with one another.

## Impact

Less Flakey tests

## Testing

Test changed to be less Flakey

### DIFF
--- a/system-test/read-rows.ts
+++ b/system-test/read-rows.ts
@@ -28,6 +28,7 @@ import {PassThrough} from 'stream';
 import * as proxyquire from 'proxyquire';
 import {TabularApiSurface} from '../src/tabular-api-surface';
 import * as mocha from 'mocha';
+import {generateId} from './common';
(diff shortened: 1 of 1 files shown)


system-test/read-rows.ts
```diff
@@ -152,6 +149,8 @@ describe('Bigtable/Table', () => {
   });
 
   describe('createReadStream', () => {
+    const INSTANCE = bigtable.instance('instance');
+    const TABLE = INSTANCE.table('table');
```
danieljbruce, line 153: This change is just better factoring. These two variables are only used in this block so should not be mixed with the other test to create confusion.

daniel-sanche (approved):

PR #1697, opened 2025-08-21
docs: For restoreTable and garbage collection rules samples use the admin client, not the handwritten layer

danieljbruce:
## Description

For all samples that exercise garbage collection rules or restoreTable on the generated layer, make sure those samples use the handwritten layer instead.

## Impact

This meets the need of encouraging users to use the generated admin client by changing the samples. It also provides the GCRuleMaker so that users can build the rules to be included in requests with gc rules.

## Testing

Samples tests still pass. All samples not included in the samples tests were checked manually in the examples folder.

## Additional Information

Next steps:

- When we get the typescript generator to generate the new client in the right place we should change the import statements so that we import the generated client in the new location instead of the old one.
- There were no samples for generateConsistencyToken and checkConsistency. Maybe we should add some
- Perhaps we should change all the other samples that use the `BigtableTableAdminClient` under the hood to instead use `BigtableTableAdminClient` directly.


samples/api-reference-doc-snippets/instance.js
```diff
@@ -113,35 +113,53 @@ const snippets = {
...
+    //     },
+    //   },
     // };
 
-    instance
-      .createTable(tableId, options)
+    adminClient
+      .createTable(request)
       .then(result => {
         const newTable = result[0];
```
danieljbruce, line 164: I ran this. It works

kevkim-codes (approved):

kevkim-codes:
I think our CI is broken because of the sample updates?

danieljbruce:
Before I merge this I think we should just make sure the API surface is right for users that get the `BigtableTableAdminClient` and perform operations with it.

PR #1710, opened 2025-09-03
docs: Update all the samples targeting the BigtableTableAdminClient to use the generated layer instead

danieljbruce:
## Description

In this pull request we migrate the rest of the samples that use the handwritten layer which calls BigtableTableAdminClient under the hood to instead call the BigtableTableAdminClient directly. This is a follow-up PR to https://github.com/googleapis/nodejs-bigtable/pull/1697.

## Impact

This is a crucial step towards deprecating handwritten methods that use the admin client because it encourages users to start migrating their code over to using the admin clients directly.

## Testing

This is a change to the samples so `npm run samples-test` is already a check in place that ensures these are correct.
We should consider extracting each sample into its own file and running each file just to ensure it does its job.


gkevinzheng (changes requested):
Could we add a code sample for how to generate a polling harness for `GenerateConsistencyToken` -> `CheckConsistency` polling?

gkevinzheng (approved):

kevkim-codes (approved):

PR #1613, opened 2025-06-05
feat: add support for Execute Query

kboroszko:
## Description
This PR adds support for two new methods `executeQuery` and `prepareStatement` in the Node.js Bigtable client library. The `prepareStatement` method allows the users to create SQL-like queries which can then be executed on  Bigtable tables.

## Impact

`executeQuery` provides a simpler alternative for users who want to retrieve filtered rows without manually building complex filter chains.

## Testing

- Unit tests were added.
- New testproxy implementation was also provided.
- Integration tests in the system-test directory were also added.

## Checklist

- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-bigtable/issues/new/choose) before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease
- [x] Appropriate docs were updated
- [x] Appropriate comments were added, particularly in complex areas or places that require background
- [x] No new warnings or issues will be generated from this change

PR #1715, opened 2025-09-09
fix: directly import JS-native impl for crc32c on non-x64 platforms

### LINKED ISSUE #1649: Segmentation fault on w. @google-cloud/bigtable@6.2.0
I get a segmentation fault error when running a node app with @google-cloud/bigtable@6.2.0 import.

There is no issue when using version 6.1.0. 

The error occurs when simply importing the library into a file and running it. 

  - OS: macOS Sequoia Version 15.5 (24F74)
  - Node.js version: v18.16.0
  - npm version: 9.6.6
  - `@google-cloud/bigtable` version: 6.2.0

#### Steps to reproduce

  1. npm install @google-cloud/bigtable
  2. const { Bigtable } = require("@google-cloud/bigtable");

feywind:
Directly imports JS-native implementation for fast-crc32c on non-x64 platforms to avoid a segmentation fault. The problem was first introduced in 6.2.0 by https://github.com/googleapis/nodejs-bigtable/pull/1613.

Arm architectures (and anything without SSE instructions, really) will segmentation fault on import of the bigtable library. This is caused by the node-fast-crc32c library shotgun-importing the various possible implementations, but this approach fails on Arm due to issues interfacing with the relevant native library. (https://github.com/ashi009/node-fast-crc32c/issues/38) This workaround will cause it to check the architecture and load the JS module directly if needed.

This will be slower, but it should at least make it run again on Macs and Arm VMs.

Fixes https://github.com/googleapis/nodejs-bigtable/issues/1649

## Checklist

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-bigtable/issues/new/choose) before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Code coverage does not decrease
- [?] Appropriate docs were updated
- [x] Appropriate comments were added, particularly in complex areas or places that require background
- [x] No new warnings or issues will be generated from this change

### DIFF
--- a/src/execute-query/values.ts
+++ b/src/execute-query/values.ts
@@ -16,7 +16,13 @@ import {Duplex, Readable} from 'stream';
 import * as SqlTypes from './types';
 import {PreciseDate} from '@google-cloud/precise-date';
 import {NamedList} from './namedlist';
-const CRC32C = require('fast-crc32c');
+
+// fast-crc32c will segfault on import on non-x64 architectures, due to a
+// lack of SSE instructions.
+const CRC32C =
+  process.arch === 'x64'
+    ? require('fast-crc32c')
+    : require('fast-crc32c/impls/js_crc32c');
 
 export type BigtableMap = EncodedKeyMap;
 


danieljbruce (approved):

feywind:
Speaking of conformance tests... I'm seeing the same thing on my local machine, so at least this is back to parity.

PR #1719, opened 2025-09-16
tests: conformance test fixes

feywind:
## Description

This will fix as many conformance test failures as possible. It's a work in progress that will be open for a bit.

## Checklist

- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-bigtable/issues/new/choose) before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Code coverage does not decrease
- [ ] Appropriate docs were updated
- [ ] Appropriate comments were added, particularly in complex areas or places that require background
- [ ] No new warnings or issues will be generated from this change

Fixes #issue_number_goes_here 🦕

### DIFF
--- a/src/row.ts
+++ b/src/row.ts
@@ -30,7 +30,7 @@ import {CallOptions} from 'google-gax';
 import {ServiceError} from 'google-gax';
 import {google} from '../protos/protos';
 import {RowDataUtils, RowProperties} from './row-data-utils';
-import {TabularApiSurface} from './tabular-api-surface';
+import {GetRowsOptions, TabularApiSurface} from './tabular-api-surface';
 import {getRowsInternal} from './utils/getRowsInternal';
 import {
   MethodName,
@@ -667,9 +667,10 @@ export class Row {
       filter = arrify(filter).concat(options.filter);
     }
 
-    const getRowsOptions = Object.assign({}, options, {
+    const getRowsOptions: GetRowsOptions = Object.assign({}, options, {
       keys: [this.id],
       filter,
+      limit: 1,
     });
 
     const metricsCollector =
--- a/src/utils/createReadStreamInternal.ts
+++ b/src/utils/createReadStreamInternal.ts
@@ -428,6 +428,19 @@ export function createReadStreamInternal(
     rowStreamPipe(rowStream, userStream);
   };
 
+  // If the timeout is exceeded for the whole operation, bail with
+  // an error as the conformance test requires.
+  if (timeout) {
+    const deadlineTimer = setTimeout(() => {
+      const err = new Error(`Total timeout of ${timeout}ms exceeded.`);
+      (err as unknown as grpc.StatusObject).code =
+        grpc.status.DEADLINE_EXCEEDED;
+      userStream.destroy(err);
+    }, timeout);
+
+    userStream.on('close', () => clearTimeout(deadlineTimer));
+  }
+
   makeNewRequest();
   return userStream;
 }
--- a/testproxy/known_failures.txt
+++ b/testproxy/known_failures.txt
@@ -1,4 +1,3 @@
-TestMutateRow_Generic_DeadlineExceeded\|
 TestMutateRow_Generic_CloseClient\|
 TestMutateRows_Retry_WithRoutingCookie\|
 TestReadRow_Generic_DeadlineExceeded\|
--- a/testproxy/known_failures.txt
+++ b/testproxy/known_failures.txt
@@ -1,6 +1,5 @@
 TestMutateRow_Generic_CloseClient\|
 TestMutateRows_Retry_WithRoutingCookie\|
-TestReadRow_Generic_DeadlineExceeded\|
 TestReadRow_Retry_WithRoutingCookie\|
 TestReadRow_Retry_WithRetryInfo\|
 TestReadRows_ReverseScans_FeatureFlag_Enabled\|
--- a/testproxy/services/mutate-row.js
+++ b/testproxy/services/mutate-row.js
@@ -28,16 +28,22 @@ const mutateRow = ({clientMap}) =>
     const appProfileId = clientMap.get(clientId).appProfileId;
     const client = clientMap.get(clientId)[v2];
 
-    await client.mutateRow({
-      appProfileId,
-      mutations,
-      tableName,
-      rowKey,
-    });
+    try {
+      await client.mutateRow({
+        appProfileId,
+        mutations,
+        tableName,
+        rowKey,
+      });
 
-    return {
-      status: {code: grpc.status.OK, details: []},
-    };
+      return {
+        status: {code: grpc.status.OK, details: []},
+      };
+    } catch (e) {
+      return {
+        status: e,
+      };
+    }
   });
 
 module.exports = mutateRow;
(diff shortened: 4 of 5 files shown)


danieljbruce (changes requested):
Overall the migration to typescript looks great and there are a few thoughts I would mention and post before doing the rest of the review.

1. We can just skip the Service Path tests and then the system-test CI check should pass. I think before these tests were just skipped anyways and used as a handy tool to be sure that the universe domain feature was working correctly.

2. Apart from that I think we should add TestReadRows_NoRetry_OutOfOrderError, TestReadRows_Generic_CloseClient and TestReadRows_Generic_DeadlineExceeded tests to the known_failures.txt list so that the CI check for mandatory conformance turns green. This tool is in place to make it really easy for us to determine exactly what we need to fix next.

src/utils/createReadStreamInternal.ts
```diff
@@ -428,6 +428,19 @@ export function createReadStreamInternal(
...
+    const deadlineTimer = setTimeout(() => {
+      const err = new Error(`Total timeout of ${timeout}ms exceeded.`);
+      (err as unknown as grpc.StatusObject).code =
+        grpc.status.DEADLINE_EXCEEDED;
+      userStream.destroy(err);
+    }, timeout);
+
+    userStream.on('close', () => clearTimeout(deadlineTimer));
+  }
+
```
danieljbruce: It looks like `TestReadRows_Generic_DeadlineExceeded` test is still failing even though we have this code block though. Can we address the `TestReadRows_Generic_DeadlineExceeded` test in a separate PR and eliminate this block of code? This readRows code is really sensitive to data loss / data duplication issues and timing relationships between stream destruction and retries.

src/row.ts
```diff
@@ -667,9 +667,10 @@ export class Row {
       filter = arrify(filter).concat(options.filter);
     }
 
-    const getRowsOptions = Object.assign({}, options, {
+    const getRowsOptions: GetRowsOptions = Object.assign({}, options, {
       keys: [this.id],
       filter,
+      limit: 1,
```
danieljbruce: This is a good addition. I would add a comment here just saying something like we only need 1 row for this method so we should only make the request for one row.

test/table.ts
```diff
@@ -1190,7 +1190,10 @@ describe('Bigtable/Table', () => {
           .on('data', done);
       });
     });
-    it('Should respect the timeout parameter passed in for UNAVAILABLE error', done => {
+
+    // Skip: This test currently doesn't make sense after conformance test
+    // changes. The error will always be DEADLINE_EXCEEDED, not UNAVAILABLE.
+    it.skip('Should respect the timeout parameter passed in for UNAVAILABLE error', done => {
```
danieljbruce: Instead of skipping this test can you just change the assert statements?

testproxy/services/close-client.ts
```diff
@@ -11,24 +11,23 @@
...
+  ICloseClientResponse
+> = ({clientMap}) =>
   normalizeCallback(async rawRequest => {
     const request = rawRequest.request;
     const {clientId} = request;
-    const bigtable = clientMap.get(clientId);
+    const bigtable = clientMap.get(clientId!);
 
     if (bigtable) {
-      await bigtable[v2].close();
```
danieljbruce, line 27: I don't know if we want to remove this. I notice that `TestReadRows_Generic_CloseClient` is now failing so this may be causing a regression in the conformance tests so I would consider adding the following lines back in:

`await bigtable[v2].close();`
`clientMap.delete(clientId);`

testproxy/services/utils/bigtable-client.ts
```diff
@@ -0,0 +1,35 @@
...
+// limitations under the License.
+
+import {Bigtable} from '../../../src';
+import {ClientSideMetricsConfigManager} from '../../../src/client-side-metrics/metrics-config-manager';
+import {IMetricsHandler} from '../../../src/client-side-metrics/metrics-handler';
+import {BigtableClient} from '../../../src/v2';
+
+const v2 = Symbol.for('v2');
+
+export function createBigtableClient(bigtable: Bigtable) {
```
danieljbruce, line 23: nit: If this createBigtableClient method is used in only one place then I would prefer to keep this method in the file where it is used to encapsulate things a little more.

Same idea with getBigtableClient.

testproxy/services/close-client.ts
danieljbruce, line 27: Should we delete this line and clientMap.delete(clientId); line? They might be causing a regression making the `TestReadRows_Generic_CloseClient` test fail.

testproxy/testlog.txt
```diff
@@ -0,0 +1,459 @@
...
+        	Error Trace:	/Users/mzp/git/cloud-bigtable-clients-test/tests/test_workflow.go:669
+        	            				/Users/mzp/git/cloud-bigtable-clients-test/tests/samplerowkeys_test.go:274
+        	Error:      	Should be empty, but was 14
+        	Test:       	TestSampleRowKeys_Retry_WithRoutingCookie
+    samplerowkeys_test.go:276: 
+        	Error Trace:	/Users/mzp/git/cloud-bigtable-clients-test/tests/samplerowkeys_test.go:276
+        	Error:      	Not equal: 
+        	            	expected: 1
+        	            	actual  : 0
+        	Test:       	TestSampleRowKeys_Retry_WithRoutingCookie
```
danieljbruce: How come you want to check the test log into the git commit?

danieljbruce (changes requested):
Overall the test proxy modernization looks really good. Outside of that there are a few comments about some of the proxy changes and source code changes we should discuss, but after those are resolved this should be good for approval.

src/utils/createReadStreamInternal.ts
feywind: I was waffling back and forth about moving the fixes thus far into a separate PR, and I think I'm just going to do that and let this one be about the testproxy.

test/table.ts
feywind: The reason I added the skip is because Gemini convinced me (I had to go research it myself to convince myself too) that this test no longer makes sense with the fixes elsewhere, and I wanted to verify that with you before deleting it. I'll just move these fixes into a separate PR though.

testproxy/services/close-client.ts
feywind: ...this might've been a mistake on my part. Thanks!

testproxy/testlog.txt
feywind: I don't 😹 Thank you.

src/row.ts
feywind: This is moved to the separate PR now.

src/utils/createReadStreamInternal.ts
feywind: This is moved to the separate PR now.

test/table.ts
feywind: This is moved to the separate PR now.

testproxy/services/utils/bigtable-client.ts
feywind: It turns out that was a mistake anyway, related to the close bugs. This is now used in two other files. The reason I'd moved it out like this is just to make the create/close logic closer together.

src/index.ts
```diff
@@ -112,6 +112,11 @@ export interface BigtableOptions extends gax.GoogleAuthOptions {
   BigtableTableAdminClient?: gax.ClientOptions;
 
   metricsEnabled?: boolean;
+
+  /**
+   * Internal only.
+   */
+  clientConfig?: gax.ClientConfig;
```
danieljbruce: Maybe we could remove this one and then I think it is good to merge.

PR #1721, opened 2025-09-18
docs: All samples should include all necessary code

danieljbruce:
## Description

This PR is an extension of https://github.com/googleapis/nodejs-bigtable/pull/1720. The basic idea is that the rest of the samples should include all the code necessary to get the sample running.

## Impact

It makes it easier for users to use our samples.

## Testing

These are changes to the samples so they are covered by samples tests


samples/tableadmin.js
```diff
@@ -13,293 +13,381 @@
 // limitations under the License.
```
danieljbruce, line 13: For pretty much all the samples in this section we are pulling:

```
const {Bigtable, GCRuleMaker} = require('@google-cloud/bigtable');
const {BigtableTableAdminClient} = require('@google-cloud/bigtable').v2;
const adminClient = new BigtableTableAdminClient();
```

into the samples and indenting the code that contains a sample so that these declarations can be done multiple times.

PR #1742, opened 2025-11-11
chore: revert "build: owlbot.py changes for merging split gapic clients correctly (#1739)

feywind:
We're backing this out temporarily into a separate branch, will be recommitting later.

This reverts commit 6e461cdc46cc54fb0d7fc5447c91eabc4e179699.


feywind:
We won't be doing this after all.

